PR #4, opened 2022-06-28
Track regions of the TVM guest physical address space

abrestic-rivos:
Introduce `VmRegionList` which tracks the distinct regions in a TVM's guest physical address space, similar in principle to KVM `memslots`. This will be used in follow-on patches to support demand faulting of zero pages and emulated MMIO (and can also be used to support shared memory).

Patches 1 and 2 are minor `PageVec` fixes, patch 3 fixes an error path when mapping pages, patch 4 introduces the VM region tracking, and patch 5 removes hooks for handling demand faults when copying to/from the guest address space.

### DIFF
--- a/src/vm_pages.rs
+++ b/src/vm_pages.rs
@@ -328,13 +328,23 @@ impl<'a, T: GuestStagePageTable> ActiveVmPages<'a, T> {
     ) -> Result<u64> {
         let converted_pages = self.get_converted_pages(from_addr, count)?;
         let mapper = to.map_pages(to_addr, count)?;
+
+        // Make sure we can initialize the full set of pages before mapping them.
+        let mut initialized_pages = LockedPageList::new(self.page_tracker.clone());
+        for (dirty, src_addr) in converted_pages.zip(src_addr.iter_from()) {
+            match dirty.try_initialize(|bytes| self.copy_from_guest(bytes, src_addr.into())) {
+                Ok(p) => initialized_pages.push(p).unwrap(),
+                Err((e, p)) => {
+                    // Unwrap ok since the page must have been locked.
+                    self.page_tracker.put_converted_page(p).unwrap();
+                    return Err(e);
+                }
+            };
+        }
+
+        // Now map & measure the pages.
         let new_owner = to.page_owner_id();
-        for (dirty, (src_addr, to_addr)) in
-            converted_pages.zip(src_addr.iter_from().zip(to_addr.iter_from()))
-        {
-            let initialized = dirty
-                .try_initialize(|bytes| self.copy_from_guest(bytes, src_addr.into()))
-                .map_err(|(e, _)| e)?;
+        for (initialized, to_addr) in initialized_pages.zip(to_addr.iter_from()) {
             // Unwrap ok since we've guaranteed there's space for another owner.
             let mappable = self
                 .page_tracker
@@ -343,6 +353,7 @@ impl<'a, T: GuestStagePageTable> ActiveVmPages<'a, T> {
             // Unwrap ok since the address is in range and we haven't mapped it yet.
             mapper.map_page_with_measurement(to_addr, mappable).unwrap();
         }
+
         Ok(count)
     }
 }
--- a/src/vm_pages.rs
+++ b/src/vm_pages.rs
@@ -26,7 +26,7 @@ use crate::vm_id::VmId;
 pub enum Error {
     GuestId(PageTrackingError),
     Paging(PageTableError),
-    PageFaultHandling, // TODO - individual errors from sv48x4
+    UnhandledPageFault,
     NestingTooDeep,
     // Page table root must be aligned to 16k to be used for sv48x4 mappings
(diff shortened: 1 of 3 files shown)


src/vm_pages.rs
```diff
@@ -471,26 +470,16 @@ impl<'a, T: GuestStagePageTable> ActiveVmPages<'a, T> {
...
-            }
+        this_cpu.enter_guest_memcpy();
+        let bytes = copy_fn(guest_addr, host_ptr, len);
+        this_cpu.exit_guest_memcpy();
+        if bytes == len {
+            Ok(())
+        } else {
+            // TODO: Report faulting address and region so that the caller can turn it into a VM
+            // exit code.
+            Err(Error::UnhandledPageFault)
```
dgreid, line 481: What will the flow look like if the guest asks to copy a measurement to an as-yet-unmapped page?

TVM -> ecall (get measuement) -> Fault -> Host(Exit code indicating fault and address) -> Fill page dance -> TvmRun -> Result gets at ecall instruction and restart from scratch?

src/vm_pages.rs
```diff
@@ -151,31 +158,201 @@ impl TlbTracker {
...
+/// A contiguous region of guest physical address space.
+struct VmRegion {
+    start: GuestPageAddr,
+    end: GuestPageAddr,
+    region_type: VmRegionType,
+}
+
+/// Represents a reserved region in `VmRegionList` that has yet to be committed. Removes the reserved
+/// region if dropped without being passed to `VmRegionList::commit()`.
+pub struct VmReservedRegion<'a> {
```
dgreid, line 180: Does this need to be pub?

src/vm_pages.rs
```diff
@@ -151,31 +158,201 @@ impl TlbTracker {
...
+            regions: Mutex::new(PageVec::new(pages, page_tracker)),
+        }
+    }
+
+    /// Reserves a region at [`start`, `end`) of type `region_type`, returning a `VmReservedRegion`
+    /// object if it does not conflict with any existing regions. After performing any actions
+    /// necessary to populate the region (e.g. filling in page tables), the `VmReservedRegion` should
+    /// be passed to `commit()` to commit it to the `VmRegionList`. If the `VmReservedRegion` is
+    /// dropped before it is committed, the reservation is removed.
+    pub fn prepare(
```
dgreid, line 225: pub needed? Seems to only be used in this file.

src/vm_pages.rs
```diff
@@ -151,31 +158,201 @@ impl TlbTracker {
...
+
+/// A contiguous region of guest physical address space.
+struct VmRegion {
+    start: GuestPageAddr,
+    end: GuestPageAddr,
+    region_type: VmRegionType,
+}
+
+/// Represents a reserved region in `VmRegionList` that has yet to be committed. Removes the reserved
+/// region if dropped without being passed to `VmRegionList::commit()`.
```
dgreid, line 179: I would find it useful to have an example or basic flow overview of how/why this is used with the `prepare` and `commit` functions below.

dgreid (commented):
overall looks good, just a few questions.

Thanks!

src/vm_pages.rs
atulkharerivos: In this case, we'll think that we'll fault in the copy_to_guest(), detect that this a demand-zero page that needs to be mapped in, exit to the host with the address and guest fault address, wait for the map page dance, then following TvmRun, will attempt to copy again. It might get interesting if the host retries without actually mapping a page (in which case we'll have a loop).

src/vm_pages.rs
atulkharerivos: I have a very similar structure in my upcoming patch for page sharing....we can look into combining them in the future.

src/vm_pages.rs
abrestic-rivos: Yup, exactly that flow. The host is free to continuously try running a vCPU without fixing the page fault if it wants to; denial-of-service by the host is out-of-scope.

src/vm_pages.rs
```diff
@@ -26,7 +26,7 @@ use crate::vm_id::VmId;
 pub enum Error {
     GuestId(PageTrackingError),
     Paging(PageTableError),
-    PageFaultHandling, // TODO - individual errors from sv48x4
+    UnhandledPageFault, // TODO - individual errors from sv48x4
```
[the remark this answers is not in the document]
abrestic-rivos: Done.

src/vm_pages.rs
abrestic-rivos: Nope, removed.

src/vm_pages.rs
abrestic-rivos: Removed.

src/vm_pages.rs
abrestic-rivos: Expanded the comment above `VmRegionList` with more details on why along with a simple example.

dgreid:
LGTM, thanks!

dgreid (approved):

PR #5, opened 2022-06-29
Add page sharing prototype

atulkharerivos:
- Adds SBI calls for TsmDefineSharedPagesRange and TsmInsertSharedPages
- Changes to map shared pages, including ensuring that the confidential and non-confidential ranges don't overlap
- Changes to test workloads to  use the new SBI calls, and fault-in shared pages on demand

### DIFF
--- a/page-tracking/src/page_info.rs
+++ b/page-tracking/src/page_info.rs
@@ -28,6 +28,11 @@ pub enum PageState {
     /// Page is mapped into the address space of the current owner.
     Mapped,
 
+    /// Page is mapped into the address space of the current owner, but has been shared with
+    /// a child-VM. Shared pages cannot be converted, and are reverted to the Mapped state when
+    /// the reference count maintained by the inner u64 drops to 0.
+    Shared(u64),
+
     /// Page is used to store hypervisor-internal state for the current owner, e.g. to back per-VM
     /// data structures or as a page-table page for the VM.
     VmState,
@@ -38,9 +43,6 @@ pub enum PageState {
     /// Page has completed the conversion operation and is eligible for assignment or to be reclaimed.
     /// The page must be locked exclusively before it can be assigned or reclaimed.
     Converted,
-
-    /// Page has completed the conversion operation and is locked pending assignment or reclaim.
-    ConvertedLocked,
 }
 
 /// The maximum length for an ownership chain. Enough for the host VM to assign to a guest VM
@@ -55,6 +57,8 @@ pub struct PageInfo {
     owners: PageOwnerVec,
     // Address of the next page in the list if != None.
     link: Option<NonZeroU64>,
+    // True if the page has been locked for exclusive ownership
+    locked: bool,
 }
 
 impl PageInfo {
@@ -65,6 +69,7 @@ impl PageInfo {
             state: PageState::Free,
             owners: PageOwnerVec::new(),
             link: None,
+            locked: false,
         }
     }
 
@@ -72,9 +77,10 @@ impl PageInfo {
     pub fn new_hypervisor_owned() -> Self {
         Self {
             mem_type: MemType::Ram,
-            state: PageState::ConvertedLocked,
+            state: PageState::Converted,
             owners: PageOwnerVec::new(),
             link: None,
+            locked: true,
         }
     }
 
@@ -85,6 +91,7 @@ impl PageInfo {
             state: PageState::Reserved,
             owners: PageOwnerVec::new(),
             link: None,
+            locked: false,
         }
     }
 
@@ -92,9 +99,10 @@ impl PageInfo {
     pub fn new_mmio(dev_type: DeviceMemType) -> Self {
         Self {
             mem_type: MemType::Mmio(dev_type),
-            state: PageState::ConvertedLocked,
+            state: PageState::Converted,
             owners: PageOwnerVec::new(),
             link: None,
+            locked: true,
         }
     }
 
@@ -102,7 +110,7 @@ impl PageInfo {
     pub fn owner(&self) -> Option<PageOwnerId> {
         use PageState::*;
         match self.state {
-            Converting(_) | Converted | ConvertedLocked | Mapped | VmState => {
+            Converting(_) | Converted | Mapped | VmState | Shared(_) => {
                 if !self.owners.is_empty() {
                     Some(self.owners[self.owners.len() - 1])
                 } else {
@@ -135,28 +143,42 @@ impl PageInfo {
     }
 
     /// Pops the current owner if there is one, returning the page to the previous owner.
-    pub fn release(&mut self) -> PageTrackingResult<PageOwnerId> {
+    pub fn release(&mut self) -> PageTrackingResult<()> {
         use PageState::*;
         match self.state {
             Mapped | VmState | Converted | Converting(_) => {
+                // Locked pages cannot be released
+                if self.locked {
+                    return Err(PageTrackingError::PageLocked);
+                }
                 if self.owners.is_empty() {
                     Err(PageTrackingError::OwnerUnderflow) // Can't pop the last owner.
                 } else {
-                    let owner = self.owners.pop().unwrap();
+                    self.owners.pop().unwrap();
                     self.state = Converted;
-                    Ok(owner)
+                    Ok(())
                 }
             }
-            ConvertedLocked => Err(PageTrackingError::PageLocked),
+            Shared(rc) => {
+                // Shared pages start with a RC of 1, so RC is always > 0 here
+                if let Some(rc) = rc.checked_sub(1) {
+                    self.state = if rc == 0 { Mapped } else { Shared(rc) };
+                    // Unwrap ok: Shared pages must have an owner
+                    Ok(())
+                } else {
+                    Err(PageTrackingError::RefCountUnderflow)
+                }
+            }
             Reserved => Err(PageTrackingError::ReservedPage),
             Free => Err(PageTrackingError::UnownedPage),
         }
     }
 
-    /// Assigns the page to `owner` with state `new_state`. The page must be locked for assingment.
+    /// Assigns the page to `owner` with state `new_state`. A page in Converted state must be locked for
+    /// assignment, and the lock is automatically released on success.
     pub fn assign(&mut self, owner: PageOwnerId, new_state: PageState) -> PageTrackingResult<()> {
         use PageState::*;
-        if !matches!(new_state, Mapped | VmState | Converted | ConvertedLocked) {
+        if !matches!(new_state, Mapped | VmState | Converted) {
             // Going back to free/reserved isn't allowed, nor does it make sense for a page to
             // immediately enter the "Converting" state.
             return Err(PageTrackingError::InvalidStateTransition);
@@ -171,8 +193,12 @@ impl PageInfo {
                 self.state = new_state;
                 Ok(())
             }
-            ConvertedLocked => {
-                if matches!(new_state, Converted | ConvertedLocked) {
+            Converted => {
+                // Pages must be locked for assignment
+                if !self.locked {
+                    return Err(PageTrackingError::PageNotLocked);
+                }
+                if matches!(new_state, Converted) {
                     // A page converted by a VM can't immediately become a converted page in a child
                     // VM.
                     return Err(PageTrackingError::InvalidStateTransition);
@@ -181,9 +207,9 @@ impl PageInfo {
                     .try_push(owner)
                     .map_err(|_| PageTrackingError::OwnerOverflow)?;
                 self.state = new_state;
+                self.unlock().unwrap();
                 Ok(())
             }
-            Converted => Err(PageTrackingError::PageNotLocked),
             _ => Err(PageTrackingError::PageNotAssignable),
         }
     }
@@ -226,41 +252,82 @@ impl PageInfo {
         }
     }
 
-    /// Transitions the page to the ConvertedLocked state from Converted, indicating that an exclusive
-    /// reference has been taken to the page in preparation for assignment or reclaim.
+    /// Returns if the page can be Shared.
+    pub fn is_shareable(&self) -> bool {
+        use PageState::*;
+        match self.state() {
+            Shared(rc) if rc == u64::MAX => false,
+            Shared(_) | PageState::Mapped => true,
+            _ => false,
+        }
+    }
+
+    /// Returns if the page is Shared
+    pub fn is_shared(&self) -> bool {
+        use PageState::*;
+        matches!(self.state, Shared(_))
+    }
+
+    /// Obtains an exclusive reference on a page in an assignable state in preparation
+    /// for assignment, reclaim, or sharing.
     pub fn lock_for_assignment(&mut self) -> PageTrackingResult<()> {
         use PageState::*;
         match self.state {
-            Converted => {
-                self.state = ConvertedLocked;
-                Ok(())
+            Converted | Mapped | Shared(_) => {
+                if !self.locked {
+                    self.locked = true;
+                    Ok(())
+                } else {
+                    Err(PageTrackingError::PageLocked)
+                }
             }
-            ConvertedLocked => Err(PageTrackingError::PageLocked),
             _ => Err(PageTrackingError::PageNotAssignable),
         }
     }
 
-    /// Drops the exclusive lock on the a converted page.
+    /// Drops the exclusive lock on a page.
     pub fn unlock(&mut self) -> PageTrackingResult<()> {
+        if self.locked {
+            self.locked = false;
+            Ok(())
+        } else {
+            Err(PageTrackingError::PageNotLocked)
+        }
+    }
+
+    /// Transitions a locked page from Mapped to Shared, or increments the RC on a Shared page.
+    /// The lock is dropped on success.
+    pub fn share(&mut self) -> PageTrackingResult<()> {
         use PageState::*;
         match self.state {
-            ConvertedLocked => {
-                self.state = Converted;
+            Mapped => {
+                self.unlock()?;
+                self.state = Shared(1);
                 Ok(())
             }
-            _ => Err(PageTrackingError::PageNotLocked),
+            Shared(rc) => {
+                if let Some(rc) = rc.checked_add(1) {
+                    self.unlock()?;
+                    self.state = Shared(rc);
+                    Ok(())
+                } else {
+                    Err(PageTrackingError::RefCountOverflow)
+                }
+            }
+            _ => Err(PageTrackingError::PageNotShareable),
         }
     }
 
     /// Reclaims the converted and locked, but unassigned, page as a Mapped page for the current owner.
+    /// The lock is dropped on success.
     pub fn reclaim(&mut self) -> PageTrackingResult<()> {
         use PageState::*;
         match self.state {
-            ConvertedLocked => {
+            Converted => {
+                self.unlock()?;
                 self.state = Mapped;
                 Ok(())
             }
-            Converted => Err(PageTrackingError::PageNotLocked),
             // TODO: Reclaim pages that are converting but not yet fully converted?
             _ => Err(PageTrackingError::PageNotReclaimable),
         }
@@ -633,8 +700,9 @@ mod tests {
         let mut page = PageInfo::new();
         assert!(page.is_free());
         assert!(page
-            .assign(PageOwnerId::hypervisor(), PageState::ConvertedLocked)
+            .assign(PageOwnerId::hypervisor(), PageState::Converted)
             .is_ok());
+        assert!(page.lock_for_assignment().is_ok());
         assert!(page.assign(PageOwnerId::host(), PageState::Mapped).is_ok());
         assert_eq!(page.owner().unwrap(), PageOwnerId::host());
         let version = TlbVersion::new();
@@ -647,9 +715,9 @@ mod tests {
         assert!(page.lock_for_assignment().is_ok());
         let guest_id = PageOwnerId::new(2).unwrap();
         assert!(page.assign(guest_id, PageState::Mapped).is_ok());
-        assert_eq!(page.release().unwrap(), guest_id);
+        assert_eq!(page.release().unwrap(), ());
         assert!(page.lock_for_assignment().is_ok());
-        assert_eq!(page.state(), PageState::ConvertedLocked);
+        assert_eq!(page.state(), PageState::Converted);
         assert!(page.reclaim().is_ok());
 
         let mut page = PageInfo::new_reserved();
--- a/page-tracking/src/page_tracker.rs
+++ b/page-tracking/src/page_tracker.rs
@@ -29,12 +29,19 @@ pub enum Error {
     ReservedPage,
     /// The page is not owned by the specified owner.
     OwnerMismatch,
+    /// Attempt to modify the owner of a shared page.
+    /// At present, all shared pages are owned by the host
+    SharedPage,
     /// The page is not in a state where it can be converted.
     PageNotConvertible,
     /// The page is not in a state where it can be assigned.
     PageNotAssignable,
     /// The page cannot be mapped back into the owner's address space.
     PageNotReclaimable,
+    /// The page cannot be shared.
+    PageNotShareable,
+    /// The page isn't in Shared state.
+    PageNotShared,
     /// Attempt to intiate an invalid page state transition.
     InvalidStateTransition,
     /// Attempt to assign or reclaim a page that is not locked.
@@ -43,6 +50,10 @@ pub enum Error {
     PageLocked,
     /// Attempt to create a link from a page that is already linked.
     PageAlreadyLinked,
+    /// The ref count was at u64::MAX.
+    RefCountOverflow,
+    /// The ref count was already 0.
+    RefCountUnderflow,
 }
 
 /// Holds the result of page tracking operations.
@@ -193,6 +204,22 @@ impl PageTracker {
         Ok(unsafe { P::MappablePage::new_with_size(page.addr(), page.size()) })
     }
 
+    /// Consumes the shared `page`, returning a page that can then be mapped into
+    /// a page table.
+    pub fn share_page<P>(&self, page: P, owner: PageOwnerId) -> Result<P::MappablePage>
+    where
+        P: ShareablePhysPage,
+    {
+        let mut page_tracker = self.inner.lock();
+        let info = page_tracker.get_mut(page.addr()).unwrap();
+        if info.owner() != Some(owner) || info.mem_type() != P::mem_type() || !info.is_shareable() {
+            return Err(Error::PageNotShareable);
+        }
+        info.share()?;
+        // Safe since we own the page and have updated its state.
+        Ok(unsafe { P::MappablePage::new_with_size(page.addr(), page.size()) })
+    }
+
     /// Assigns `page` as an internal state page for `owner`, returning a page that is eligible to
     /// be used with various internal collection types (e.g. `PageBox<>`).
     pub fn assign_page_for_internal_state(
@@ -220,7 +247,8 @@ impl PageTracker {
     pub fn release_page_by_addr(&self, addr: SupervisorPageAddr, owner: PageOwnerId) -> Result<()> {
         let mut page_tracker = self.inner.lock();
         let info = page_tracker.get_mut(addr)?;
-        if info.owner() != Some(owner) {
+        // Shared pages might be owned by the parent
+        if info.owner() != Some(owner) && !info.is_shared() {
             return Err(Error::OwnerMismatch);
         }
         info.release()?;
@@ -274,8 +302,8 @@ impl PageTracker {
         Ok(unsafe { P::DirtyPage::new(addr) })
     }
 
-    /// Releases an exclusive reference to a converted and locked page.
-    pub fn put_converted_page<P: ConvertedPhysPage>(&self, page: P) -> Result<()> {
+    /// Releases an exclusive reference to a locked page
+    pub fn unlock_page<P: PhysPage>(&self, page: P) -> Result<()> {
         let mut page_tracker = self.inner.lock();
         let info = page_tracker.get_mut(page.addr())?;
         info.unlock()
@@ -296,6 +324,25 @@ impl PageTracker {
         } else {
             false
         }
+    }
+
+    /// Acquires an exclusive reference to the shareable page at `addr` if it's owned by owner,
+    /// and in Mapped or Shared state.
+    pub fn get_shareable_page<P: ShareablePhysPage>(
+        &self,
+        addr: SupervisorPageAddr,
+        owner: PageOwnerId,
+    ) -> Result<P> {
+        let mut page_tracker = self.inner.lock();
+        let info = page_tracker.get_mut(addr)?;
+        if info.owner() != Some(owner) || info.mem_type() != P::mem_type() || !info.is_shareable() {
+            return Err(Error::PageNotShareable);
+        }
+
+        info.lock_for_assignment()?;
+        // Safe since we've taken exclusive ownership of the page, verified its typing.
+        // TODO: Page size
+        Ok(unsafe { P::new(addr) })
     }
 
     /// Returns true if and only if `addr` is a page owned by `owner` with type `mem_type` and
@@ -377,9 +424,9 @@ impl HypPageAlloc {
         let mut tail: Option<SupervisorPageAddr> = None;
         while let Some(next) = self.next_page {
             let info = self.pages.get_mut(next).unwrap();
-            info.assign(PageOwnerId::hypervisor(), PageState::ConvertedLocked)
+            info.assign(PageOwnerId::hypervisor(), PageState::Converted)
                 .unwrap();
-
+            info.lock_for_assignment().unwrap();
             // Safe to create this page as it was previously free and we just took ownership.
             let page: Page<ConvertedDirty> = unsafe { Page::new(next) };
             page.clean();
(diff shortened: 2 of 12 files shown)


riscv-page-tables/src/page_table.rs
```diff
@@ -797,6 +879,50 @@ impl<T: PagingMode> PlatformPageTable<T> {
...
+
+        let mut pages = LockedSharedPageList::new(inner.page_tracker.clone());
+        for a in addr.iter_from().take(num_pages as usize) {
+            // We verified above that we can safely unwrap here.
+            let entry = inner.get_shareable_4k_leaf(a, MemType::Ram).unwrap();
+            // Unwrap ok, PFN must have been properly aligned in order to have been mapped.
+            let paddr = entry.page_addr();
+            inner
+                .page_tracker
+                .lock_shareable_page(paddr, inner.owner, MemType::Ram)
```
abrestic-rivos, line 914: nit: `get_shareable_page()` to be consistent with `get_converted_page()`.

riscv-page-tables/src/page_table.rs
```diff
@@ -797,6 +879,50 @@ impl<T: PagingMode> PlatformPageTable<T> {
...
+        page_size: PageSize,
+        num_pages: u64,
+    ) -> Result<LockedSharedPageList<P>> {
+        if page_size.is_huge() {
+            return Err(Error::PageSizeNotSupported(page_size));
+        }
+
+        let mut inner = self.inner.lock();
+
+        // First make sure the entire range can be unmapped before we start invalidating things.
```
abrestic-rivos, line 897: I don't think the first pass is necessary here since locking the page for sharing doesn't have any host-visible side-effects like invalidation does and `LockedPageList` will clean up any locked pages if we bail out early.

riscv-pages/src/page.rs
```diff
@@ -576,6 +579,7 @@ impl ReclaimablePhysPage for Page<ConvertedClean> {
     type MappablePage = Page<MappableClean>;
 }
 
+impl ShareablePhysPage for Page<Shareable> {}
 /// An iterator of the 64-bit words contained in a page.
```
abrestic-rivos, line 583: nit: newline between the `impl ...` and the comment below.

riscv-pages/src/state.rs
```diff
@@ -111,3 +111,9 @@ impl InternalData for InternalDirty {}
 impl Cleanable for InternalDirty {
     type Cleaned = InternalClean;
 }
+
+/// A mapped page that is eligible for conversion to a Shared page
+#[derive(Debug)]
+pub enum Shareable {}
```
abrestic-rivos, line 117: I wonder if there should be two states here: `Mapped` pages that are currently mapped in a page table but haven't yet been marked as shared, and `MappableShared` for `Mapped` pages that have been marked as shared in `PageTracker`.

sbi/src/sbi.rs
```diff
@@ -738,6 +738,35 @@ pub enum TeeFunction {
     ///
     /// a6 = 15
     TsmLocalFence,
+    /// Defines the single GPA range that will be used to map non-confidential shared pages.
+    /// The call made not be made after TVM is finalized, and the pages in the range cannot overlap
+    /// with the GPAs for confidential pages
+    /// a6 = 17
+    TsmDefineSharedPagesRange {
```
abrestic-rivos, line 745: Two things:
- The `Tvm` prefix would be more appropriate here since it's an operation on a TVM rather than the TSM itself.
- Limiting the TVM to one shared page range seems a bit overly restrictive, at least for a public ABI (the implementation of course could refuse to add more than one).

sbi/src/sbi.rs
```diff
@@ -738,6 +738,35 @@ pub enum TeeFunction {
...
+        /// a0 = guest id
+        guest_id: u64,
+        /// a1 = guest physical base address of the shared range
+        guest_addr: u64,
+        /// a2 = page size (must be 4K for now)
+        page_type: TsmPageType,
```
abrestic-rivos, line 751: Do we care about the page type at this point? The range is unpopulated until the host goes to insert a page, at which point it can choose what page size to use.

sbi/src/sbi.rs
```diff
@@ -738,6 +738,35 @@ pub enum TeeFunction {
...
+        /// a3 = number of pages
+        num_pages: u64,
+    },
+    /// Maps non-confidential shared pages in the range previously defined by `TsmDefineSharedPageRange`.
+    /// The call may be made before or after TVM finalization, and shared pages can be mapped-in on demand.
+    /// a6 = 18
+    TsmInsertSharedPages {
```
abrestic-rivos, line 758: Same comment as above re: `Tsm` vs `Tvm` prefix.

src/vm.rs
```diff
@@ -946,6 +962,52 @@ impl<T: GuestStagePageTable> Vm<T, VmStateFinalized> {
...
+        page_type: sbi::TsmPageType,
+        num_pages: u64,
+        guest_addr: u64,
+    ) -> sbi::Result<u64> {
+        if page_type != TsmPageType::Page4k || num_pages == 0 {
+            return Err(SbiError::InvalidParam);
+        }
+        let page_addr = self.guest_addr_from_raw(page_addr)?;
+        let guest = self.guest_by_id(guest_id)?;
+        if let Some(guest_vm) = guest.as_finalized_vm() {
```
abrestic-rivos, line 999: Do we want to the host to pre-fault shared pages?

src/vm_pages.rs
```diff
@@ -578,6 +615,39 @@ impl Drop for PtePagePool {
     }
 }
 
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+struct PageRangeInfo {
```
abrestic-rivos, line 619: Thoughts on using the `VmRegionList` I just added to support shared page ranges instead? It supports multiple regions and RAII-based error recovery, which would simplify the implementation.

src/vm_pages.rs
```diff
@@ -595,6 +665,8 @@ pub struct VmPages<T: GuestStagePageTable, S = VmStateFinalized> {
     root: PlatformPageTable<T>,
     measurement: Mutex<Sha256Measure>,
     pte_pages: PtePagePool,
+    confidential_page_range: Mutex<Option<PageRangeInfo>>,
```
abrestic-rivos, line 668: Limiting a TVM to a single contiguous confidential page range seems overly restrictive.

src/vm_pages.rs
```diff
@@ -826,6 +898,52 @@ impl<T: GuestStagePageTable> VmPages<T, VmStateFinalized> {
...
+            let shared_list = from
+                .root
+                .get_shareable_range::<Page<Shareable>>(from_addr, PageSize::Size4k, num_pages)
+                .map_err(|_| Error::SharedPageNotMapped)?;
+
+            let mapper = VmPagesMapper::new_shareable(self, guest_addr, num_pages)?;
+            (shared_list, mapper)
+        };
+
+        for (page, addr) in shared_list.zip(guest_addr.iter_from()) {
```
abrestic-rivos, line 941: The state transition from `Mapped` to `Shared` (or bumping of the `Shared` refcount) probably shouldn't happen until we've done all the setup work above. Something like:

```
for (page, addr) in shared_list.zip(guest_addr.iter_from()) {
    let shared = page_tracker.share_page(page).unwrap();
    mapper.map_page(addr, shared).unwrap();
}
```

test-workloads/src/bin/tellus.rs
```diff
@@ -253,35 +268,48 @@ extern "C" fn kernel_init(hart_id: u64, fdt_addr: u64) {
         }
     }
 
-    // Add zeroed (non-measured) pages
-    // TODO: Make sure that these guest pages are actually zero
-    convert_pages(next_page, NUM_GUEST_ZERO_PAGES);
-    let msg = SbiMessage::Tee(sbi::TeeFunction::TvmAddZeroPages {
+    let shared_pages_base = fdt.base_addr() as u64 - 4096;
```
abrestic-rivos, line 271: Sticking the shared page range right before our FDT seems a bit odd.

abrestic-rivos (changes requested):

riscv-page-tables/src/page_table.rs
```diff
@@ -575,6 +576,35 @@ impl<T: PagingMode> PageTableInner<T> {
         }
     }
 
+    /// Locks the leaf PTE mapping `vaddr`, filling in any missing intermediate page tables
+    /// using `get_pte_page`. Since shared pages can be remapped to a different SPA, the
+    /// entry might already be valid
+    fn lock_4k_leaf_for_shared_mapping(
```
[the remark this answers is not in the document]
atulkharerivos: This was an artifact of the merge. I think the original functions were different, but I have really lost track now with the frequent re-basing :).

src/vm.rs
atulkharerivos: The original implementation did have it, but I removed it since I thought it probably makes sense to just say that shared pages are faulted-in on demand, but can always add it back.

src/vm_pages.rs
atulkharerivos: Yes, noticed it too, but haven't really internalized the purpose.

src/vm_pages.rs
atulkharerivos: I am open to suggestions, but think that we do need to have a single contiguous range to be able to efficiently search the GPA-map when DefineSharedPagesRange is called. Basically, we want to quickly rule out the possibility that the shared page range doesn't overlap with any confidential address space. To give an extreme example, suppose that the host defines a confidential page at the bottom and top of the GPA, and then tries to define a shared address space somewhere in the middle. Unless we maintain some sort of sparse bit vector to allow us to identify where confidential pages are mapped, it's computationally expensive to perform a search of the GPA space.

page-tracking/src/page_info.rs
```diff
@@ -138,7 +146,7 @@ impl PageInfo {
     pub fn release(&mut self) -> PageTrackingResult<PageOwnerId> {
         use PageState::*;
         match self.state {
-            Mapped | VmState | Converted | Converting(_) => {
+            Mapped | VmState | Converted | Converting(_) | Shared => {
```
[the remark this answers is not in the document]
atulkharerivos: Yes, there was an exception that got lost in the merge.

sbi/src/sbi.rs
atulkharerivos: The intent was to able to compute the upper bound, but we can just pass it directly.

page-tracking/src/page_info.rs
```diff
@@ -28,6 +28,13 @@ pub enum PageState {
     /// Page is mapped into the address space of the current owner.
     Mapped,
 
+    /// Page is mapped into the address space of the current owner, but has been shared with
+    /// a child-VM. Shared pages cannot be converted
+    Shared,
```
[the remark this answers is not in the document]
rsahita: Why should the TSM need to track the sharing with other guests - once a page is converted into a non-confidential page, it is under the VMMs purview to share as needed; for the page to be repurposed, it will have to be converted to a confidential (unassigned) page and at that point the exclusive property can be enforced.

src/vm.rs
rsahita: @abrestic-rivos What do you mean by pre-fault - do you mean pre-populate?

page-tracking/src/page_info.rs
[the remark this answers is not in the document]
abrestic-rivos: > for the page to be repurposed, it will have to be converted to a confidential (unassigned) page and at that point the exclusive property can be enforced.

That is exactly the point of reference counting how many guests have mapped the shared page. Otherwise, how does the TSM, upon receiving a request to convert a page to confidential memory, enforce that the page is no longer shared?

src/vm.rs
abrestic-rivos: Yup

riscv-pages/src/state.rs
atulkharerivos: Possibly, but likely doesn't belong in this MR?

test-workloads/src/bin/tellus.rs
```diff
@@ -344,6 +374,33 @@ extern "C" fn kernel_init(hart_id: u64, fdt_addr: u64) {
...
+                            page_type: sbi::TsmPageType::Page4k,
+                            num_pages: 1,
+                            guest_addr: fault_addr,
+                        });
+                        // Safety: `TvmAddSharedPages` only touches pages owned by us
+                        unsafe {
+                            ecall_send(&msg).expect("Tellus -- TvmAddSharedPages failed");
+                        }
+
+                        // Safety: We own the page, and are writing a value expected by the guest
```
dgreid, line 396: This is giving up exclusive ownership of the page. It would be nice if the comment mentioned that and that tellus won't access it again, or must use volatile, shared memory access patterns.

dgreid (commented):
Thanks. Took a quick look and have a few questions. Will leave all the state transition reviews to agrestic.

Also the first commit needs a the commit message updated (GitHub has the same problem as Gitlab about not allowing comments on commit messages...)

test-workloads/src/bin/tellus.rs
```diff
@@ -344,6 +374,33 @@ extern "C" fn kernel_init(hart_id: u64, fdt_addr: u64) {
...
+                    sbi::TvmCpuExitCode::SharedPageFault => {
+                        // Figure out where the fault occurred.
+                        let fault_addr = get_fault_address(vmid);
+                        if fault_addr != SHARED_PAGES_START_ADDRESS {
+                            println!("Fault in shared page region at {fault_addr:x}");
+                            panic!("Unexpected shared page fault address");
+                        }
+                        let msg = SbiMessage::Tee(sbi::TeeFunction::TvmAddSharedPages {
+                            guest_id: vmid,
+                            page_addr: next_page,
```
abrestic-rivos, line 386: I'd save the address of the shared page in a separate variable in case someone else inserts code in between that bumps `next_page`.

abrestic-rivos (commented):
Thanks, getting closer. I think we can clean up the state transitions a bit so that it looks more like the existing page assignment flow, but otherwise mostly nits.

page-tracking/src/page_tracker.rs
```diff
@@ -220,7 +229,8 @@ impl PageTracker {
     pub fn release_page_by_addr(&self, addr: SupervisorPageAddr, owner: PageOwnerId) -> Result<()> {
         let mut page_tracker = self.inner.lock();
         let info = page_tracker.get_mut(addr)?;
-        if info.owner() != Some(owner) {
+        // Shared pages might be owned by the parent
```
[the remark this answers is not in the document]
abrestic-rivos: Right, it shouldn't be popping the owner in the shared case, just decrementing the refcount; I missed that `PageInfo::release()` was popping the owner even for shared pages.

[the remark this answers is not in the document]
atulkharerivos: Yes, I realized the exact same thing, and was discussing it on the phone with Andrew :).

page-tracking/src/page_info.rs
```diff
@@ -248,10 +302,42 @@ impl PageInfo {
                 self.state = Converted;
```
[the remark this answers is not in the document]
atulkharerivos: Done/

src/vm_pages.rs
abrestic-rivos: Not resolved; see my other comment about splitting this into two steps so that we don't bump the refcount until we've guaranteed that the page will actually get mapped.

page-tracking/src/page_info.rs
```diff
@@ -28,6 +28,10 @@ pub enum PageState {
     /// Page is mapped into the address space of the current owner.
     Mapped,
 
+    /// Page is mapped into the address space of the current owner, but has been shared with
+    /// a child-VM. Shared pages cannot be converted
```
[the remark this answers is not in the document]
atulkharerivos: Done.

page-tracking/src/page_info.rs
```diff
@@ -147,13 +181,23 @@ impl PageInfo {
                     Ok(owner)
                 }
             }
+            Shared(rc) => {
+                // Shared pages start with a RC of 1, so RC is always > 0 here
+                if let Some(rc) = rc.checked_sub(1) {
+                    self.state = if rc == 0 { Mapped } else { Shared(rc) };
+                    // Unwrap ok: a Shared page must have an owner
+                    Ok(self.owner().unwrap())
```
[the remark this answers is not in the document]
atulkharerivos: Actually, it's used by the unit test, so just left it alone.

riscv-page-tables/src/page_table.rs
```diff
@@ -797,6 +798,36 @@ impl<T: PagingMode> PlatformPageTable<T> {
...
+    ) -> Result<LockedPageList<P>> {
+        if page_size.is_huge() {
+            return Err(Error::PageSizeNotSupported(page_size));
+        }
+
+        let mut inner = self.inner.lock();
+        let mut pages = LockedPageList::new(inner.page_tracker.clone());
+        for a in addr.iter_from().take(num_pages as usize) {
+            let entry = inner.get_mapped_4k_leaf(a, P::mem_type())?;
+            // Unwrap ok, PFN must have been properly aligned in order to have been mapped.
```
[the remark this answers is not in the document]
atulkharerivos: Deleted.

src/vm_pages.rs
```diff
@@ -218,15 +221,19 @@ impl VmRegionList {
...
-        if let Some(ref mut before) = regions.get_mut(index - 1) {
-            if before.end == start && before.region_type == region_type {
-                before.end = end;
-                start = before.start;
-                regions.remove(index);
-                index -= 1;
+        // Avoid the potential underflow in the index - 1 expression below
+        if index > 0 {
+            // Coalesce with same-typed regions before and after this one, if possible.
+            if let Some(ref mut before) = regions.get_mut(index - 1) {
```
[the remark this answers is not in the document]
atulkharerivos: Done/

src/vm_pages.rs
```diff
@@ -588,6 +600,30 @@ impl<T: GuestStagePageTable, S> VmPages<T, S> {
...
+            .root
+            .get_shareable_range::<Page<Shareable>>(from_addr, PageSize::Size4k, num_pages)
+            .map_err(|_| Error::SharedPageNotMapped)?;
+
+        let mapper =
+            VmPagesMapper::new_in_region(self, guest_addr, num_pages, VmRegionType::Shared)?;
+
+        for (page, addr) in shared_list.zip(guest_addr.iter_from()) {
+            // Unwrap OK since we have exclusive ownership and have already filled in the PTE
+            mapper.map_page(addr, page).unwrap();
```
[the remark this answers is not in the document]
atulkharerivos: Aggh...I had an explicit call to unlock_page(), but it got lost in the shuffle. It's been added back.

page-tracking/src/page_tracker.rs
```diff
@@ -298,6 +324,55 @@ impl PageTracker {
...
+        // Safe since we've taken exclusive ownership of the page, verified its typing.
+        // TODO: Page size
+        Ok(unsafe { P::new(addr) })
+    }
+
+    /// Bumps the reference count on a the shareable page at `addr` if it's owned by owner, in Mapped or Shared state, and has a
+    /// current reference count < u64::MAX. The exclusive reference to the page is released following a successful share.
+    pub fn share_page(
+        &self,
+        addr: SupervisorPageAddr,
```
abrestic-rivos, line 361: I've been trying to avoid `PageTracker` APIs that operate on raw addresses and instead have been using `PhysPage` when possible since that implies exclusive ownership, whereas a raw address type does not imply anything about ownership. Can we make this take, say, a `ShareablePhysPage` and return a `MappablePhysPage` in a similar manner to `reclaim_page()` or `assign_page_*()`?

page-tracking/src/page_tracker.rs
```diff
@@ -298,6 +324,55 @@ impl PageTracker {
...
+        match info.state() {
+            // Ensure that the share RC can be incremented before attempting to obtain the lock.
+            PageState::Shared(rc) if rc == u64::MAX => {
+                return Err(Error::PageNotShareable);
+            }
+            PageState::Shared(_) | PageState::Mapped => {}
+            _ => {
+                return Err(Error::PageNotShareable);
+            }
+        };
```
abrestic-rivos, line 349: nit: You could put this in a `PageInfo::is_shareable()` helper, similar in spirit to the existing `is_convertible()` helper.

page-tracking/src/page_tracker.rs
```diff
@@ -274,8 +286,22 @@ impl PageTracker {
...
-    pub fn put_converted_page<P: ConvertedPhysPage>(&self, page: P) -> Result<()> {
+    /// Returns the lock status of the page
+    pub fn is_locked_page(&self, addr: SupervisorPageAddr) -> Result<bool> {
+        let mut page_tracker = self.inner.lock();
+        let info = page_tracker.get_mut(addr)?;
+        Ok(info.is_locked())
+    }
+
+    /// Obtains an exclusive reference on an assignable page
+    pub fn lock_for_assignment(&self, addr: SupervisorPageAddr) -> Result<()> {
```
abrestic-rivos, line 297: Why do we need public APIs to lock a page or check if a page is locked? There's only two cases where we need to lock a page presently: when acquiring a reference to a converted page (for assignment/reclaim) or acquiring a reference to a mapped page (for sharing); can we keep the locking internal to `get_converted_page()` and `get_shareable_page()`?

Also, see my API below on trying to avoid `PageTracker` APIs that operate on raw addresses.

src/vm.rs
```diff
@@ -1155,6 +1227,11 @@ impl<T: GuestStagePageTable> HostVm<T, VmStateInitializing> {
                 vm_addr.bits() & (T::TOP_LEVEL_ALIGN - 1),
                 page.addr().bits() & (T::TOP_LEVEL_ALIGN - 1)
             );
+            // Unwrap ok: the page is valid
+            if !page_tracker.is_locked_page(page.addr()).unwrap() {
```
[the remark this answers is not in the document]
atulkharerivos: It's necessitated by the check for the new check for is_locked() in assign() (to get rid of ConvertedLock). The code path fails without this check, and while we can move things elsewhere, I do see the need to lock the page in some flows (but not unconditionally).

page-tracking/src/collections/page_box.rs
```diff
@@ -279,6 +279,7 @@ mod tests {
             .by_ref()
             .take(1)
             .map(|p| {
+                page_tracker.lock_for_assignment(p.addr()).unwrap();
```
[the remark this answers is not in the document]
atulkharerivos: It's necessitated by the check for the new check for is_locked() in assign() (to get rid of ConvertedLock). The code path fails without this check, and while we can move things elsewhere, I do see the need to lock the page in some flows (but not unconditionally).

page-tracking/src/page_tracker.rs
atulkharerivos: It all came down to the same underlying cause -- pages allocated by HyperAlloc() were in Converted state, but not locked. Adding a lock() there obviates the need for public APIs to check for and lock pages in the tracker.

page-tracking/src/page_info.rs
```diff
@@ -226,41 +277,63 @@ impl PageInfo {
...
-                self.state = Converted;
+            Converted | Mapped | Shared(_) => self.unlock_page(),
+            _ => Err(PageTrackingError::PageNotLocked),
+        }
+    }
+
+    /// Transitions a locked page from Mapped to Shared, or increments the RC on a Shared page.
+    pub fn share_page(&mut self) -> PageTrackingResult<()> {
+        use PageState::*;
+        if !self.locked {
```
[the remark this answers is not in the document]
atulkharerivos: The lock was being dropped in the outer share() in page_tracker, but will remove it completely.

page-tracking/src/page_info.rs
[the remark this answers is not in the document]
atulkharerivos: Yeah, I kept it around because it covered the popping of owners, but nuked it now.

src/vm.rs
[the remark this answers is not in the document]
atulkharerivos: Removed...HyperAlloc has been modified to lock Converted pages.

page-tracking/src/page_info.rs
```diff
@@ -134,11 +142,40 @@ impl PageInfo {
         self.mem_type
     }
 
+    /// Returns if the page is locked.
+    pub fn is_locked(&self) -> bool {
+        self.locked
+    }
+
+    /// Obtains an exclusive reference to the page.
+    pub fn lock_page(&mut self) -> PageTrackingResult<()> {
```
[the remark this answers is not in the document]
atulkharerivos: Changed.

page-tracking/src/page_info.rs
```diff
@@ -226,41 +277,63 @@ impl PageInfo {
...
         match self.state {
-            ConvertedLocked => {
-                self.state = Converted;
+            Converted | Mapped | Shared(_) => self.unlock_page(),
+            _ => Err(PageTrackingError::PageNotLocked),
+        }
+    }
+
+    /// Transitions a locked page from Mapped to Shared, or increments the RC on a Shared page.
+    pub fn share_page(&mut self) -> PageTrackingResult<()> {
```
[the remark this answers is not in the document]
atulkharerivos: Changed.

page-tracking/src/collections/page_box.rs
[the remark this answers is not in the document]
atulkharerivos: Removed; changed HyperAlloc.

page-tracking/src/page_tracker.rs
atulkharerivos: Changed.

page-tracking/src/page_tracker.rs
```diff
@@ -298,6 +326,25 @@ impl PageTracker {
         }
     }
 
+    /// Acquires an exclusive reference to the shareable page at `addr` if it's owned by owner,
+    /// and in Mapped or Shared state.
+    pub fn get_shareable_page<P: PhysPage>(
```
abrestic-rivos, line 331: Shouldn't `P` be a `ShareablePhysPage`?

page-tracking/src/page_info.rs
```diff
@@ -134,29 +142,68 @@ impl PageInfo {
         self.mem_type
     }
 
+    /// Returns if the page is locked.
+    pub fn is_locked(&self) -> bool {
```
abrestic-rivos, line 146: Does this need to be `pub`?

page-tracking/src/page_info.rs
```diff
@@ -134,29 +142,68 @@ impl PageInfo {
...
+        self.locked
+    }
+
+    /// Obtains an exclusive reference to the page.
+    fn lock_page(&mut self) -> PageTrackingResult<()> {
```
abrestic-rivos, line 151: See my previous comment about the naming of these helpers: it's not clear what the difference is between `lock_page` vs `lock` based on the name. I'm not even sure you really need them anymore (i.e. you could just use `self.locked` directly), but if you do want to keep them I'd rename them to something like `do_lock` / `do_unlock`.

page-tracking/src/page_tracker.rs
```diff
@@ -193,6 +204,22 @@ impl PageTracker {
         Ok(unsafe { P::MappablePage::new_with_size(page.addr(), page.size()) })
     }
 
+    /// Consumes the shared `page`, returning a page that can then be mapped into
+    /// a page table.
+    pub fn share_page_for_mapping<P>(&self, page: P, owner: PageOwnerId) -> Result<P::MappablePage>
```
abrestic-rivos, line 209: nit: just `share_page()`. the `for_mapping` is redundant since all we can do with shared pages is map them.

page-tracking/src/page_tracker.rs
```diff
@@ -193,6 +204,22 @@ impl PageTracker {
...
+    where
+        P: AssignablePhysPage<MeasureOptional>,
```
abrestic-rivos, line 211: Shouldn't `P` be a `ShareablePhysPage`?

riscv-pages/src/page.rs
```diff
@@ -393,6 +393,9 @@ pub trait AssignablePhysPage<M: MeasureRequirement>: ConvertedPhysPage {
 /// by unmapping a previously-mapped page in a VM's address space.
 pub trait InvalidatedPhysPage: PhysPage {}
 
+/// Trait representing a mapped page that's eligible for conversion to a Shared state
+pub trait ShareablePhysPage: PhysPage {}
```
abrestic-rivos, line 397: Add an associated type for the result of the transition to a shared & mappable page?

e.g.

```
pub trait ShareablePhysPage: PhysPage {
    type MappablePage: MappablePhysPage<MeasureOptional>;
}
```

And then, if you take my suggestion on type names, you'd have:

```
impl ShareablePhysPage for Page<Mapped> {
    type MappablePage = Page<MappableShared>;
}
```

riscv-pages/src/state.rs
```diff
@@ -111,3 +111,18 @@ impl InternalData for InternalDirty {}
...
+#[derive(Debug)]
+pub enum SharedDirty {}
+
+/// A mapped page that is eligible for conversion to a Shared page.
+#[derive(Debug)]
+pub enum Shareable {}
+impl State for Shareable {}
+impl Converted for Shareable {}
+impl State for SharedDirty {}
+impl Mappable<MeasureOptional> for SharedDirty {}
```
abrestic-rivos, line 125: nit: I'd call this state `MappableShared` or (`MappableDirty`) rather than `SharedDirty`. Shared pages will always be dirty; the point we're trying to convey with this type name is that it's a shared page and it is now ready to be mapped.

riscv-pages/src/state.rs
```diff
@@ -111,3 +111,18 @@ impl InternalData for InternalDirty {}
...
+impl Assignable<MeasureOptional> for Shareable {
```
abrestic-rivos, line 126: I don't think `Assignable` makes sense here; ownership isn't changing for shared pages.

page-tracking/src/page_tracker.rs
```diff
@@ -298,6 +308,28 @@ impl PageTracker {
         }
     }
 
+    /// Acquires an exclusive reference to the shareable page at `addr` if it's owned by owner,
+    /// and in Mapped or Shared state
+    pub fn lock_for_shareable_assignment(
```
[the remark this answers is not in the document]
atulkharerivos: Believe it's resolved per our conversation.

page-tracking/src/page_tracker.rs
atulkharerivos: Changed.

page-tracking/src/page_info.rs
atulkharerivos: Removed.

page-tracking/src/page_tracker.rs
atulkharerivos: Changed.

page-tracking/src/page_tracker.rs
atulkharerivos: Changed.

riscv-pages/src/page.rs
atulkharerivos: Done. FWIW, I am finding it extremely difficult to infer the intended purpose of various traits from the code. It's possible that I missed it, but a written documentation explaining the purpose of each of these traits and how are are associated will be very useful.

src/vm_pages.rs
atulkharerivos: Believe this is resolved per our conversation.

page-tracking/src/page_info.rs
atulkharerivos: Removed.

riscv-pages/src/state.rs
atulkharerivos: I think this has been resolved, but as mentioned below, it has been very difficult for me to infer the intent of the various traits and the interactions from the code.

page-tracking/src/page_info.rs
atulkharerivos: Apologies...the gibhub interface isn't easy to navigate when there are several comments. Should be deleted for sure now.

page-tracking/src/page_info.rs
```diff
@@ -226,41 +262,79 @@ impl PageInfo {
...
+                    Err(PageTrackingError::PageLocked)
+                }
             }
-            ConvertedLocked => Err(PageTrackingError::PageLocked),
             _ => Err(PageTrackingError::PageNotAssignable),
         }
     }
 
-    /// Drops the exclusive lock on the a converted page.
+    /// Drops the exclusive lock on an assignable page.
```
abrestic-rivos, line 298: "on an assignable or shareable page"

page-tracking/src/page_info.rs
```diff
@@ -134,29 +142,53 @@ impl PageInfo {
         self.mem_type
     }
 
+    /// Releases an exclusive reference to the page.
+    fn unlock_page(&mut self) -> PageTrackingResult<()> {
```
abrestic-rivos, line 146: Similar to the previous `lock()` vs `lock_page()` comment, I think you can get rid of `unlock_page()` entirely.

abrestic-rivos (approved):

page-tracking/src/page_info.rs
```diff
@@ -226,41 +262,79 @@ impl PageInfo {
...
         use PageState::*;
         match self.state {
-            ConvertedLocked => {
-                self.state = Converted;
+            Converted | Mapped | Shared(_) => self.unlock_page(),
+            _ => Err(PageTrackingError::PageNotLocked),
+        }
+    }
+
+    /// Transitions a locked page from Mapped to Shared, or increments the RC on a Shared page.
```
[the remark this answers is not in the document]
atulkharerivos: Done.

page-tracking/src/page_info.rs
```diff
@@ -226,41 +262,79 @@ impl PageInfo {
...
+        }
+    }
+
+    /// Returns if the page is Shared
+    pub fn is_shared(&self) -> bool {
+        use PageState::*;
+        matches!(self.state, Shared(_))
+    }
+
+    /// Obtains an exclusive reference on a page in an assignable state in preparation
```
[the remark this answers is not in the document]
atulkharerivos: Changed.

PR #11, opened 2022-07-01
Rework guest memory access in preparation for MMIO

abrestic-rivos:
Since we'll need to be more permissive about which faults can be fixed up when fetching guest instructions (due to using guest virtual addresses), let's constrain recoverable faults to specific instructions which are allowed to fault. Taking some inspiration from Linux, this series creates an exception fixup table which holds the addresses of such instructions.


dgreid (approved):

atulkharerivos (commented):
LGTM.

PR #13, opened 2022-07-05
Start PCI support

dgreid:
This series adds _very_ basic PCI functionality. It's not enough to be immediately useful, the goal is to provide a starting point for working toward PCI device forwarding. It's broken in to three parts.

## Patches 1-4:

Pulls two crosvm crates and converts them for use with no_std. Having the endian types makes dealing with non-native device formats easier. These can be broken out to separate repositories once we have that ability smoothed out in github and CI.

## Patches 5-6:

Preparation of existing code. Adds some helpers and a memory type for use by PCI memory.

## Patches 7-10:

The real PCI work. Adds some basic types for addresses and headers.
Uses those types to do a PCI scan based on the hypervisor device tree.
Invokes that scan from main and prints devices to check that it's working end-to-end.


drivers/src/pci/config_space.rs
```diff
@@ -0,0 +1,296 @@
...
+        if header_addr.function() != Function::default() {
+            return 0..0;
+        }
+        self.header_for(header_addr)
+            .map(|header| if header.multi_function() { 0..8 } else { 0..1 })
+            .unwrap_or(0..0)
+    }
+
+    // Returns the offset of the given address within a PciConfigSpace starting at
+    // `config_bus_start`.
```
abrestic-rivos, line 153: "... of the given address within this PciConfigSpace" ?

dgreid: yes, applied. Thanks!

drivers/src/pci/header.rs
```diff
@@ -0,0 +1,229 @@
...
+}
+
+// The MSB of the header type byte indicates multi-funciton, the lower 7 are the type.
+const HEADER_TYPE_MASK: u8 = 0x7f;
+const HEADER_MULTI_FUNCTION_MASK: u8 = 0x80;
+
+// Index of words in the PCI Configuration header for a function.
+enum HeaderWord {
+    Vendor = 0,
+    //Status = 1,
```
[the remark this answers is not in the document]
dgreid: yes, will remove.

PR #17, opened 2022-07-11
Trap on WFI in guest VMs

abrestic-rivos:
The purpose of this series is mostly to serve as a template for handling virtual instruction exceptions, which we'll eventually need for emulating access to S-level CSRs. The WFI case is pretty simple: just bump SEPC and exit to the parent VM.


atulkharerivos (approved):
LGTM.

PR #18, opened 2022-07-12
TVM measurement

sameo:
This PR brings the initial TVM measurement building and evidence reporting.

Each TVM is assigned a set of measurement through an attestation manager. Salus calls into the manager to extend the TVM measurement for each measured page. The measurement extends a specific register for guest/OS measured pages.


attestation/src/certificate.rs
```diff
@@ -125,10 +126,11 @@ impl<'a> Certificate<'a> {
     /// @cdi_id: The CDI identifier of the certificate issuer (e.g. the TSM).
     /// @key_pair: The ED25519 key pair generated from the CDI.
     /// @certificate_buffer: A buffer to hold the DER encoded certificate data.
```
abrestic-rivos, line 128: nit: update the function doc comment?

src/vm.rs
```diff
@@ -195,11 +199,14 @@ impl From<EcallResult<u64>> for EcallAction {
...
 }
 
+type AttestationSha384 = AttestationManager<sha2::Sha384>;
+
 /// A VM that is being run.
 pub struct Vm<T: GuestStagePageTable, S = VmStateFinalized> {
     vcpus: VmCpus,
     vm_pages: VmPages<T, S>,
     guests: Option<Guests<T>>,
+    attestation_mgr: Mutex<AttestationSha384>,
```
abrestic-rivos, line 206: Might it be cleaner to move the mutex inside `AttestationManager`? That way you don't have to do `&mut self.attestation_mgr.lock()` everywhere.

src/vm_pages.rs
```diff
@@ -281,15 +283,14 @@ impl VmRegionList {
 
 /// Wrapper for a `PageTableMapper` created from the page table of `VmPages`. Measures pages as
 /// they are inserted, if necessary.
-pub struct VmPagesMapper<'a, T: GuestStagePageTable, S> {
+pub struct VmPagesMapper<'a, T: GuestStagePageTable> {
     mapper: PageTableMapper<'a, T>,
-    vm_pages: &'a VmPages<T, S>,
 }
 
-impl<'a, T: GuestStagePageTable, S> VmPagesMapper<'a, T, S> {
```
abrestic-rivos, line 289: Why remove the state type parameter (`S`)? It was there to enforce (via the type system) that we didn't add measured pages to a running VM.

test-workloads/src/bin/guestvm.rs
```diff
@@ -128,11 +133,43 @@ extern "C" fn kernel_init(_hart_id: u64, shared_page_addr: u64) {
             panic!("Cert parsing error");
         }
         Ok(cert) => {
```
abrestic-rivos, line 114: nit: To reduce the amount of nesting, maybe do something like:

```
let cert = Certificate::from_der(...).expect("Cert parsing error");
let mut tcb_info_extn = ...;
...
```

test-workloads/src/bin/guestvm.rs
```diff
@@ -128,11 +133,43 @@ extern "C" fn kernel_init(_hart_id: u64, shared_page_addr: u64) {
...
+            let mut tcb_info_extn = DiceTcbInfo::default();
+            cert.tbs_certificate.extensions.as_ref().map(|extensions| {
+                for extn in extensions.iter() {
+                    if extn.extn_id != TCG_DICE_TCB_INFO {
+                        continue;
+                    }
+
+                    tcb_info_extn = match DiceTcbInfo::from_der(extn.extn_value) {
```
abrestic-rivos, line 122: Along the same lines as the above, it would be a bit more concise to use `expect()` rather than matching against the `Result<>`.

src/vm.rs
```diff
@@ -1232,8 +1243,14 @@ impl<T: GuestStagePageTable> Vm<T, VmStateFinalized> {
...
-        let cert_bytes =
-            Certificate::from_csr(&csr, cdi, &key_pair, &mut cert_bytes_buffer).unwrap();
+        let cert_bytes = Certificate::from_csr(
+            &csr,
+            cdi,
+            &key_pair,
+            &self.attestation_mgr.lock(),
+            &mut cert_bytes_buffer,
+        )
+        .unwrap();
```
atulkharerivos, line 1162: Could use a comment about why unwrap() is OK, or else change to unwrap_or_else, etc.

src/vm.rs
```diff
@@ -1411,7 +1428,13 @@ impl<T: GuestStagePageTable> HostVm<T, VmStateInitializing> {
...
                 .assign_page_for_mapping(page, self.inner.page_owner_id())
                 .unwrap();
-            mapper.map_page_with_measurement(vm_addr, mappable).unwrap();
+            mapper
+                .map_page_with_measurement(
+                    vm_addr,
+                    mappable,
+                    &mut self.inner.attestation_mgr.lock(),
+                )
+                .unwrap();
```
atulkharerivos, line 1346: Could use a comment about why unwrap() is OK, or else change to unwrap_or_else, etc.

Cargo.toml
```diff
@@ -14,11 +14,14 @@ panic = "abort"
 arrayvec = { version = "0.7.2", default-features = false }
 assertions = { path = "./assertions" }
 attestation = { path = "./attestation" }
+const-oid = { version = "0.9.0", features = ["db"] }
```
atulkharerivos, line 17: Not related to this, but nit about commit message: "Each VM use" should be "Each VM uses"

attestation/src/lib.rs
```diff
@@ -39,6 +39,9 @@ pub enum Error {
     /// Invalid DER payload
     InvalidDer(der::Error),
 
+    /// Invalid X.509 extension DER
```
atulkharerivos, line 42: Not related, but pedantic nit about commit message: "The API is minimal for now, it" should be "The API is minimal for now, **and** it"

attestation/Cargo.toml
```diff
@@ -8,7 +8,9 @@ edition = "2021"
 arrayvec = { version = "0.7.2", default-features = false }
 const-oid = { version = "0.9.0", features = ["db"] }
 der = { version = "0.6.0", features = ["derive", "oid"] }
+digest = {version = "0.10.3", default-features = false }
```
atulkharerivos, line 11: Not related, but commit message "It allows for _carrying publish_ a TCB evidence as a certificate custom
extension." seems off.

attestation/src/certificate.rs
```diff
@@ -130,7 +130,7 @@ impl<'a> Certificate<'a> {
         cdi_id: &'a [u8],
         key_pair: &'a [u8],
         certificate_buf: &'a mut [u8],
-    ) -> Result<'a, &'a [u8]> {
+    ) -> Result<&'a [u8]> {
```
atulkharerivos, line 135: Nit: Consider rephrasing the commit message "That allows for removing the lifetime bound from the crate Error
enum." as "This eliminates the lifetime bound from...".

atulkharerivos (commented):
LGTM to me as well, modulo Andrew's comments, and some minor nits.

src/vm_pages.rs
sameo: Sorry I missed that. I added the state type parameter back, through a phantom field.

src/vm.rs
sameo: The `unwrap()` was there already, but I replaced it with a proper error mapping now. Thanks.

src/vm.rs
sameo: The whole `add_measured_pages()` method (together with a few more from the `HostVm` implementation do not carry errors and unwrap. I followed the same pattern here, I think this should be handled through a more general cleanup for those functions, as a separate PR.

abrestic-rivos: Yeah, when building the `HostVm` we just unwrap since there isn't anything we can do about the errors.

abrestic-rivos (approved):

PR #22, opened 2022-07-14
Discover PCI BAR resources

abrestic-rivos:
Probe for the memory and IO BAR resources of the PCIe root complex from the device tree and add them to the memory map. While we'll need to trap-and-emulate the config space for the host VM, BAR spaces can be directly mapped into its address space.


drivers/src/pci/root.rs
```diff
@@ -78,7 +78,7 @@ impl PcieRoot {
             // Any overlaps will be caught by `add_mmio_region` and the error will be propagated.
             mem_map
                 .add_mmio_region(
-                    DeviceMemType::PciRoot,
+                    DeviceMemType::PciConfig,
```
atulkharerivos, line 150: Not related, but commit message needs a small tweak: "...since we're using **_it_** to...".

drivers/src/pci/config_space.rs
```diff
@@ -38,19 +38,18 @@ impl PciConfigSpace {
         }
     }
 
-    /// Returns an iterator across the top-level buses.
-    pub fn busses(&self) -> HostControllersIter {
+    /// Returns the top-level bus.
```
atulkharerivos, line 41: Not related, but in the commit message, "buses" might be a more common plural form.

atulkharerivos (approved):
LGTM; just some minor nits about commit messages.

dgreid (approved):
thanks!

drivers/src/pci/config_space.rs
abrestic-rivos: Done.

drivers/src/pci/root.rs
abrestic-rivos: Done.

PR #29, opened 2022-07-20
Derive the TVM CDI and attestation keys

sameo:
This PR uses a currently artificial TSM Compound Device Identifier (CDI) to derive a TVM specific CDI and an attestation key pair. The latter is used to signed TVM originated Certificate Signing Request (CSR) certificates carrying the TVM attestation evidence.

The TSM CDI will eventually be received from the TSM driver, as the TSM underlying DICE TCB layer (The TSM driver is responsible for building and creating the TSM CDI).

Patches 1 and 2 are minor build and constant definition fixes.
Patch 3 adds the key/CDI derivation feature to the the attestation manager and passes a TVM unique identifier (Its `PageOwnerId`) to the VM constructor for the attestation manager to use when generating TVM specific CDI and key pairs.

### DIFF
--- a/attestation/src/extensions/pkix/authkeyid.rs
+++ b/attestation/src/extensions/pkix/authkeyid.rs
@@ -10,7 +10,7 @@ use const_oid::{AssociatedOid, ObjectIdentifier};
 use der::asn1::{OctetStringRef, UIntRef};
 use der::Sequence;
 
-pub(crate) const AUTH_KEY_ID_EXTENSION_LEN: usize = 32;
+pub(crate) const AUTH_KEY_ID_EXTENSION_LEN: usize = 64;
 
 /// AuthorityKeyIdentifier as defined in [RFC 5280 Section 4.2.1.1].
 ///
--- a/attestation/src/measurement.rs
+++ b/attestation/src/measurement.rs
@@ -3,10 +3,13 @@
 // SPDX-License-Identifier: Apache-2.0
 
 use arrayvec::ArrayVec;
-use const_oid::{db::rfc5912::ID_SHA_384, ObjectIdentifier};
+use const_oid::ObjectIdentifier;
+use core::marker::PhantomData;
 use der::Encode;
 use digest::{Digest, OutputSizeUser};
+use ed25519_dalek::{Keypair, SecretKey, SECRET_KEY_LENGTH};
 use generic_array::GenericArray;
+use hkdf::HmacImpl;
 use spin::RwLock;
 
 use crate::{
@@ -14,6 +17,7 @@ use crate::{
         dice::tcbinfo::{DiceTcbInfo, MAX_FWID_LEN, MAX_TCB_INFO_HEADER_LEN, TCG_DICE_TCB_INFO},
         Extension,
     },
+    kdf::{derive_attestation_cdi, derive_secret_key, extract_cdi},
     Error, Result,
 };
 
@@ -27,6 +31,12 @@ const MSMT_REGISTERS: usize = STATIC_MSMT_REGISTERS + DYNAMIC_MSMT_REGISTERS;
 
 /// Largest possible DiceTcbInfo extension length
 pub const MAX_TCB_INFO_EXTN_LEN: usize = (MSMT_REGISTERS * MAX_FWID_LEN) + MAX_TCB_INFO_HEADER_LEN;
+
+const CDI_LEN: usize = 32;
+
+/// The TVM CDI ID length
+/// The TVM CDI ID is derived from the TSM CDI.
+pub const CDI_ID_LEN: usize = 20;
 
 macro_rules! is_static_measurement {
     ($idx:ident) => {{
@@ -47,6 +57,13 @@ pub enum MeasurementIndex {
     /// Dynamic TVM pages use the first DRTM register (PCR17)
     TvmPageExtension = 4,
 }
+
+#[derive(Clone, Debug)]
+enum AttestationCdi {}
+#[derive(Clone, Debug)]
+enum CsrCdi {}
+#[derive(Clone, Debug)]
+enum SealingCdi {}
 
 const TCG_PCR_MAPPING: [u8; MSMT_REGISTERS] = [
     0,  // TCG PCR 0  - SRTM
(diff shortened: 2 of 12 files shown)


attestation/src/measurement.rs
```diff
@@ -28,6 +31,12 @@ const MSMT_REGISTERS: usize = STATIC_MSMT_REGISTERS + DYNAMIC_MSMT_REGISTERS;
 /// Largest possible DiceTcbInfo extension length
 pub const MAX_TCB_INFO_EXTN_LEN: usize = (MSMT_REGISTERS * MAX_FWID_LEN) + MAX_TCB_INFO_HEADER_LEN;
 
+const CDI_LEN: usize = 32;
+
+/// The TVM CDI IDlength
```
abrestic-rivos, line 36: nit: missing space between "ID" and "length"

attestation/src/measurement.rs
```diff
@@ -204,10 +251,38 @@ impl<'a, D: Digest> AttestationManager<D> {
...
+        let pk = (&sk).into();
+
+        Ok(Keypair {
+            public: pk,
+            secret: sk,
+        })
+    }
+
+    /// Encode the CDI ID into a UTF-8 slice.
+    pub fn encode_cdi_id(&'a self, output: &mut [u8]) -> Result<()> {
```
abrestic-rivos, line 285: Why is an explicit lifetime needed here?

src/vm.rs
```diff
@@ -234,8 +234,7 @@ impl<T: GuestStagePageTable> Vm<T, VmStateInitializing> {
                 // Fake compound device identifier (DICE CDI)
                 // TODO Get the CDI from e.g. the TSM driver.
                 b"THISISARANDOMCDI",
-                // TODO Use a TVM unique address, e.g. the TVM state one.
-                0,
+                state_addr,
```
abrestic-rivos, line 237: I'm not sure the guest physical address of the start of RAM really qualifies as unique since, by convention, RAM usually starts at 0x8000_0000 in the (guest) physical address space.

Does this identifier really need to be an address? Could you use, say, the PageOwnerId of the VM instead?

sameo: Yes, we can use the `PageOwnerId`. As long as it's unique per VM, it's good. Thanks for the suggestion.

src/vm.rs
```diff
@@ -226,11 +226,19 @@ impl<T: GuestStagePageTable, S> Vm<T, S> {
...
             guests: None,
-            attestation_mgr: AttestationSha384::default(),
+            attestation_mgr: AttestationSha384::new(
+                // Fake compound device identifier (DICE CDI)
+                // TODO Get the CDI from e.g. the TSM driver.
+                b"THISISARANDOMCDI",
+                vm_id,
+                const_oid::db::rfc5912::ID_SHA_384,
+            )
+            .expect("Failed to create attestation manager"),
```
abrestic-rivos, line 241: Is it possible for this to fail? If so, then we shouldn't be using `expect()`. If not, can you put a comment briefly explaining why it can't fail?

abrestic-rivos (approved):

attestation/src/extensions/pkix/authkeyid.rs
```diff
@@ -10,7 +10,7 @@ use const_oid::{AssociatedOid, ObjectIdentifier};
 use der::asn1::{OctetStringRef, UIntRef};
 use der::Sequence;
 
-pub(crate) const AUTH_KEY_ID_EXTENSION_LEN: usize = 32;
+pub(crate) const AUTH_KEY_ID_EXTENSION_LEN: usize = 64;
```
atulkharerivos, line 13: The commit message uses hex values with the 0x prefix.

attestation/src/measurement.rs
```diff
@@ -204,10 +251,38 @@ impl<'a, D: Digest> AttestationManager<D> {
...
+    /// This key pair will be used for signing certificates requests coming from TVMs.
+    #[allow(clippy::needless_borrow)]
+    pub fn key_pair(&self) -> Result<Keypair> {
+        // From the OpenDice implementation
+        let asym_salt: [u8; 64] = [
+            0x63, 0xB6, 0xA0, 0x4D, 0x2C, 0x07, 0x7F, 0xC1, 0x0F, 0x63, 0x9F, 0x21, 0xDA, 0x79,
+            0x38, 0x44, 0x35, 0x6C, 0xC2, 0xB0, 0xB4, 0x41, 0xB3, 0xA7, 0x71, 0x24, 0x03, 0x5C,
+            0x03, 0xF8, 0xE1, 0xBE, 0x60, 0x35, 0xD3, 0x1F, 0x28, 0x28, 0x21, 0xA7, 0x45, 0x0A,
+            0x02, 0x22, 0x2A, 0xB1, 0xB3, 0xCF, 0xF1, 0x67, 0x9B, 0x05, 0xAB, 0x1C, 0xA5, 0xD1,
+            0xAF, 0xFB, 0x78, 0x9C, 0xCD, 0x2B, 0x0B, 0x3B,
```
atulkharerivos, line 265: If this a an immutable const array, perhaps it doesn't need to be allocated on the stack?

attestation/src/extensions/pkix/authkeyid.rs
sameo: No, the commit message uses decimal values. 32 is not enough for 20 bytes long ID that translate to 40 bytes when encoding them to a readable string.

attestation/src/measurement.rs
```diff
@@ -149,9 +172,33 @@ impl<'a, D: Digest> AttestationManager<D> {
...
+            .map_err(Error::InvalidCdiExpansion)?;
+
+        // Derive a reduced and salted CDI ID from the TVM specific CDI
+        let id_salt: [u8; 64] = [
+            // From the OpenDice implementation.
+            0xDB, 0xDB, 0xAE, 0xBC, 0x80, 0x20, 0xDA, 0x9F, 0xF0, 0xDD, 0x5A, 0x24, 0xC8, 0x3A,
+            0xA5, 0xA5, 0x42, 0x86, 0xDF, 0xC2, 0x63, 0x03, 0x1E, 0x32, 0x9B, 0x4D, 0xA1, 0x48,
+            0x43, 0x06, 0x59, 0xFE, 0x62, 0xCD, 0xB5, 0xB7, 0xE1, 0xE0, 0x0F, 0xC6, 0x80, 0x30,
+            0x67, 0x11, 0xEB, 0x44, 0x4A, 0xF7, 0x72, 0x09, 0x35, 0x94, 0x96, 0xFC, 0xFF, 0x1D,
+            0xB9, 0x52, 0x0B, 0xA5, 0x1C, 0x7B, 0x29, 0xEA,
```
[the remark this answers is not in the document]
sameo: They are immutable and constant, but I'm expecting the function they live in to be rarely called (The attestation manager constructor, called once per VM lifetime. And the key_pair generation, called with each `GetEvidence` SBI request), so I figure it would make sense to have them on those rare stack frames.

src/vm.rs
sameo: Yes, it can fail is the provided CDI is bogus or not formatted properly. I'll have `Vm::new()` return a `Result`.

sameo: Fixed.

dgreid (approved):

sameo:
@atulkharerivos Are you fine with that PR now?

attestation/src/measurement.rs
```diff
@@ -127,14 +136,28 @@ impl<D: Digest> MeasurementRegister<D> {
 
 /// The attestation manager.
 #[derive(Debug)]
-pub struct AttestationManager<D: Digest> {
+pub struct AttestationManager<D: Digest, H: HmacImpl<D> = hmac::Hmac<D>> {
     // Measurement registers
     measurements: RwLock<ArrayVec<MeasurementRegister<D>, MSMT_REGISTERS>>,
+
+    // Compound Device Identifier (CDI) for the managed TVM
+    cdi_attest: [u8; CDI_LEN],
```
rsahita, line 144: we should put a sealing cdi also in place

atulkharerivos (approved):

attestation/src/measurement.rs
```diff
@@ -149,9 +172,33 @@ impl<'a, D: Digest> AttestationManager<D> {
             );
         }
 
-        AttestationManager {
+        // Derive a TVM specific CDI from the TSM CDI.
+        let kdf = Hkdf::<D, H>::new(Some(&vm_id.to_le_bytes()), cdi);
```
rsahita, line 176: so the recommendation for HKDF salt is to be a fixed random value or NULL - the vm_id could be used as part of the info

sameo: Fixed as well.

rsahita:
lgtm

rsahita (approved):

PR #30, opened 2022-07-20
SBI cleanup and externalization prep

dgreid:
This series cleans up the SBI crate and prepares to move it to an external crate that other projects can consume.

- Patch 1 removes a dependency on riscv-regs, it wasn't providing much value and less path dependencies will make sbi easier to break out.
- Patch 2 adds an API module to better isolate unsafe code and provide users with an easier method of using TSM functions. If this looks good, I'll add wrappers for the other extensions as well.
- Patches 3-8 break individual extensions code out of the main sbi.rs file which had become rather cluttered.


sbi/src/api/tsm.rs
```diff
@@ -0,0 +1,280 @@
+// Copyright (c) 2022 by Rivos Inc.
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
+
+use crate::TeeFunction::*; // TODO
```
abrestic-rivos, line 5: What're the TODO's for?

sbi/src/api/tsm.rs
```diff
@@ -0,0 +1,280 @@
...
+/// Returns information about the TSM that booted this host context.
+pub fn get_info() -> Result<TsmInfo> {
+    let mut tsm_info = TsmInfo::default();
+    let tsm_info_size = core::mem::size_of::<TsmInfo>() as u64;
+    let msg = SbiMessage::Tee(TsmGetInfo {
+        dest_addr: &mut tsm_info as *mut _ as u64,
+        len: tsm_info_size,
+    });
+    // Safety: The passed info pointer is uniquely owned so it's safe to modify in SBI.
+    let tsm_info_len = unsafe { ecall_send(&msg).expect("TsmGetInfo failed") };
```
abrestic-rivos, line 25: Propagate the error instead of `expect()`ing?

sbi/src/api/tsm.rs
```diff
@@ -0,0 +1,280 @@
...
+
+use crate::TeeFunction::*; // TODO
+use crate::{ecall_send, Error, Result, SbiMessage};
+use crate::{TsmInfo, TsmPageType, TvmCpuRegister, TvmCreateParams}; //TODO
+
+/// Initiates a TSM fence on this CPU.
+pub fn initiate_fence() -> Result<()> {
+    let msg = SbiMessage::Tee(TsmInitiateFence);
+    // Safety: TsmInitiateFence doesn't read or write any memory we have access to.
+    unsafe { ecall_send(&msg).map(|_| ()) }
```
abrestic-rivos, line 13: The `ecall_send(&msg).map(..)` pattern gets repeated a lot -- maybe something like this is cleaner?

```
unsafe { ecall_send(&msg) }?;
Ok(())
```

sbi/src/api/tsm.rs
```diff
@@ -0,0 +1,280 @@
...
+        page_addr: addr,
+        page_type: TsmPageType::Page4k,
+        num_pages,
+    });
+    // Safety: The passed-in pages are unmapped and we do not access them again until they're
+    // reclaimed.
+    ecall_send(&msg)?;
+
+    // Fence the pages we just converted.
+    initiate_fence()
```
abrestic-rivos, line 52: Not sure we want to imply `initate_fence()` in convert pages, e.g. if we want to batch convert a bunch of pages first before fencing. Also we probably should have a wrapper for `local_fence()`.

sbi/src/api/tsm.rs
```diff
@@ -0,0 +1,280 @@
...
+
+/// Destroys a TVM created with `tvm_create`.
+pub fn tvm_destroy(vmid: u64) -> Result<()> {
+    let msg = SbiMessage::Tee(TvmDestroy { guest_id: vmid });
+    // Safety: destroying a VM doesn't write to memory that's accessible from the host.
+    unsafe { ecall_send(&msg).map(|_| ()) }
+}
+
+/// Runs the given vcpu of the specified TVM.
+pub fn tvm_run(vmid: u64, vcpu_id: u64) -> Result<u64> {
```
abrestic-rivos, line 120: Maybe convert the u64 to a `TvmCpuExitCode`?

sbi/src/api/tsm.rs
```diff
@@ -0,0 +1,280 @@
...
+/// Copies the data from the pages backing `src_data` to the guest and records their measurement for
+/// attestation.  src_data must be aligned to the given page size.
+pub fn add_measured_pages(
+    vmid: u64,
+    src_data: &[u8],
+    dest_addr: u64,
+    page_type: TsmPageType,
+    guest_addr: u64,
+) -> Result<()> {
+    if src_data
```
abrestic-rivos, line 217: I'm not sure `src_data` actually needs to be aligned here since it isn't the source or destination address of a mapping -- we're just copying from it. (I notice the implementation currently requires it to be page-aligned, but it seems like that could be relaxed.)

abrestic-rivos (commented):
LGTM, nice cleanup!

sbi/src/api/tsm.rs
dgreid: done

sbi/src/api/tsm.rs
dgreid: good idea, done.

sbi/src/api/tsm.rs
dgreid: for sure, I meant to remove that, done.

sbi/src/api/tsm.rs
dgreid: good idea, done

sbi/src/api/tsm.rs
dgreid: Relaxed the constraint here. The implementation can return an error if it wants.

abrestic-rivos (approved):

PR #31, opened 2022-07-21
More sbi functions

dgreid:
Move hart state and reset to have api functions like TSM. I have left attestation in case @sameo has plans to change it. If not I'll do that one too.

### DIFF
--- a/sbi/src/api/mod.rs
+++ b/sbi/src/api/mod.rs
@@ -2,6 +2,10 @@
 // Licensed under the Apache License, Version 2.0, see LICENSE for details.
 // SPDX-License-Identifier: Apache-2.0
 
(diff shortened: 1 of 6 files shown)


sameo:
> Move hart state and reset to have api functions like TSM. I have left attestation in case @sameo has plans to change it. If not I'll do that one too.

The attestation [SBI proposal](https://github.com/sameo/riscv-sbi-doc/pull/1) is still being discussed and reviewed.
If that's ok with you, I'll take care of having the attestation code implemented as API functions once we feel the proposal is getting closer to an agreement.

sameo (approved):

dgreid:
> 

Thanks Sam, attestation sounds good to me. no rush on the api.

PR #36, opened 2022-07-28
Basic PCI config space emulation

abrestic-rivos:
This series enables basic PCI config space emulation, enough for a Linux host VM to enumerate a PCI hierarchy and assign resources. Notably missing are any form interrupt support (needs virtualized capabilites + IOMMU) and the ability for devices to do DMA (needs IOMMU). In general we pass through as much control as possible, with the main exception being bus number assignments since we need to keep requester IDs stable for the IOMMU. Instead, we virtualize the bus number assignments to the host VM and then use these assignments to do the reverse mapping from MMIO trap offset to device.

- Patch 1 is a minor change to avoid dynamic dispatch for `PciDevice`.
- Patch 2 adds helpers for emulating config space reads/writes.
- Patch 3 implements the config space emulation.

### DIFF
--- a/drivers/src/pci/device.rs
+++ b/drivers/src/pci/device.rs
@@ -146,17 +146,6 @@ impl core::fmt::Display for PciDeviceInfo {
     }
 }
 
-/// Common functionality implemented by PCI devices, regardless of type.
-pub trait PciDevice {
-    /// Returns the `PciDeviceInfo` structure read from the config space of this device.
-    fn info(&self) -> &PciDeviceInfo;
-
-    /// Returns the PCI bus address of this device.
-    fn address(&self) -> Address {
-        self.info().address()
-    }
-}
-
 /// Represents a PCI endpoint.
 pub struct PciEndpoint {
     _registers: &'static mut EndpointRegisters,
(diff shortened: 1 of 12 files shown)


drivers/src/pci/device.rs
```diff
@@ -146,19 +149,295 @@ impl core::fmt::Display for PciDeviceInfo {
...
+    dest & !mask | ((src << (offset * 8)) & mask)
+}
+
+// A builder for emulated MMIO config space read operations.
+//
+// In order to support partial accesses to word or dword registers, pushing a word or dword when
+// we're unaligned results in pushing only the bytes between the offset within the word/dword and
+// the end of the word/dword. For example, pushing a word register when `self.offset` is 1 results
+// in only the upper byte of the word being pushed into the result.
+struct MmioReadBuilder {
```
dgreid, line 174: nit: the MmioRead/WriteBuilder seems big enough to have its own file and maybe own commit in this series.

dgreid (approved):
Nice, that macros took a bit to parse, but pays off when declaring all those spans. Just one nit of layout, LGTM.

drivers/src/pci/device.rs
abrestic-rivos: Done.

PR #37, opened 2022-07-28
Add hpmcounter CSR accessors

abrestic-rivos:
We'd like to access these via an index at runtime since it's a bank of 32 identically-formatted registers, however this is kind of a pain due to the CSR number being encoded as part of the instruction (i.e. there's no indirection via a register). 

This series attempts to make `hpmcounter` CSR accesses a little more ergonomic by introducing a trait representing a CSR (patch 1) and then using that trait object to create an array of `hpmcounter` CSRs that we can index into at runtime (patch 2).

There's still some code duplication in the CSR instantiation, but that appears unavoidable without the use of procedural macros, so I'm leaving that as a future exercise.


atulkharerivos (approved):
LGTM.

PR #40, opened 2022-07-29
Stub PCI capability list virtualization

abrestic-rivos:
This series implements pass-through of the PCI capability linked-list structure itself to the host VM. The layout we pass through is equivalent to what's in hardware, but with capabilities we don't care about omitted. For now, all structures in the list appear as null capabilities; we'll implement the various types in a follow on PR.

Patch 1 adds capability list parsing, while patch 2 passes through the stub list.

### DIFF
--- a/drivers/src/pci/capabilities.rs
+++ b/drivers/src/pci/capabilities.rs
@@ -5,6 +5,8 @@
 use arrayvec::ArrayVec;
 use tock_registers::interfaces::Readable;
 
+use super::error::*;
+use super::mmio_builder::{MmioReadBuilder, MmioWriteBuilder};
 use super::registers::*;
 
 // Standard PCI capability IDs.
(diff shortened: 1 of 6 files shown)


drivers/src/pci/capabilities.rs
```diff
@@ -0,0 +1,146 @@
...
+    caps: ArrayVec<PciCapability, MAX_PCI_CAPS>,
+}
+
+impl PciCapabilities {
+    /// Creates a new `PciCapabilities` by parsing the PCI capability linked-list starting at
+    /// `start_offset` within the standard PCI configuration space pointed to be `config_ptr`.
+    ///
+    /// # Safety
+    ///
+    /// `config_ptr` must refer to a valid and uniquely-owned PCI configuration space.
```
dgreid, line 147: what does the caller check to make sure that CommonRegisters are valid? Could that be encapsulated in the type (meaning only valid CommonRegisters can exist)? If so, can this function take a reference to a CommonRegisters and be safe knowing that ref is valid?

drivers/src/pci/capabilities.rs
```diff
@@ -36,11 +38,94 @@ impl CapabilityId {
...
+
+impl PciCapability {
+    // Creates a new capability of type `id` at `ptr`, which itself is at `offset` within the
+    // configuration space.
+    //
+    // # Safety
+    //
+    // `ptr` must refer to a uniquely owned PCI capability structure of the specified type.
+    unsafe fn new(
+        _ptr: NonNull<CapabilityHeader>,
```
dgreid, line 64: Similar question about passing a reference safely instead of a pointer.

drivers/src/pci/device.rs
```diff
@@ -623,15 +588,18 @@ impl PciDevice {
...
+                    self.capabilities().emulate_read(&mut op);
+                }
                 offset => {
                     if offset <= PCI_COMMON_HEADER_END {
                         // Everything else in the common part of the header is unimplemented and we can
                         // safely return 0.
                         op.push_byte(0);
                     } else {
-                        // Make everything beyond the standard header appear unimplemented.
+                        // Make everything beyond the standard PCI configuration space.
```
dgreid, line 600: make everything what?

dgreid (commented):
just a couple safety questions. looks good overall. Thanks!

drivers/src/pci/capabilities.rs
abrestic-rivos: Done.

drivers/src/pci/device.rs
abrestic-rivos: Whoops, done.

dgreid (approved):

PR #41, opened 2022-08-01
More complete PCI capability emulation

abrestic-rivos:
This series adds support for emulating the small set of capabilities we expect to find with QEMU-emulated devices (including `virtio`): MSI(-X), PCIe, PMC, vendor-specific, and bridge SSVID. We use the `enum_dispatch` crate to dispatch calls to the type-specific capability implementations in order to avoid dynamic memory allocation (`Box<dyn Trait>`) and too much code repetition.

With these changes, a Linux host VM is able to fully probe a basic PCI hierarchy.

### DIFF
--- a/drivers/src/pci/capabilities.rs
+++ b/drivers/src/pci/capabilities.rs
@@ -94,6 +94,17 @@ mod bridge_subsys_offsets {
     define_field_span!(BridgeSubsystemRegisters, ssid, u16);
 }
 
+mod express_offsets {
+    use super::ExpressRegisters;
+    use crate::define_field_span;
+
+    define_field_span!(ExpressRegisters, exp_caps, u16);
+    define_field_span!(ExpressRegisters, dev_caps, u32);
+    define_field_span!(ExpressRegisters, dev_control, u16);
+    define_field_span!(ExpressRegisters, link_caps, u32);
+    define_field_span!(ExpressRegisters, link_status, u16);
+}
+
 // Type-specific capability structures.
 #[enum_dispatch]
 enum CapabilityType {
@@ -102,6 +113,7 @@ enum CapabilityType {
     MsiX,
     Vendor,
     BridgeSubsystem,
+    PciExpress,
 }
 
 // Common functionality required by all capabilities.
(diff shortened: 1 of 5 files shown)


drivers/src/pci/capabilities.rs
```diff
@@ -45,21 +47,98 @@ mod header_offsets {
...
+        op.pop_byte();
+    }
+}
+
 // Represents a single PCI capability.
 struct PciCapability {
     id: CapabilityId,
     offset: usize,
     next: usize,
+    cap_type: Option<CapabilityType>,
```
dgreid: Can you add a comment that `None` is used as a type for unimplemented capabilities?

drivers/src/pci/registers.rs
```diff
@@ -298,6 +328,47 @@ impl RegisterMasks for BridgeControl::Register {
     }
 }
 
+// Hide D1/D2 and PME support for now.
```
dgreid, line 462: Should we file a bug to allow power state transitions? For the proof of concept I don't think we need to handle it.

atulkharerivos (approved):
LGTM. I had read about the enum_dispatch crate before, but hadn't seen it in action.

dgreid (approved):

dgreid:
what qemu devices are you adding to test this?

drivers/src/pci/capabilities.rs
abrestic-rivos: Done.

drivers/src/pci/registers.rs
abrestic-rivos: Done.

abrestic-rivos:
> what qemu devices are you adding to test this?

My config looks like this:

```
   -fsdev local,id=root,path=/rivos/sysroot/riscv/,security_model=none \
   -device virtio-9p-pci,fsdev=root,mount_tag=lroot \
   -fsdev local,id=workspace,path=/workspace,security_model=none \
   -device virtio-9p-pci,fsdev=workspace,mount_tag=lworkspace \
   -device nvme,serial=deadbeef,bus=pcie.0 \
   -device pcie-root-port,id=rp1,bus=pcie.0,chassis=1,addr=4.0,multifunction=on \
   -netdev user,id=usernet,hostfwd=tcp:127.0.0.1:7722-0.0.0.0:22 \
   -device virtio-net-pci,netdev=usernet,bus=rp1 \
   -device pcie-root-port,id=rp2,bus=pcie.0,chassis=2,addr=4.1 \
   -device e1000e,bus=rp2 \
   -device pcie-root-port,id=rp3,bus=pcie.0,chassis=3,addr=8.0 \
   -device virtio-rng-pci,rng=rng0 \
   -object rng-random,filename=/dev/urandom,id=rng0 \
```

Still pretty useless without DMA though, but at least the kernel is able to probe things, find MSI vectors, etc. Note that `virtio-rng-pci` fails to probe because it's INTx only.

PR #44, opened 2022-08-02
Vm: Separate the vCPU context switch from the virtualization mode switch

abrestic-rivos:
This patch separates the state that needs to be saved/restored on virtualization entry and exit vs the state that needs to be saved/restored when switching between vCPUs which should make VM entry/exit lighter-weight, especially after we add PMU support.

@atulkharerivos PTAL. I _think_ this makes the PMU context-switching cleaner: snapshot counter values when swapping out a vCPU, and apply the saved counter configuration & values when swapping a vCPU back in. Then, when a vCPU makes a PMU SBI call, forward along the SBI call to firmware (with flag changes, of course) and update our internal copy of the counter state so we know what we need to save/restore.

### DIFF
--- a/src/vm.rs
+++ b/src/vm.rs
@@ -20,7 +20,7 @@ use sbi::*;
 use crate::guest_tracking::{GuestState, Guests};
 use crate::print_util::*;
 use crate::smp;
-use crate::vm_cpu::{VirtualRegister, VmCpuExit, VmCpuStatus, VmCpus, VM_CPU_BYTES};
+use crate::vm_cpu::{ActiveVmCpu, VirtualRegister, VmCpuExit, VmCpuStatus, VmCpus, VM_CPU_BYTES};
 use crate::vm_pages::Error as VmPagesError;
 use crate::vm_pages::{
     ActiveVmPages, InstructionFetchError, PageFaultType, VmPages, VmRegionList,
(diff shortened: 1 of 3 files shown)


atulkharerivos (approved):
LGTM -- I will have to remember the type erasure trick to get rid of lifetimes.

src/vm.rs
```diff
@@ -405,26 +405,28 @@ impl<T: GuestStagePageTable> Vm<T, VmStateFinalized> {
...
         // Take the vCPU out of self.vcpus, giving us exclusive ownership.
         let mut vcpu = self
             .vcpus
             .take_vcpu(vcpu_id)
             .map_err(|_| EcallError::Sbi(SbiError::InvalidParam))?;
         let exit_code = {
-            let mut vcpu = vcpu.lock();
+            let mut locked_vcpu = vcpu.lock();
+            // Activate this vCPU, saving the state of our parent vCPU.
+            let mut active_vcpu = locked_vcpu.activate(&self.vm_pages, parent_vcpu).unwrap();
```
dgreid, line 421: Is the unwrap Ok here or can this fail?

dgreid (commented):
nice, thanks, just one unwrap question.

src/vm.rs
abrestic-rivos: `activate()` will only fail if there's a mismatch between the page owner ID of `self.vm_pages` and the vCPU, so it's guaranteed to succeed here. But I went ahead and changed it to check for the error since it's easy enough to bail out here and it's conceivable that we'd add something down the road that could make `activate()` fail for other reasons.

Issue #43, opened 2022-08-02
IOMMU support


abrestic-rivos:
In order to enable DMA (including MSI-X) with the passed-through PCI devices to the host VM, we need the IOMMU to perform the 2nd-stage translation for VMs and to isolate Salus / other VMs from the DMAs.

PR #50, opened 2022-08-08
Probe for an IOMMU device

abrestic-rivos:
This series adds support to the PCI driver for claiming a PCI device for internal hypervisor usage and then uses this support to probe for and claim a RISC-V IOMMU device on the root bus.

Patches 1-7 are cleanups and prep-work. Notably they add support for per-device ownership, which we'll use to hide hypervisor-claimed devices from VMs.

Patch 8 adds support for claiming a device for hypervisor-internal usage. This includes the very basic ability to assign BAR resources to a PCI device, which for now is limited to just devices directly behind the host bridge (which is fine for the IOMMU since it's always a RCiEP). Resources assigned to hypervisor devices are taken from the end of the host bridge windows and are hidden from the host VM.

Patch 9 adds a stub IOMMU driver which finds the IOMMU on the PCI bus and verifies it supports the necessary features.

Tested with the RFC RISC-V IOMMU support patches. A QEMU branch with these patches (+ Sstc support) can be found here: https://github.com/abrestic-rivos/qemu/tree/testing/sstc-iommu.

With these changes, Salus is able to discover the IOMMU and hide it from a Linux host VM.

Related to #43

### DIFF
--- a/drivers/src/pci/bus.rs
+++ b/drivers/src/pci/bus.rs
@@ -9,14 +9,14 @@ use super::address::*;
 use super::config_space::PciConfigSpace;
 use super::device::PciDevice;
 use super::error::*;
-use super::root::{PciDeviceArena, PciDeviceId};
+use super::root::{PciArenaId, PciDeviceArena};
 
 /// A entry for a single device on a `PciBus`.
 pub struct BusDevice {
     /// The PCI bus address of the device.
     pub address: Address,
     /// The ID of the device in the `PciDeviceArena`.
-    pub id: PciDeviceId,
+    pub id: PciArenaId,
 }
 
 /// Represents a PCI bus.
--- a/drivers/src/pci/bus.rs
+++ b/drivers/src/pci/bus.rs
@@ -3,6 +3,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 use alloc::vec::Vec;
+use spin::Mutex;
 
 use super::address::*;
 use super::config_space::PciConfigSpace;
@@ -48,7 +49,7 @@ impl PciBus {
                 // by creating at most one device per PCI address.
                 let pci_dev = unsafe { PciDevice::new(registers_ptr, info.clone()) }?;
                 let id = device_arena
-                    .try_insert(pci_dev)
+                    .try_insert(Mutex::new(pci_dev))
                     .map_err(|_| Error::AllocError)?;
                 let entry = BusDevice {
                     address: info.address(),
@@ -64,8 +65,9 @@ impl PciBus {
         for bd in devices.iter() {
             let bridge_id = bd.id;
             let sec_bus = cur_bus.next().ok_or(Error::OutOfBuses)?;
-            match device_arena.get_mut(bridge_id) {
-                Some(PciDevice::Bridge(bridge)) => {
+            // ID must be valid, we just added it above.
+            match *device_arena.get(bridge_id).unwrap().lock() {
+                PciDevice::Bridge(ref mut bridge) => {
                     // Set the bridge to cover everything beyond sec_bus until we've enumerated
                     // the buses behind the bridge.
                     bridge.assign_bus_range(BusRange {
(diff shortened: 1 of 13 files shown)


drivers/src/pci/bus.rs
```diff
@@ -64,8 +65,8 @@ impl PciBus {
         for bd in devices.iter() {
             let bridge_id = bd.id;
             let sec_bus = cur_bus.next().ok_or(Error::OutOfBuses)?;
-            match device_arena.get_mut(bridge_id) {
-                Some(PciDevice::Bridge(bridge)) => {
+            match *device_arena.get(bridge_id).unwrap().lock() {
```
atulkharerivos, line 68: Unwrap might need comment unless obvious from context.

drivers/src/pci/bus.rs
```diff
@@ -81,8 +82,8 @@ impl PciBus {
 
             // Avoid double mutable borrow of device_arena by re-acquiring the reference to the bridge
             // device here. PciBus::enumerate() may have added devices and re-allocated the arena.
-            match device_arena.get_mut(bridge_id) {
-                Some(PciDevice::Bridge(bridge)) => {
+            match *device_arena.get(bridge_id).unwrap().lock() {
```
atulkharerivos, line 85: Unwrap might need comment unless obvious from context.

drivers/src/pci/root.rs
```diff
@@ -424,19 +359,77 @@ impl PcieRoot {
         if !valid_config_mmio_access(offset, len) {
             return Err(Error::UnsupportedConfigAccess);
         }
-        let mut devices = self.devices.lock();
-        let (dev_id, dev_offset) = devices.virtual_config_offset_to_device(offset as usize)?;
+        let (dev_id, dev_offset) = self.virtual_config_offset_to_device(offset as usize)?;
         // If the device ID is present in the hierarchy, then it must be in the arena.
-        let dev = devices.device_arena.get_mut(dev_id).unwrap();
+        let dev = self.device_arena.get(dev_id).unwrap();
```
atulkharerivos: Unwrap might need comment unless obvious from context.

drivers/src/pci/root.rs
```diff
@@ -424,19 +359,77 @@ impl PcieRoot {
...
     }
+
+    // Calls `f` for each device on `bus`.
+    fn for_each_device_on<F: FnMut(PciDeviceId)>(&self, bus: &PciBus, f: &mut F) {
+        for bd in bus.devices() {
+            f(bd.id);
+            // If the ID appears on a bus, it must be in the arena.
+            let dev = self.device_arena.get(bd.id).unwrap();
+            if let PciDevice::Bridge(ref bridge) = *dev.lock() {
+                // Recrusively walk the buses behind the bridge.
```
atulkharerivos, line 481: Spelling.

dgreid (approved):

drivers/src/pci/bus.rs
abrestic-rivos: Added a comment.

drivers/src/pci/bus.rs
abrestic-rivos: This one should be clear from the comment immediately above it (+ the context of having just looked up the device a few lines up).

drivers/src/pci/root.rs
abrestic-rivos: There's a comment already on the line right above this.

drivers/src/pci/root.rs
abrestic-rivos: Done.

drivers/src/pci/root.rs
```diff
@@ -248,6 +284,56 @@ impl PcieRoot {
...
+    /// it for use within the hypervisor by assigning it resources. Returns a `PciDeviceId` which
+    /// can be used to retrieve a reference to the device on success.
+    pub fn take_and_enable_hypervisor_device(
+        &self,
+        vendor_id: VendorId,
+        device_id: DeviceId,
+    ) -> Result<PciArenaId> {
+        let mut dev_id: Option<PciArenaId> = None;
+        self.for_each_device_on(&self.root_bus, &mut |id| {
+            let dev = self.device_arena.get(id).unwrap().lock();
```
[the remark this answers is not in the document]
abrestic-rivos: This is actually the instance where I want the ID :). I'm saving it into `dev_id` so I can lookup the device and do all the resource assignment / enablement stuff outside of a closure.

(I'd also considered keeping the ID inside `PciDevice` for this purpose, but that didn't end up looking any cleaner..)

atulkharerivos:
LGTM..regarding the unwrap() comment, github collapses the previous lines, so they aren't visible at times. Note to myself to read the patch in the full editor instead of just the diffs.

atulkharerivos (approved):

PR #51, opened 2022-08-10
IMSIC cleanups in preparation for MSI page tables

abrestic-rivos:
Prepare for MSI page table support by introducing `ImsicGeometry` and `ImsicLocation` which support translation between an IMSIC location specifier (effectively a group + hart + guest index tuple) and the address of the interrupt file.

- Patch 1 fixes a compile error recently introduced by a nightly update.
- Patch 2 separates `imsic.rs` into multiple modules to avoid making the main IMSIC module too crowded.
- Patch 3 makes `Imsic::probe_from()` propagate errors rather than asserting/panicing. Not strictly related to the rest of the series, but is good practice nonetheless.
- Patch 4 adds the new `ImsicGeometry` and related types.
- Patch 5 is a minor cleanup to remove the unnecessary `ImsicMmioRegion` type.
- Patch 6 simplifies the IMSIC per-CPU state management and initial mapping of the IMSIC into the host VM.

Tested by booting SMP Linux; no regressions.

### DIFF
--- a/drivers/src/imsic/error.rs
+++ b/drivers/src/imsic/error.rs
@@ -27,6 +27,10 @@ pub enum Error {
     InvalidMmioRegion(u64),
     /// Invalid parent interrupt specification in the device tree.
     InvalidParentInterrupt(u32, u32),
+    /// The given CPU does not have a valid IMSIC location specified in the device tree.
+    InvalidCpuLocation(CpuId),
+    /// The given CPU was referenced multiple times in the device tree.
+    DuplicateCpuEntries(CpuId),
     /// There were more interrupt files than CPUs found in the device tree.
     TooManyInterruptFiles,
     /// There were fewer interrupt files than CPUs found in the device tree.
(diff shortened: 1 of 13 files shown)


drivers/src/imsic/error.rs
```diff
@@ -4,9 +4,33 @@
 
 use crate::CpuId;
 
-/// Errors that can be returned when claiming or releasing guest interrupt files.
+/// Errors that can be returned by the IMSIC driver.
 #[derive(Clone, Copy, Debug, PartialEq, Eq)]
 pub enum Error {
+    /// The IMSIC node was not present in the deivce tree.
```
atulkharerivos, line 10: Spelling.

dgreid (approved):
Looks like it's going in a great direction!

Only nit is that the name 'main.rs' is confusing with a binary rust program.

drivers/src/imsic/error.rs
abrestic-rivos: Done.

abrestic-rivos:
> Looks like it's going in a great direction!
> 
> Only nit is that the name 'main.rs' is confusing with a binary rust program.

Good point -- renamed `main.rs` to `core.rs`. (Would've preferred `imsic/imsic.rs` but clippy didn't like that.)

PR #52, opened 2022-08-12
IOMMU: Add support for building the device directory and MSI page tables

abrestic-rivos:
Add support for building the two translation structures used by the IOMMU that we'll be using Salus: the device directory (used to map devices to translation contexts) and the MSI page table (used for translating MSI writes). For now we need only deal with 2nd-stage translation (equivalent to the VMs CPU page tables) and MSI translation for each device.

- Patches 1 & 2 are minor cleanups.
- Patch 3 adds wrappers for `fence` instructions that we'll need for ordering accesses to DMA-able memory and MMIO.
- Patch 4 adds support for building MSI page tables.
- Patch 5 adds support for building the device directory.


dgreid (approved):
couple nits and questions. LGTM. Thanks for adding the tests that helped follow how it'll be used.

PR #63, opened 2022-08-18
Don't assume FDT is contiguous with the hypervisor image

abrestic-rivos:
Fixes an assert that trips when booting with 2GB of memory and QEMU's bundled OpenSBI binary.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -120,26 +120,29 @@ fn build_memory_map<T: GuestStagePagingMode>(fdt: &Fdt) -> MemMapResult<HwMemMap
     // Safe because we trust the linker placed these symbols correctly.
     let start = unsafe { core::ptr::addr_of!(_start) as u64 };
     let stack_end = unsafe { core::ptr::addr_of!(_stack_end) as u64 };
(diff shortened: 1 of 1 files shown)


dgreid (approved):
This makes a lot of sense. Thanks!

atishp04:
While the change looks good, the reasoning in the commit text is partially incorrect. 
Which firmware type you were using ? I guess you are using fw_jump or fw_payload. Both of them has a fixed offset of DTB loading address. That's why OpenSBI relocated the DTB to FW_JUMP_FDT_ADDR which is defined in object.mk

FW_JUMP_FDT_ADDR=$(shell printf "0x%X" $$(($(FW_TEXT_START) + 0x2200000))).

Coincidentally, the address turned out to be right after salus image.

However, the default OpenSBI image in Qemu is fw_dynamic which doesn't relocate DTB at all. Fw_dynamic allows the loader to place the blobs at the correct place.

If you provide fw_dynamic at -bios the DTB will be loaded to 0xbfe00000 as well because that's where qemu loader loads the DTB.
@abrestic-rivos ^^

PR #66, opened 2022-08-22
Support setting AIA virtualization params for TVMs

abrestic-rivos:
This PR introduces two new TEECALLs (in a new extension) for configuring AIA virtualization in a TVM: setting the TVM's IMSIC geometry (`TvmInitAia`) and setting a vCPU's IMSIC location (`TvmCpuSetImsicAddr`). We'll build on this in following PRs to enable assignment of guest interrupt files from the host VM to its child TVMs.

I've initially chosen to put this in a separate extension, though I'm definitely open to other options here. Rationale for making it an optional extension:
- We may want to support nested TVMs, and nested AIA virtualization gets extremely complicated.
- It's possible to do interrupt virtualization without hardware AIA virtualization using emulated MMIO and (to-be-implemented) support for injecting interrupts.

Patches 1-3 are minor cleanups, patch 4 introduces the SBI definitions, patches 5&6 wire up the new TEECALLs, and patch 7 tests them out in `tellus`.

### DIFF
--- a/src/vm.rs
+++ b/src/vm.rs
@@ -663,6 +663,7 @@ impl<T: GuestStagePagingMode> Vm<T, VmStateFinalized> {
             SbiMessage::Attestation(attestation_func) => {
                 self.handle_attestation_msg(attestation_func, active_vcpu.active_pages())
             }
+            SbiMessage::TeeAia(_) => EcallAction::Continue(SbiReturn::from(SbiError::NotSupported)),
             SbiMessage::Pmu(pmu_func) => self.handle_pmu_msg(pmu_func, active_vcpu).into(),
         }
     }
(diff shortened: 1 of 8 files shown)


src/vm.rs
```diff
@@ -352,6 +357,27 @@ impl<T: GuestStagePagingMode> Vm<T, VmStateInitializing> {
 
     /// Completes intialization of the `Vm`, returning it in a finalized state.
     pub fn finalize(self) -> Result<Vm<T, VmStateFinalized>> {
+        // If AIA virtualization is enabled, make sure all vCPUs got assigned an IMSIC address and
+        // that there are no aliases.
+        for i in 0..self.vcpus.num_vcpus() {
```
dgreid, line 362: nit: this might be easier to read and test if contained in a function like `has_imsic_overlap` or similar.

dgreid (approved):

src/vm.rs
abrestic-rivos: Done, moved this to a separate helper method.

PR #68, opened 2022-08-23
Console driver cleanup

abrestic-rivos:
Introduce a `Console` object for implementing `print!` and `println!` that can have the writer set at runtime via a `ConsoleWriter` trait object (patches 1 & 2). This allows us to switch from SBI to UART console, and to use the console outside of the binary crates. Finally, clean up the UART driver by moving it to the `drivers` crate (patch 3) and probe for it from the device tree so that we don't have to use hard-coded addresses (patches 4 & 5).


s-mode-utils/src/print.rs
```diff
@@ -0,0 +1,63 @@
+// Copyright (c) 2021 by Rivos Inc.
```
dgreid, line 1: 2022

s-mode-utils/src/sbi_console.rs
```diff
@@ -0,0 +1,34 @@
+// Copyright (c) 2021 by Rivos Inc.
```
dgreid, line 1: 2022

drivers/src/uart.rs
```diff
@@ -15,17 +36,37 @@ pub struct UartDriver {
...
+        {
+            return Err(Error::InvalidRegisterLocation);
+        }
+        // Safety: We trust that the device tree accurately described the location of the UART.
+        unsafe {
+            mem_map
+                .add_mmio_region(DeviceMemType::Uart, RawAddr::supervisor(base_address), len)
+                .map_err(Error::AddingMmioRegion)
+        }?;
+        // Unwrap ok, we've already rified that base_address is non-NULL.
```
dgreid, line 63: verified?

dgreid (approved):
way better, thanks!

s-mode-utils/src/print.rs
abrestic-rivos: Copy-pasta. Done.

s-mode-utils/src/sbi_console.rs
abrestic-rivos: Done.

drivers/src/uart.rs
abrestic-rivos: Done.

PR #78, opened 2022-08-29
Initial SBI attestation update

### LINKED ISSUE #77: Update the attestation SBI support
The second round of reviews for the proposed attestation SBI extensions (https://github.com/sameo/riscv-sbi-doc/pull/1) received only minor feedback.
We should now start updating the Salus implementation accordingly.

sameo:
This PR updates the `GetEvidence` function handler and adds a high level API for it.
It also adds a new function: `GetCapabilities`, that lets the attestation SBI callers know which attestation capabilities and features are available before calling into the rest of the API.

This partially fixes #77

### DIFF
--- a/sbi/src/api/attestation.rs
+++ b/sbi/src/api/attestation.rs
@@ -1,0 +1,62 @@
+// Copyright (c) 2022 by Rivos Inc.
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
+
+use arrayvec::ArrayVec;
+
+use crate::{
+    ecall_send, AttestationFunction, Error, EvidenceFormat, Result, SbiMessage,
+    EVIDENCE_DATA_BLOB_SIZE,
+};
+
+/// Maximum supported size for the attestation evidence certificate.
+pub const MAX_CERT_SIZE: usize = 4096;
+
+/// Get an attestation evidence.
+/// This function returns a serialized, DER formatted X.509 certificate.
+/// The attestation evidence is included as a certificate extension.
+///
+/// # Arguments
+///
+/// * `cert_request` - The [Certificate Signing Request]((https://datatracker.ietf.org/doc/html/rfc2986) buffer.
+/// * `request_data` - A data blob that will be included in the generated
+///                    certificate, as [UserNotice]((https://datatracker.ietf.org/doc/html/rfc2986)
+///                    X.509 certificate extension. This is typically used to
+///                    pass a cryptographic nonce.
+/// * `evidence_format` - The format of the attestation evidence as defined by [`EvidenceFormat`](crate::api::attestation::EvidenceFormat).
+pub fn get_evidence(
+    cert_request: &[u8],
+    request_data: &[u8],
+    evidence_format: EvidenceFormat,
+) -> Result<ArrayVec<u8, MAX_CERT_SIZE>> {
+    if request_data.len() != EVIDENCE_DATA_BLOB_SIZE {
+        return Err(Error::InvalidParam);
+    }
+
+    let mut cert_bytes = ArrayVec::<u8, MAX_CERT_SIZE>::new();
+    let msg = SbiMessage::Attestation(AttestationFunction::GetEvidence {
+        cert_request_addr: cert_request.as_ptr() as u64,
+        cert_request_size: cert_request.len() as u64,
+        request_data_addr: request_data.as_ptr() as u64,
+        evidence_format: evidence_format as u64,
+        cert_addr_out: (cert_bytes.as_ptr()) as u64,
+        cert_size: MAX_CERT_SIZE as u64,
+    });
+
+    // Safety: GetEvidence only reads the pages pointed to by `cert_request` and
+    // `request_data`. This is safe because they're owned by the borrowed slices
+    // passed as arguments.
+    // GetEvidence writes to a single reference to `cert_bytes``, which is
+    // defined in this scope.
+    let len = unsafe { ecall_send(&msg) }?;
+
+    // Safety: cert_bytes is backed by a MAX_CERT_SIZE array.
+    unsafe {
+        if len as usize > MAX_CERT_SIZE {
+            return Err(Error::Failed);
+        }
+        cert_bytes.set_len(len as usize);
+    };
+
+    Ok(cert_bytes)
+}
--- a/sbi/src/api/attestation.rs
+++ b/sbi/src/api/attestation.rs
@@ -5,12 +5,26 @@
 use arrayvec::ArrayVec;
 
 use crate::{
-    ecall_send, AttestationFunction, Error, EvidenceFormat, Result, SbiMessage,
-    EVIDENCE_DATA_BLOB_SIZE,
+    ecall_send, AttestationCapabilities, AttestationFunction, Error, EvidenceFormat, Result,
+    SbiMessage, EVIDENCE_DATA_BLOB_SIZE,
 };
 
 /// Maximum supported size for the attestation evidence certificate.
 pub const MAX_CERT_SIZE: usize = 4096;
+
+/// Get the attestation capabilities.
+pub fn get_capabilities() -> Result<AttestationCapabilities> {
+    let caps = AttestationCapabilities::default();
+    let msg = SbiMessage::Attestation(AttestationFunction::GetCapabilities {
+        caps_addr_out: (&caps as *const AttestationCapabilities) as u64,
+        caps_size: core::mem::size_of::<AttestationCapabilities>() as u64,
+    });
+
+    // Safety: &caps is the single reference to a variable defined in this scope.
+    unsafe { ecall_send(&msg) }?;
+
+    Ok(caps)
+}
 
 /// Get an attestation evidence.
 /// This function returns a serialized, DER formatted X.509 certificate.
--- a/sbi/src/attestation.rs
+++ b/sbi/src/attestation.rs
@@ -5,28 +5,60 @@
 use crate::error::*;
 use crate::function::*;
 
+/// The data blob passed to the GetEvidence call must be 64 bytes long.
+pub const EVIDENCE_DATA_BLOB_SIZE: usize = 64;
+
+/// Attestation evidence formats.
+#[derive(Copy, Clone, Debug)]
+#[repr(u64)]
+pub enum EvidenceFormat {
+    /// Single layer DICE TCB
+    /// https://trustedcomputinggroup.org/resource/dice-attestation-architecture/
+    DiceTcbInfo = 0,
+
+    /// X.509 extension for multiple layers DICE TCB.
+    /// https://trustedcomputinggroup.org/resource/dice-attestation-architecture/
+    DiceMultiTcbInfo = 1,
+
+    /// Open DICE profile
+    /// https://pigweed.googlesource.com/open-dice/+/HEAD/docs/specification.md
+    OpenDice = 2,
+}
+
 /// Functions provided by the attestation extension.
 #[derive(Copy, Clone)]
 pub enum AttestationFunction {
-    /// Get an attestion evidence from a CSR (https://datatracker.ietf.org/doc/html/rfc2986).
+    /// Get an attestion evidence from a Certificate Signing Request (CSR)
+    /// (https://datatracker.ietf.org/doc/html/rfc2986).
     /// The caller passes the CSR and its length through the first 2 arguments.
-    /// The third argument is the address where the generated certificate will be placed.
+    /// The third argument is the address where the caller places a data blob
+    /// that will be included in the generated certificate. Typically, this is a
+    /// cryptographic nonce.
+    /// The fourth argument is the evidence format: DiceTcbInfo (0),
+    /// DiceMultiTcbInfo (1) or OpenDice (2).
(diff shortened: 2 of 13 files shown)


sameo:
There will be a follow up PR for the measurement registers read and extension SBI API update.

sbi/src/api/attestation.rs
```diff
@@ -0,0 +1,65 @@
...
+///                    X.509 certificate extension. This is typically used to
+///                    pass a cryptographic nonce.
+/// * `evidence_format` - The format of the attestation evidence as defined by [`EvidenceFormat`](crate::api::attestation::EvidenceFormat).
+/// * `cert_out` - The generated X.509 certificate buffer.
+pub fn get_evidence<'a>(
+    cert_request: &'a [u8],
+    request_data: &'a [u8],
+    evidence_format: EvidenceFormat,
+    cert_out: &'a mut [u8],
+) -> Result<u64> {
```
abrestic-rivos, line 159: What's the return value? If it's always 0 on success, would `Result<()>` be a more appropriate return type?

sameo: I changed the API to return the certificate buffer itself. It makes it for a nicer high level SBI API.

sbi/src/api/attestation.rs
```diff
@@ -0,0 +1,65 @@
...
+///                    certificate, as [UserNotice]((https://datatracker.ietf.org/doc/html/rfc2986)
+///                    X.509 certificate extension. This is typically used to
+///                    pass a cryptographic nonce.
+/// * `evidence_format` - The format of the attestation evidence as defined by [`EvidenceFormat`](crate::api::attestation::EvidenceFormat).
+/// * `cert_out` - The generated X.509 certificate buffer.
+pub fn get_evidence<'a>(
+    cert_request: &'a [u8],
+    request_data: &'a [u8],
+    evidence_format: EvidenceFormat,
+    cert_out: &'a mut [u8],
```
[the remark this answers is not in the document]
sameo: They're not, thanks for catching that.

sbi/src/api/attestation.rs
```diff
@@ -2,29 +2,142 @@
...
+    /// The supported hash algorithm.
+    pub hash_algorithm: HashAlgorithm,
+    /// The supported evidence formats. This is a bitmap.
+    pub evidence_formats: FlagSet<EvidenceFormat>,
+    /// Number of static measurement registers.
+    pub static_measurements: u8,
+    /// Number of runtime measurement registers.
+    pub runtime_measurements: u8,
+    /// Array of all measurement register descriptors.
+    pub measurement_registers: ArrayVec<MeasurementRegisterDescriptor, 32>,
```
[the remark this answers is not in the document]
sameo: Makes sense, fixed now.

sbi/src/api/attestation.rs
```diff
@@ -2,29 +2,142 @@
...
+
 use crate::ecall_send;
 use crate::{AttestationFunction, Error, Result, SbiMessage};
 
-/// The data blob passed to the GetEvidence call must be 64 bytes long.
-pub const REQUEST_DATA_SIZE: usize = 64;
+/// A list of supported hash algorithms.
+#[derive(Copy, Clone, Default, Debug)]
+#[repr(u8)]
+pub enum HashAlgorithm {
```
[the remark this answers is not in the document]
sameo: So I tried to keep all the external SBI attestation API (Including the returned structures) under `sbi/src/api` as I was assuming this is where we'd put our external API definitions.
Should I only keep the functions there?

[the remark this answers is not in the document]
dgreid: The distinction for me was that src/<extension>.rs covers definitions from what will eventually be the SBI spec. src/api/<extension>.rs is rust helpers for hiding unsafe and avoid manual struct building at the call sites.

If that makes sense to others, I can add something to the readme so we have a chance at keeping it that way.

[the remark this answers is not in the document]
atulkharerivos: @dgreid That approach does make sense to me.

sbi/src/api/attestation.rs
```diff
@@ -13,21 +14,134 @@ pub const MAX_CERT_SIZE: usize = 4096;
...
+}
+
+impl MeasurementRegisterDescriptor {
+    /// Create a new measurement register descriptor.
+    pub fn new(
+        tcb_layer_index: u8,
+        fwid_index: u8,
+        tcg_pcr_index: u8,
+        runtime: bool,
+    ) -> MeasurementRegisterDescriptor {
```
atulkharerivos, line 123: Nit: -> Self

sbi/src/api/attestation.rs
```diff
@@ -13,21 +14,134 @@ pub const MAX_CERT_SIZE: usize = 4096;
...
+            fwid_index,
+            tcg_pcr_index,
+            runtime,
+        }
+    }
+}
+
+/// Get the attestation capabilities.
+pub fn get_capabilities() -> Result<AttestationCapabilities> {
+    let caps: AttestationCapabilities = Default::default();
```
atulkharerivos, line 135: Nit: l`et caps = AttestationCapabilities::default();`

test-workloads/src/bin/guestvm.rs
```diff
@@ -176,6 +176,18 @@ pub fn test_vector() {
 const TEST_CSR: &[u8] = include_bytes!("test-ed25519.der");
 
 fn test_attestation() {
+    let caps = attestation::get_capabilities().expect("Failed to get attestation capabilities");
```
atulkharerivos, line 179: Should we execute the SBI probe below first before panicking?

sbi/src/api/attestation.rs
```diff
@@ -13,21 +14,134 @@ pub const MAX_CERT_SIZE: usize = 4096;
...
+
+impl AttestationCapabilities {
+    /// Create new attestation capabilities structure.
+    pub fn new(
+        tcb_svn: u64,
+        hash_algorithm: HashAlgorithm,
+        evidence_formats: impl Into<FlagSet<EvidenceFormat>>,
+        static_measurements: u8,
+        runtime_measurements: u8,
+    ) -> AttestationCapabilities {
```
atulkharerivos, line 80: Nit: -> Self

sbi/src/api/attestation.rs
```diff
@@ -13,21 +14,134 @@ pub const MAX_CERT_SIZE: usize = 4096;
...
+    pub hash_algorithm: HashAlgorithm,
+    /// The supported evidence formats. This is a bitmap.
+    pub evidence_formats: FlagSet<EvidenceFormat>,
+    /// Number of static measurement registers.
+    pub static_measurements: u8,
+    /// Number of runtime measurement registers.
+    pub runtime_measurements: u8,
+    /// Array of all measurement register descriptors.
+    pub measurement_registers: [MeasurementRegisterDescriptor; 32],
+
```
atulkharerivos, line 68: Nit: Delete newline or add a comment? Also, seems off to have an one-off non-public member, but it may be necessary in this case.

sbi/src/api/attestation.rs
```diff
@@ -13,21 +14,134 @@ pub const MAX_CERT_SIZE: usize = 4096;
...
+flags! {
+    /// Attestation evidence formats.
+    #[derive(Default)]
+    #[repr(u8)]
+    pub enum EvidenceFormat: u8 {
+        /// Single layer DICE TCB
+        /// https://trustedcomputinggroup.org/resource/dice-attestation-architecture/
+        #[default]
+        DiceTcbInfo = 1,
+
```
atulkharerivos, line 37: Nit: Redundant newline here and below.

sbi/src/api/attestation.rs
sameo: Fixed, thanks.

test-workloads/src/bin/guestvm.rs
sameo: Yep, good catch.

sbi/src/api/attestation.rs
sameo: Fixed, thanks.

sbi/src/api/attestation.rs
sameo: Fixed, thanks.

sbi/src/api/attestation.rs
sameo: I added a comment. Yes, this is an internal field for the measurement array index tracking. We check for out of bounds and return an error when that happens.

sbi/src/api/attestation.rs
sameo: I personally find it more readable with those newline, but I see the rest of the code does not follow that pattern. Thanks.

sbi/src/api/attestation.rs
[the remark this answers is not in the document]
sameo: @dgreid I'm fine with that approach, I fixed the PR accordingly.

Are we planning to have the SBI crate grow as an external crate to be used for Rust SBI implementations? If so, I suggest to add the README snippet under `sbi/README.md` ?

sbi/src/error.rs
```diff
@@ -24,6 +24,10 @@ pub enum Error {
     AlreadyStarted = -7,
     /// Some of the given counters have already been stopped.
     AlreadyStopped = -8,
+    /// Some resources are busy and could not process the request.
+    Busy = -9,
+    /// The response to the SBI call is too large and overflows the allocated capacities.
+    Overflow = -10,
```
abrestic-rivos, line 30: Have these been added to the spec?

sbi/src/attestation.rs
```diff
@@ -5,29 +5,147 @@
...
+    pub evidence_formats: FlagSet<EvidenceFormat>,
+    /// Number of static measurement registers.
+    pub static_measurements: u8,
+    /// Number of runtime measurement registers.
+    pub runtime_measurements: u8,
+    /// Array of all measurement register descriptors.
+    pub measurement_registers: [MeasurementRegisterDescriptor; MAX_MEASUREMENT_REGISTERS],
+
+    // Track the current measurement_registers array index.
+    measurement_registers_idx: usize,
```
abrestic-rivos, line 67: Is this only used when building the `AttestationCapabilities` struct, or does the receiver of this struct also use it to determine how many `measurement_registers` are populated? If it's the former, then I'm not sure it belongs in the ABI.

sbi/src/error.rs
sameo: We're proposing to add those with the attestation SBI extension:
https://github.com/sameo/riscv-sbi-doc/pull/1/files

sbi/src/attestation.rs
```diff
@@ -5,29 +5,147 @@
...
+
+    /// Add a measurement register to the attestation capabilities.
+    pub fn measurement_register(
+        &mut self,
+        register: MeasurementRegisterDescriptor,
+    ) -> Result<&mut AttestationCapabilities> {
+        let m = self
+            .measurement_registers
+            .get_mut(self.measurement_registers_idx)
+            .ok_or(Error::Overflow)?;
```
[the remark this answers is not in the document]
sameo: Here the error is related to the SBI implementation that would potentially be trying to push a measurement register description out of bound. So it's not related to the validity of any of the passed parameter, which is why we're using this new error code (That we'd also use for callers trying to add measurement out of bounds).

sbi/src/attestation.rs
sameo: I think you're right. I've added an index to the `add_measurement_register` call and let the SBI implementation decide where the descriptor should land in the array.

abrestic-rivos (approved):

sbi/src/error.rs
atulkharerivos: A related question about the semantics (and the associated name). Since the response would overflow the passed in buffer, the presumably the passed-in input buffer was too small, so perhaps it should be InsufficientCapacity / BufferTooSmall, or something like that?

sameo: The passed in buffer could be valid (The attestation caps structure has a known max size), but the SBI implementation could try to write passed the maximum number of allowed  registers. I think we should look a this as a different error than passing too small of a buffer.

sbi/src/attestation.rs
```diff
@@ -5,29 +5,149 @@
...
+    /// Add a measurement register to the attestation capabilities.
+    pub fn add_measurement_register(
+        &mut self,
+        register: MeasurementRegisterDescriptor,
+        index: usize,
+    ) -> Result<&mut AttestationCapabilities> {
+        // We do not allow for a sparse measurement register array.
+        if index + 1 > (self.static_measurements + self.runtime_measurements) as usize {
+            return Err(Error::Overflow);
+        }
```
sameo, line 95: @abrestic-rivos FYI I added this check after you approved the PR.
We want to make sure the SBI implementation does not create a sparse array of measurement registers.

atulkharerivos (approved):

sbi/src/error.rs
abrestic-rivos: So this is really an internal error and there's nothing the caller can do about it, right? If that's the case, would `Failed` be more appropriate?

sameo: That's less intrusive, yes. I'll introduce the 2 new errors with the PRs introducing the measurement registers read and extension implementations.

PR #79, opened 2022-08-29
Fix an out-of-bounds store

abrestic-rivos:
`_fetch_guest_instruction` actually ends up writing 6 bytes when fetching a 32-bit instruction due to use of `sw` rather than `sh` for making a 16-bit store. Oops. This resulted in corruption of the neighboring 16-bit value on the stack, which was breaking `run_debian`.

Patch 2 is a trivial clippy fix, courtesy of sameo@.


dgreid (approved):
Good catch!

PR #82, opened 2022-08-30
salus: added save/restore for vector context on context switch

stillson:
-added support for compiler target with vector extension
-various additions for vector registers
-runtime checking that we can support register width
-test code to check register restor and save
-extended some error messages in tellus

### DIFF
--- a/drivers/src/cpu.rs
+++ b/drivers/src/cpu.rs
@@ -36,9 +36,11 @@ pub struct CpuInfo {
     has_sstc: bool,
     // True if the Sscofpmf extension is supported.
     has_sscofpmf: bool,
+    // True if the vector extension is supported
+    has_vector: bool,
     // CPU timer frequency.
     timer_frequency: u32,
-    // ISA string as reprted in the device-tree. All CPUs are expected to have the same ISA.
+    // ISA string as reported in the device-tree. All CPUs are expected to have the same ISA.
     isa_string: ArrayString<MAX_ISA_STRING_LEN>,
     // Mapping of logical CPU index to hart IDs.
     hart_ids: ArrayVec<u32, MAX_CPUS>,
@@ -77,6 +79,16 @@ fn intc_phandle_from_cpu_node(dt: &DeviceTree, node: &DeviceTreeNode) -> u32 {
 
 fn isa_string_has_extension(isa_string: &str, extension: &str) -> bool {
     isa_string.split('_').any(|f| f == extension)
+}
+
+fn isa_string_has_base_extension(isa_string: &str, extension: char) -> bool {
+    isa_string
+        .split('_')
+        .next()
+        .unwrap_or("rv32")
+        .chars()
+        .skip(4)
+        .any(|c| c == extension)
 }
 
 impl CpuInfo {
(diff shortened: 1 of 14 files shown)


drivers/src/cpu.rs
```diff
@@ -79,6 +81,16 @@ fn isa_string_has_extension(isa_string: &str, extension: &str) -> bool {
     isa_string.split('_').any(|f| f == extension)
 }
 
+fn isa_string_has_base_extension(isa_string: &str, extension: char) -> bool {
+    isa_string
+        .split('_')
+        .next()
+        .expect("ISA string too short")
```
atulkharerivos, line 88: Should we panic if no features are detected?

riscv-regs/src/regs.rs
```diff
@@ -117,3 +117,22 @@ impl GeneralPurposeRegisters {
 #[derive(Default)]
 #[repr(C)]
 pub struct FloatingPointRegisters([u64; 32]);
+
+/// The vector register file. We don't expect to directly interact with a guest's vector state
+/// other than for saving/restoring the registers, so simply treat the register file as an array
+/// of 256b values. This actually depends on the vlenb csr, so if the register is greater than 256
+/// bits (i.e. the vlenb csr is greater than 32) we will need to increase this.
+
+// the width of a vector register in bytes
```
atulkharerivos, line 126: Nit: s/the/The/

src/main.rs
```diff
@@ -402,6 +404,28 @@ pub fn setup_csrs() {
     trap::install_trap_handler();
 }
 
+#[cfg(target_feature = "v")]
+fn verify_vector_registers() {
+    // Because we just ran setup_csrs(), we know vectors are off
+    // Turn vectors on
+    CSR.sstatus.read_and_set_bits(0x200);
```
atulkharerivos, line 411: Nit: Consider defining a const value for 0x200.

src/vectors.S
```diff
@@ -0,0 +1,112 @@
+.section .text.init
+.global _run_guest_vector,_guest_exit_vector
+
+// t5 -> cached sstatus
+// restore vectors for guest
```
atulkharerivos, line 5: Nit (here and elsewhere): Capitalize first letter of sentence if applicable.

drivers/src/cpu.rs
stillson: is that even possible? next() will just return rv64i....

there is always at least one, isn't there?

riscv-regs/src/regs.rs
stillson: fixed

src/main.rs
stillson: There should be a way of getting from the tock register definition, but I couldn't figure it out (in 2 minutes) and I knew 0x200 off the top of my head. 

Will Fix

drivers/src/cpu.rs
atulkharerivos: Sure, it's nearly certain that we'll have at least one feature, but what I meant is that we probably don't need to go out of our way to panic (when we can just easily mark the feature as not detected).

src/vectors.S
stillson: fixed

PR #83, opened 2022-08-30
Remove dead FP restore code

abrestic-rivos:
Should've been removed as part of #80 now that FP save/restore is in separate functions.


dgreid (approved):
Thanks!

Issue #94, opened 2022-09-08
Use shared memory for communicating vCPU state between TSM and host


abrestic-rivos:
Handling a TVM exit currently requires a bunch of trapping into the TSM via the `TvmCpu{Get,Set}Register` calls. Consider using a region of shared memory, set aside at TVM creation time, for communicating this state between the host and TSM. This shared structure would be updated by the TSM on a TVM exit, and would largely be write-only by the TSM. The host is free to otherwise do whatever it wants to the memory as there are only a couple situations where the TSM would ever need to read something the host has written (e.g. handling an emulated MMIO read).

It could look something like:

```
struct vcpu_shared_state {
    // Written by host; read by TSM at TVM finalization.
    u64 entry_sepc;
    u64 entry_a1;

    // Filtered trap CSR values written by TSM on TVM exit.
    u64 scause;
    u64 stval;
    u64 htval;
    u64 htinst;
    u64 vstimecmp;

    ....
};
```

cc @atishp04

PR #96, opened 2022-09-12
Use shared-memory to handle vCPU exits

abrestic-rivos:
This patch defines the shared-memory structure used to communicate non-confidential state between the TSM and host, and the API to register it.

Two open questions:
 - Should we try to replicate more standard RISC-V registers, or continue to use the custom fields/values we've defined previously? i.e. use sanitized `scause`, `htval`, `htinst`, etc values rather than our custom `exit_cause0`/`exit_cause1` plus associated enums. The former will have better re-use with any sort of nested virtualization extension (and existing KVM code in general), while the latter will be more space efficient and theoretically mean less work for the OS/VMM to do on vCPU exit (e.g. no need to decode instructions again, or look up if a page fault is confidential or not). I ended up going with the standard RISC-V register approach here to reduce the overall surface area of the API. (I realize that my re-use of `htinst` for WFI isn't standard-compliant.)
  - Do we want to reserve entire pages for these shared memory structs, or let the OS/VMM place them wherever they want (in non-confidential memory, of course)? I'm leaning towards the latter since it's mostly informational for the OS/VMM, i.e. causing the structs to overlap or otherwise place them somewhere where they could be freely read/written won't violate the confidentiality of a TVM.

Thoughts?

### DIFF
--- a/sbi/src/tsm.rs
+++ b/sbi/src/tsm.rs
@@ -171,6 +171,115 @@ impl TvmCpuRegister {
     }
 }
 
+/// Identifies a register set in the vCPU shared-memory state area layout.
+///
+/// Register sets are roughly grouped by the specification or extension that defines them.
+/// While it is likely that registers from multiple register sets will be used in the course
+/// of handling a vCPU exit, the register sets are grouped in this way in order to allow for
+/// extensibility by future extensions, and to avoid being overly prescriptive with regards
+/// to TSM or hypervisor implementation.
+#[repr(u16)]
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub enum RegisterSetId {
+    /// General purpose registers.
+    Gprs = 0,
+    /// Supervisor CSRs.
+    SupervisorCsrs = 1,
+    /// HS-level CSRs.
+    HypervisorCsrs = 2,
+}
+
+impl RegisterSetId {
+    /// Returns the `RegisterSetId` corresponding to `val`.
+    pub fn from_raw(val: u16) -> Result<Self> {
+        match val {
+            0 => Ok(RegisterSetId::Gprs),
+            1 => Ok(RegisterSetId::SupervisorCsrs),
+            2 => Ok(RegisterSetId::HypervisorCsrs),
+            _ => Err(Error::InvalidParam),
+        }
+    }
+
+    /// Returns the size in bytes of the register set structure corresponding to this ID.
+    pub fn struct_size(&self) -> usize {
+        match self {
+            RegisterSetId::Gprs => core::mem::size_of::<Gprs>(),
+            RegisterSetId::SupervisorCsrs => core::mem::size_of::<SupervisorCsrs>(),
+            RegisterSetId::HypervisorCsrs => core::mem::size_of::<HypervisorCsrs>(),
+        }
+    }
+}
+
+/// Specifies the location of a particular register set in the vCPU shared-memory state area.
+#[repr(C)]
+#[derive(Clone, Copy, Debug, Default)]
+pub struct RegisterSetLocation {
+    /// The type of the register set, identified by the `RegisterSetId` enum.
+    pub id: u16,
+    /// The version of the register set structure.
+    ///
+    /// As the definition of a particular register set may change across revisions of this API,
+    /// this version number can be used to identify the specific layout that is being used.
+    pub version: u16,
+    /// The offset of the register set from the start of the vCPU's shared-memory state area.
+    pub offset: u32,
+}
+
+/// General purpose registers. Structure for register sets of type `RegisterSetId::Gprs`.
+///
+/// The TSM will always read or write the minimum number of registers in this set to complete
+/// the requested action, in order to avoid leaking information from the TVM.
+///
+/// The TSM will write to these registers upon return from `TvmCpuRun` when:
+///  - The vCPU takes a store guest page fault in an emulated MMIO region.
+///  - The vCPU makes an ECALL that is to be forwarded to the host.
+///
+/// The TSM will read from these registers when:
+///  - The vCPU takes a load guest page fault in an emulated MMIO region.
+///  - The host calls `TvmFinalize`, latching the entry point argument (stored in 'A1') for the
+///    TVM's boot vCPU.
+#[repr(C)]
+#[derive(Default)]
+pub struct Gprs(pub [u64; 32]);
+
+/// Supervisor CSRs. Structure for register sets of type `RegisterSetId::SupervisorCsrs`.
+#[repr(C)]
+#[derive(Default)]
+pub struct SupervisorCsrs {
+    /// Initial SEPC value (entry point) of a TVM vCPU. Latched for the TVM's boot vCPU at
+    /// `TvmFinalzie`; ignored for all other vCPUs.
+    pub sepc: u64,
+    /// SCAUSE value for the trap taken by the TVM vCPU. Written by the TSM upon return from
+    /// `TvmCpuRun`.
+    pub scause: u64,
+    /// STVAL value for guest page faults or virtual instruction exceptions taken by the TVM vCPU.
+    /// Written by the TSM upon return from `TvmCpuRun`.
+    ///
+    /// Note that guest virtual addresses are not exposed by the TSM, so only the 2 LSBs will
+    /// ever be non-zero for guest page fault exceptions.
+    pub stval: u64,
+}
+
+/// HS-level CSRs. Structure for register sets of type `RegisterSetId::HypervisorCsrs`.
+#[repr(C)]
+#[derive(Default)]
+pub struct HypervisorCsrs {
+    /// HTVAL value for guest page faults taken by the TVM vCPU. Written by the TSM upon return
+    /// from `TvmCpuRun`.
+    pub htval: u64,
+    /// HTINST value for guest page faults or virtual instruction exceptions taken by the TVM vCPU.
+    /// Written (in certain circumstances) by the TSM upon return from `TvmCpuRun`.
+    ///
+    /// The TSM will only write `htinst` in the following cases:
+    ///  - MMIO load page faults. The value written to the register in `gprs` corresponding to the
+    ///    'rd' register in the instruction will be used to complete the load upon the next call to
+    ///    `TvmCpuRun` for this vCPU.
+    ///  - MMIO store page faults. The TSM will write the value to be stored by the vCPU to the
+    ///    register in `gprs` corresponding to the 'rs2' register in the instruction upon return
+    ///    from `TvmCpuRun`.
+    pub htinst: u64,
+}
+
 /// Provides the state of the confidential VM supervisor.
 #[repr(u32)]
 #[derive(Copy, Clone, PartialEq, Eq, Default)]
@@ -328,8 +437,33 @@ pub enum TeeFunction {
         /// a1 = vCPU id
         vcpu_id: u64,
     },
-    /// Adds a vCPU with ID `vcpu_id` to the guest `guest_id`. vCPUs may not be added after the TVM
-    /// is finalized.
+    /// Writes the layout that the TSM will use for the vCPU shared-memory state area for vCPUs
+    /// of the TVM identified by `guest_id` to the non-confidential physical address `layout_addr`.
+    /// The caller uses this to discover the size and layout of the structure that will be used
+    /// to communicate vCPU state in shared-memory.
+    ///
+    /// Returns the number of bytes written to `layout_addr` upon success, or an error if
+    /// `layout_addr` is invalid or `layout_len` is insufficiently large to describe the entire
+    /// layout.
+    ///
+    /// a6 = 21
+    TvmCpuGetMemLayout {
+        /// a0 = guest id
+        guest_id: u64,
+        /// a1 = base physical address of the array of `RegisterSetLayout` structs
+        layout_addr: u64,
+        /// a2 = length of the `RegisterSetLayout` array in bytes
+        layout_len: u64,
+    },
+    /// Adds a vCPU with ID `vcpu_id` to the guest `guest_id`, registering `shared_page_addr` as
+    /// the location of the shared-memory state area for this vCPU.
+    ///
+    /// `shared_page_addr` must be page-aligned and point to a sufficient number of non-confidential
+    /// pages to hold a structure with the layout specified by `TvmCpuGetMemLayout`. These pages
+    /// are "pinned" in the non-confidential state (i.e. cannot be converted to confidential) until
+    /// the TVM is destroyed.
+    ///
+    /// vCPUs may not be added after the TVM is finalized.
     ///
     /// a6 = 8
     TvmCpuCreate {
@@ -337,9 +471,13 @@ pub enum TeeFunction {
         guest_id: u64,
         /// a1 = vCPU id
         vcpu_id: u64,
+        /// a2 = page address of shared state structure
+        shared_page_addr: u64,
     },
     /// Sets the register identified by `register` to `value` in the vCPU with ID `vcpu_id`. See
     /// the defintion of `TvmCpuRegister` for details on which registers are writeable and when.
+    ///
+    /// TODO: Remove once communication of vCPU state over shared memory is implemented.
     ///
     /// a6 = 9
     TvmCpuSetRegister {
@@ -425,6 +563,8 @@ pub enum TeeFunction {
     /// Gets the register identified by `register` in the vCPU with ID `vcpu_id`. See the definition
     /// of `TvmCpuRegister` for details on which registers are readable and when. The contents of
     /// the specified register are returned upon success.
+    ///
+    /// TODO: Remove once communication of vCPU state over shared memory is implemented.
     ///
     /// a6 = 16
     TvmCpuGetRegister {
@@ -527,6 +667,7 @@ impl TeeFunction {
             8 => Ok(TvmCpuCreate {
                 guest_id: args[0],
                 vcpu_id: args[1],
+                shared_page_addr: args[2],
             }),
             9 => Ok(TvmCpuSetRegister {
                 guest_id: args[0],
@@ -585,6 +726,11 @@ impl TeeFunction {
                 num_pages: args[3],
                 guest_addr: args[4],
             }),
+            21 => Ok(TvmCpuGetMemLayout {
+                guest_id: args[0],
+                layout_addr: args[1],
+                layout_len: args[2],
+            }),
             _ => Err(Error::NotSupported),
         }
     }
@@ -619,6 +765,7 @@ impl SbiFunction for TeeFunction {
             TvmCpuCreate {
                 guest_id: _,
                 vcpu_id: _,
+                shared_page_addr: _,
             } => 8,
             TvmCpuSetRegister {
                 guest_id: _,
@@ -677,6 +824,11 @@ impl SbiFunction for TeeFunction {
                 num_pages: _,
                 guest_addr: _,
             } => 20,
+            TvmCpuGetMemLayout {
+                guest_id: _,
+                layout_addr: _,
+                layout_len: _,
+            } => 21,
         }
     }
 
@@ -708,6 +860,7 @@ impl SbiFunction for TeeFunction {
             TvmCpuCreate {
                 guest_id,
                 vcpu_id: _,
+                shared_page_addr: _,
             } => *guest_id,
             TvmCpuSetRegister {
                 guest_id,
@@ -761,6 +914,11 @@ impl SbiFunction for TeeFunction {
                 num_pages: _,
                 guest_addr: _,
             } => *guest_id,
+            TvmCpuGetMemLayout {
+                guest_id,
+                layout_addr: _,
+                layout_len: _,
+            } => *guest_id,
             _ => 0,
         }
     }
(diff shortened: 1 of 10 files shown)


atishp04:
I prefer the standard register approach for the reasons you already mentioned. It aligns well with the nested virtualization. 

For the 2nd, I can see one benefit for reserving entire page. We can leverage the shared_page construct for this. Otherwise, we end up in confidential memory region, shared memory region, mmio region and a shared vcpu region. Thoughts ?

dgreid:
For the standard register approach, what's the impact of having to redo instruction decode? Will that entail SBI calls for HL* instructions to read guest memory?

abrestic-rivos:
> For the 2nd, I can see one benefit for reserving entire page. We can leverage the shared_page construct for this. Otherwise, we end up in confidential memory region, shared memory region, mmio region and a shared vcpu region. Thoughts ?

We can still use the same page sharing construct without reserving an entire page by marking each page the struct occupies a shared page and bumping the shared refcount. The shared page refcount is dropped when the TVM is destroyed. This allows the host to pack multiple structs in a single page and prevents the page from being converted to confidential memory.

> For the standard register approach, what's the impact of having to redo instruction decode? Will that entail SBI calls for HL* instructions to read guest memory?

Pretty minor. If there's a need for the host to have the faulting instruction (right now just MMIO load/store and WFI virtual instruction), the TSM will stuff it in `htinst`. So the host would have to decode the instruction (load vs store, width) from the `htinst` value rather than using the `TvmMmioOpCode` enum. My plan was to turn all MMIO load/store faults into loads to or stores from register A0 to make things simple to avoid leaking unnecessary information to the host.

For example, say a TVM takes an MMIO load fault at guest physical address `X` from the instruction `LD a5, 16(s1)`. The TSM:
 - writes `X` to `htval`/`stval`
 - writes "guest load page fault" to `scause`
 - writes the transformed instruction `LD a0` to `htinst`
 - reads the value written by the host to `a0` and puts in the guest's `a5` on the next `TvmCpuRun`.

One thing to note here (and as I alluded to in the original PR description) is that my planned use of `htinst` for virtual instruction exceptions (just `WFI` for now) is not spec-compliant: the spec doesn't allow standard instructions to be written to the `htinst` CSR on virtual instruction exceptions. Nested virt might take a different route, e.g. by providing a call to do the equivalent of an `HLVX`, but that requires the TSM to keep track of more state and in general seems like overkill given the very limited cases where we'd want to forward a virtual instruction exception to the host.

atulkharerivos:
Don't have anything more to add to what's already been said. I really like the overall idea, and how we are scrubbing information by not divulging unnecessary execution stream details (like always reporting a0 for LD for example).

dgreid:
Looks good to me, thanks. htinst seems Ok for now (we should bring up using it for more in nested virt with RVI). If that isn't acceptable we can always define an extra value similar to how stval2 was added

abrestic-rivos:
Updated the PR with the implementation of shared-memory vCPU communication scheme. Tried to split it up into more bite-sized patches, at the cost of having to support both the shared-memory and virtual-register-TEECALL path in the middle of the series.

Note that I decided to go with reporting the faulting instruction in `stval` for virtual instruction exceptions. Implementations are allowed to write the faulting instruction on illegal instruction exceptions to `stval`, so doing the same for virtual instruction exceptions seems like a bit less of a bending of the rules than using `htinst`.

- Patch 1 defines the new struct and SBI interface.
- Patch 2 adds a helper for making volatile accesses to the shared-memory struct.
- Patches 3 & 4 add support for pinning the pages backing the shared-memory structs in the "Shared" page state.
- Patches 5 & 6 add support within Salus for making use of the shared-memory struct, temporarily using it to shadow the virtual registers from the old interface.
- Patches 7 & 8 convert the host VM run loop (in Salus) and guest VM run loop (in Tellus) to use the new shared-memory interface.
- Patch 9 removes the old virtual-register interface.

There are more cleanups to be done (e.g. no need to go through an intermediate MMIO op enum, or to propagate conf vs non-conf page fault types), but we'll leave those for follow-on patches.

Closes #94

sbi/src/tsm.rs
```diff
@@ -171,6 +171,51 @@ impl TvmCpuRegister {
...
+    /// The TSM will write to these registers upon return from `TvmCpuRun` when:
+    ///  - The vCPU takes a store guest page fault in an emulated MMIO region.
+    ///  - The vCPU makes an ECALL that is to be forwarded to the host.
+    ///
+    /// The TSM will read from these registers when:
+    ///  - The vCPU takes a load guest page fault in an emulated MMIO region.
+    ///  - The host calls `TvmFinalize`, latching the entry point argument (stored in 'A1') for the
+    ///    TVM's boot vCPU.
+    pub gprs: [u64; 32],
+}
```
atishp04, line 50: As per our discussion today morning, we probably need more h-mode CSRs in the shared state. 
How about defining one SBI call just to setup the shared memory at the beginning ? 

The TSM/KVM can read/write the specific data in respective functions. The function definition spec will specify which entry to write/read when shared memory approach is available.

atishp04: This will probably reused in nested virtualization acceleration. Can we make it more flexible in terms of shared memory layout so that the guest hypervisor can specify which CSRs it wants in the shared memory ?

For nested virt, we need VSmode CSRs in the shared memory as well.

abrestic-rivos: > As per our discussion today morning, we probably need more h-mode CSRs in the shared memory.
How about defining one SBI call just to setup the shared memory at the beginning ?

When is "at the beginning"? Are you suggesting having a shared-memory region per physical CPU rather than per TVM vCPU? I'm not sold on that being the cleanest approach:
 - How would CSRs that are not typically set per vCPU (e.g. `hedeleg`) get set? Some separate SBI interface? (Does such an interface even make sense for AP-TEE given that the TSM will toss the values anyway?)
 - How would discovery of WARL fields (e.g. `hgatp.mode`) work?
 - Migration of vCPUs between physical CPUs would get more expensive as you'd need to copy between the the shared memory areas for the physical CPUs. Not a big deal for TVMs since there's very little useful state that's kept in shared memory, but for nested virt there's potentially much more architectural state that would be stored in shared memory between guest & host hypervisors.

> This will probably reused in nested virtualization acceleration. Can we make it more flexible in terms of shared memory layout so that the guest hypervisor can specify which CSRs it wants in the shared memory ?

Could we just do two separate struct layouts for TVMs vs nested virt, or maybe a `union`? It would be nice to avoid user-provided struct layouts as that seems error prone.

atishp04: > > As per our discussion today morning, we probably need more h-mode CSRs in the shared memory.
> > How about defining one SBI call just to setup the shared memory at the beginning ?
> 
> When is "at the beginning"? Are you suggesting having a shared-memory region per physical CPU rather than per TVM vCPU? I'm not sold on that being the cleanest approach:
> 

It will still be per vCPU. I got confused with rust syntax. You are passing the shared memory address in the vcpu_create
which is what I was suggesting too. Sorry for the confusion :). 

> * How would CSRs that are not typically set per vCPU (e.g. `hedeleg`) get set? Some separate SBI interface? (Does such an interface even make sense for AP-TEE given that the TSM will toss the values anyway?)
> * How would discovery of WARL fields (e.g. `hgatp.mode`) work?
> * Migration of vCPUs between physical CPUs would get more expensive as you'd need to copy between the the shared memory areas for the physical CPUs. Not a big deal for TVMs since there's very little useful state that's kept in shared memory, but for nested virt there's potentially much more architectural state that would be stored in shared memory between guest & host hypervisors.
> 
> > This will probably reused in nested virtualization acceleration. Can we make it more flexible in terms of shared memory layout so that the guest hypervisor can specify which CSRs it wants in the shared memory ?
> 
> Could we just do two separate struct layouts for TVMs vs nested virt, or maybe a `union`? It would be nice to avoid user-provided struct layouts as that seems error prone.

As we are talking about all Hmode CSRs, fixing the layout will end up getting revisions over time.
Instead, if we make it flexible, the interface doesn't change. We need vsmode CSRs in shared memory for nested virt.
Thus, the hypervisor shared memory extension can be carved out of TEE and used across multiple use cases.

abrestic-rivos: > As we are talking about all Hmode CSRs, fixing the layout will end up getting revisions over time.
Instead, if we make it flexible, the interface doesn't change. We need vsmode CSRs in shared memory for nested virt.
Thus, the hypervisor shared memory extension can be carved out of TEE and used across multiple use cases.

Given that there will be two "main" layouts -- AP-TEE & nested virtualization -- I'm not sure the complexity of having a completely customizable memory layout for vCPU shared-memory state makes sense. Further, it seems like it should be the host (TSM or host hypervisor) that specifies the layout as it will naturally already know what the hardware capabilities are. That said, here are some approaches I can think of:

1. Make the shared-state structs append only so that the virtualization mode (AP-TEE vs nested) + struct length (reported by host) + (possibly) host version are enough for the guest to determine the layout. That's basically the scheme this series as implemented would employ.
2. Let the guest (or host) specify the entire struct layout via an SBI call that takes a list of <register identifer> + <offset> pairs. I think this is is overkill given the relatively limited option space, and it makes things more expensive for whoever is the "receiver" of the custom layout as they'll need to go through an intermediate lookup table when trying to find registers in the shared-memory space.
3. Adopt a command-buffer-like interface where the guest specifies a list of registers it wants to read/write the next time a vCPU is run. A little cleaner IMO than (2) and with the same flexibility, though we now have to pay the cost of validating and copying the struct on each call to `vcpu_run()`.
4. Have the host report the struct layout in a "capability list" format, similar to PCIe or VFIO. Each capability would identify the type and location of a register set in the shared memory space (register set meaning a group of related bits of architectural state, e.g. "GPRs" or "H* CSRs"). Similar to (2) I guess, but more coarse-grained.

Open to suggestions, but I think (4) is a reasonable compromise on extensibility. For PoC purposes though I fell like (1) is sufficient to get started.

atishp04: > > As we are talking about all Hmode CSRs, fixing the layout will end up getting revisions over time.
> > Instead, if we make it flexible, the interface doesn't change. We need vsmode CSRs in shared memory for nested virt.
> > Thus, the hypervisor shared memory extension can be carved out of TEE and used across multiple use cases.
> 
> Given that there will be two "main" layouts -- AP-TEE & nested virtualization -- I'm not sure the complexity of having a completely customizable memory layout for vCPU shared-memory state makes sense. Further, it seems like it should be the host (TSM or host hypervisor) that specifies the layout as it will naturally already know what the hardware capabilities are. That said, here are some approaches I can think of:
> 
> 1. Make the shared-state structs append only so that the virtualization mode (AP-TEE vs nested) + struct length (reported by host) + (possibly) host version are enough for the guest to determine the layout. That's basically the scheme this series as implemented would employ.
> 2. Let the guest (or host) specify the entire struct layout via an SBI call that takes a list of  +  pairs. I think this is is overkill given the relatively limited option space, and it makes things more expensive for whoever is the "receiver" of the custom layout as they'll need to go through an intermediate lookup table when trying to find registers in the shared-memory space.
> 3. Adopt a command-buffer-like interface where the guest specifies a list of registers it wants to read/write the next time a vCPU is run. A little cleaner IMO than (2) and with the same flexibility, though we now have to pay the cost of validating and copying the struct on each call to `vcpu_run()`.
> 4. Have the host report the struct layout in a "capability list" format, similar to PCIe or VFIO. Each capability would identify the type and location of a register set in the shared memory space (register set meaning a group of related bits of architectural state, e.g. "GPRs" or "H* CSRs"). Similar to (2) I guess, but more coarse-grained.
> 

I was thinking along the lines of 4 as well. But instead of host reporting the capability list format, I was relying on the guest(KVM in VS mode) to set it. 

The SBI interface may just define something like:

H_CSR_REGION_ID 0x0
VS_CSR_REGION_ID 0x1
GPR_CSR_REGION_ID 0x2

sbi_smem_setup_region(unsigned long shared_memory_area, unsigned long region_id, unsigned long region_offset)
The guest will have to make multiple SBI call for set the offset of different regions. 

However, I am yet to find a clean solution how to find the offset of each register in that region. There are few options

1. Define it in the spec. Simple but the spec needs to be modified with future extensions.
2. Sequential based on CSR Number defined in priv spec. Confusing and not all H CSRs in the same region.
3. Keep some metadata in the beginning of shared memory region to report which CSRs available in the region.

#3 is probably similar to the capability list you are talking about but at individual register granularity. 

> Open to suggestions, but I think (4) is a reasonable compromise on extensibility. For PoC purposes though I fell like (1) is sufficient to get started.

rsahita:
just caught up with some of the discussion on this PR so far - like Atul, will wait for Atish's questions to get resolved. Until then @abrestic-rivos I wanted you to also take a look (if you have not) at the GHCB proposed for AMD SEV-SNP and the GHCI for TDX - also please cc me on such PRs so I can chime in; I happened to find the PR link via a comment  in the SBI extension PR :-)

We may want to allocate minimally a page, and setup a versioned structure so that it can expand in the future as needed. I agree with your approach of using the standard risc-v registers with some convention around whats exposed to the VMM and what is scrubbed. Should also specify how information is returned to the TVM via this interface for TVM requesting services from the host.

abrestic-rivos:
> just caught up with some of the discussion on this PR so far - like Atul, will wait for Atish's questions to get resolved. Until then @abrestic-rivos I wanted you to also take a look (if you have not) at the GHCB proposed for AMD SEV-SNP and the GHCI for TDX - also please cc me on such PRs so I can chime in; I happened to find the PR link via a comment in the SBI extension PR :-)

Yeah, I've seen GHCB/GHCI and they're related, but they're for (AFAICT) direct guest <-> VMM communication via shared memory whereas what I'm proposing here is TSM <-> VMM communication of (scrubbed) guest state via shared memory (in an effort to minimize the burden on the guest and host vs native or legacy virtualization).
 
> Should also specify how information is returned to the TVM via this interface for TVM requesting services from the host.

So my thought here is that the TVM-requesting-services-from-the-host case would look like an ECALL exit to the host and that the host would put the return values in A0/A1 in the shared memory block as it would for an ordinary ECALL. The TSM would then update the TVM's actual A0/A1 with these values the next time the TVM is run. This works for simple cases where all the host needs to do is to return a pair of integers, but would be tricky for calls that require access to TVM memory. I'm not sure we have any planned use cases that would require the latter and wouldn't already be using virtio, though.

sbi/src/tsm.rs
atishp04: > > I was thinking along the lines of 4 as well. But instead of host reporting the capability list format, I was relying on the guest(KVM in VS mode) to set it.
> 
> Why have the guest set the layout instead of the host/TSM? That would avoid the whole dance with the guest having to probe which region types are supported in individual SBI calls.
> 

It aligns well with other shared memory based SBI extensions where the guest owns the shared memory.
If the TSM defines the offsets of each region, TSM have to own the shared memory because it only knows how much to allocate. This is also fine. We had a similar debate happening with STA proposal. There was no clear answer :).

> > However, I am yet to find a clean solution how to find the offset of each register in that region. There are few options
> > 
> > 1. Define it in the spec. Simple but the spec needs to be modified with future extensions.
> > 2. Sequential based on CSR Number defined in priv spec. Confusing and not all H CSRs in the same region.
> > 3. Keep some metadata in the beginning of shared memory region to report which CSRs available in the region.
> 
> I was thinking (1) actually, possibly using an (extension, version) pair to identify the layout, e.g. (H-extension, v1) would refer to a layout containing the H-extension CSRs as of v1.0 the hypervisor ISA. Sstc, AIA, etc would have their own entries.

Given how RVI manages the version, I wouldn't rely on versioning information from misa string.
But we can always define the layout in the spec for each group of CSRs. The TSM can return a capability list saying which groups are supported.

The downside is that, we may have to change the spec with future extensions. I think that should be okay ?

In that case, the SBI interface defines each group and its layout.

H_CSR_REGION_ID 0x0
VS_CSR_REGION_ID 0x1
GPR_CSR_REGION_ID 0x2
AIA_CSR_REGION_ID 0x3
SSTC_CSR_REGION_ID 0x4

sbi_virt_get_capability() can return a bitmask of each region supported by the native hypervisor.
The register layout in each region will be defined in the specification.

atishp04: The mechanism to find out the offset of each region depends upon which approach we prefer (guest vs hypervisor owned shared memory)

abrestic-rivos: > If the TSM defines the offsets of each region, TSM have to own the shared memory because it only knows how much to allocate. This is also fine. We had a similar debate happening with STA proposal. There was no clear answer :).

So I kinda prefer the idea of the TSM or native hypervisor owning the layout because:
- it minimizes the complexity on the TSM side, which is a TCB element for the TVM
- the TSM or native hypervisor already knows what the hardware natively supports
- between the TSM and OS/VMM, the TSM is likely to be updated more infrequently and thus have a more "stable" interface
- there's overall less work to be done w.r.t. validation of the shared memory layout be either party

As far as communicating the layout goes, I was thinking of having a `vcpu_get_mem_layout()` SBI call that returned the layout of the shared memory region as an array of (region id, offset) pairs. This also tells the OS/VMM how much memory it needs to dedicate for each vCPU's shared memory region.

> The downside is that, we may have to change the spec with future extensions. I think that should be okay ?

Yeah, I think this is inevitable. But by versioning the various region types we can at least make it easier to maintain compatibility in software.

Anyway, let me try implementing this and we'll see what it looks like.

abrestic-rivos:
Ok, I've reworked the interface such that the layout is more extensible. The TSM defines the layout, and the host enumerates it through a separate `TvmCpuGetMemLayoutCall`. Take a look and let me know what you all think before I go down the path of updating the rest of the series.

BTW, I'm aware that the register sets I'm defining are missing many of the things one would want for nested virtualization. I'm not trying to solve that problem here, but rather provide a mechanism that could potentially be reused by nested virt. And there are definitely still some interactions on the nested virt set (particularly around interrupts) that need to be worked through before we can define an interface.

sbi/src/tsm.rs
```diff
@@ -171,6 +171,85 @@ impl TvmCpuRegister {
...
+
+/// Specifies the location of a particular register set in the vCPU shared-memory state area.
+#[repr(C)]
+#[derive(Clone, Copy, Debug)]
+pub struct RegisterSetLocation {
+    /// The type of the register set, identified by the `RegisterSetId` enum.
+    pub id: u16,
+    /// The version of the register set structure.
+    pub version: u16,
```
atishp04, line 202: Just additional clarification for the documentation: the layout if each register set(e.g.) will be defined in the specification with a version. The version allows us to modify the register set layout in the future.

sbi/src/tsm.rs
```diff
@@ -171,6 +171,85 @@ impl TvmCpuRegister {
...
+///  - The host calls `TvmFinalize`, latching the entry point argument (stored in 'A1') for the
+///    TVM's boot vCPU.
+#[repr(C)]
+#[derive(Default)]
+pub struct Gprs([u64; 32]);
+
+/// HS-level CSRs related to trap handling.
+#[repr(C)]
+#[derive(Default)]
+pub struct HypervisorTrapCsrs {
```
atishp04, line 224: So we are grouping CSRs based on path where it is being accessed. This may be bit confused as the gprs or VS registers will probably be based on type.

May be define it in terms of context ?

GPR - Guest CPU Context
HypervisorTrapCsrs - Trap context
HTIMEDELTA - TIME Context
VS mode CSRs & HVIP - load_put_sync context

This aligns well with KVM. The downside is that other hypervisor may save/restore particular CSRs at different context.

How about just grouping them based on type ?

GPR, HMODE, VSMODE, SMODE

sbi/src/tsm.rs
abrestic-rivos: Expanded the comment to include this.

sbi/src/tsm.rs
abrestic-rivos: > How about just grouping them based on type ?
> 
> GPR, HMODE, VSMODE, SMODE

Changed to group by spec/extension (and added a comment). I went the other way at first because having a bunch of register sets with only 1 or 2 CSRs seemed like overkill. But I agree that this grouping is more extensible.

sbi/src/tsm.rs
```diff
@@ -171,6 +171,103 @@ impl TvmCpuRegister {
...
+    /// The version of the register set structure.
+    ///
+    /// As the definition of a particular register set may change across revisions of this API,
+    /// this version number can be used to identify the specific layout that is being used.
+    pub version: u16,
+    /// The offset of the register set from the start of the vCPU's shared-memory state area.
+    pub offset: u32,
+}
+
+/// General purpose registers. Structure for register sets of type `RegisterSetId::Gprs`.
```
dgreid, line 219: A comment here that the TSM will only read/write the minimum subset of registers needed to complete an action(and not leak other guest data) would be helpful.

dgreid (approved):
lgtm if atishp is OK with it.

atishp04 (approved):
LGTM.

abrestic-rivos:
Ok, updated the PR with the reset of the implementation. The remaining patches look roughly the same, except updated to use the new custom-layout scheme.

sbi/src/tsm.rs
atulkharerivos: Believe this comment can be marked as resolved.

atulkharerivos (approved):

Issue #65, opened 2022-08-22
Avoid copies / blowing up the stack when initializing large structures


abrestic-rivos:
We have a few large structs (`Vm`, `VmCpu`) that we place in `Page` containers but, in the absence of guaranteed copy elision, we may end up with the struct being constructed on the stack and then copied into the page. And in fact, this is exactly what appears to be happening currently with `VmCpu`. This is going to become more of a problem as we add more per-vCPU state: vector state, PMU, IMSIC, etc.

We should consider initializing these structs in-place. Unfortunately Rust does not have support for a "placement new" or "emplace" so it looks like the best we can do right now is `MaybeUninit` + a bunch of `unsafe`.

abrestic-rivos:
A couple notes:
- `Vm::finalize()` and `VmPages::finalize()` effectively turn into a huge `memcpy()`. I don't see a good way to avoid this with the type-state pattern; we may need an alternate approach to type-safe `Vm` states.
- The [moveit](https://docs.rs/moveit/latest/moveit/index.html#) crate, or something like it, might be of use to us here for in-place construction.

PR #98, opened 2022-09-15
Use wrapper types to encode Vm & VmPages state

abrestic-rivos:
While encoding the state of the VM as a type parameter in `Vm` and `VmPages` creates clean separation between the APIs in each state, it means that transitions between states require consuming (and as a result, copying) these large structs. Introduce wrapper `VmRef` / `VmPagesRef` types which expose the appropriate functionality for the current state of the VM instead, which still gets us the nice API separation with an extra layer of indirection. This is a pre-requisite to #65, and will enable us to do in-place initialization of these structs.

Patches 1-3 are small cleanups which enable patch 4 (wrapper state types for `VmPages`) and patch 5 (wrapper state types for `Vm`).

### DIFF
--- a/src/guest_tracking.rs
+++ b/src/guest_tracking.rs
@@ -3,14 +3,13 @@
 // SPDX-License-Identifier: Apache-2.0
 
 use core::marker::PhantomData;
-use core::ops::Deref;
 use page_tracking::collections::{PageArc, PageVec};
 use page_tracking::PageTracker;
 use riscv_page_tables::GuestStagePagingMode;
 use riscv_pages::{InternalClean, Page, PageOwnerId, SequentialPages};
 use spin::{Mutex, RwLock, RwLockReadGuard};
 
-use crate::vm::{Vm, VmStateFinalized, VmStateInitializing};
+use crate::vm::{AnyVm, FinalizedVm, InitializingVm, Vm, VmRef};
 
 /// Guest tracking-related errors.
 #[derive(Debug)]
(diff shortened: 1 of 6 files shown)


src/vm_pages.rs
```diff
@@ -1102,37 +1104,44 @@ impl<T: GuestStagePagingMode> VmPages<T, VmStateInitializing> {
...
-            RawAddr::from(page_addr)
-                .checked_increment(len)
-                .ok_or(Error::AddressOverflow)?,
-        )
-        .ok_or(Error::UnalignedAddress)?;
-        self.regions.add(page_addr, end, VmRegionType::Mmio)
+        self.do_add_region(page_addr, len, VmRegionType::Mmio)
+    }
+
+    /// Adds a PCI BAR memory regino of `len` bytes starting at `page_addr` to this VM's address
```
atulkharerivos, line 1288: s/regino/region

src/vm_pages.rs
```diff
@@ -732,29 +737,72 @@ pub struct VmPages<T: GuestStagePagingMode, S = VmStateFinalized> {
...
+    }
+
     /// Returns the `PageOwnerId` associated with the pages contained in this machine.
     pub fn page_owner_id(&self) -> PageOwnerId {
         self.page_owner_id
     }
 
-    /// Returns the address of the root page table for this VM.
-    pub fn root_address(&self) -> SupervisorPageAddr {
+    /// Returns a `VmPagesRef` to `self` in state `S`. The caller must ensure that `S` matches
```
atulkharerivos, line 769: Do we have some way to enforce this safety requirement? Marking it as unsafe is probably overkill.

src/guest_tracking.rs
```diff
@@ -24,142 +23,124 @@ pub enum Error {
...
-        };
-    }
 }
 
-/// A refernce to a VM in a paritcular state. While this refernce is held the wrapped `Vm` is
-/// guaranteed not to change state.
-pub struct VmRef<'a, T: GuestStagePagingMode, S> {
-    vm: RwLockReadGuard<'a, GuestStateInner<T>>,
-    phantom: PhantomData<S>,
+/// A shared reference to a `Vm` in a particular state. While this refernce is held the wrapped `Vm`
```
atulkharerivos, line 59: s/refernce/reference

src/vm_pages.rs
abrestic-rivos: Yeah, not sure it meets the definition of `unsafe` since it doesn't affect Salus' memory safety. Another way might be to hold a `&VmRef`, but that gets a little messy.

It's worth noting that getting the "wrong" type out of `as_ref()` would require explicitly naming the type, e.g. `let x = vm_pages.as_ref::<VmStateFinalized>()`, which should hopefully be an obvious code smell.

atulkharerivos:
I will take a second look, but it looks good from an initial pass.

src/vm.rs
```diff
@@ -1938,6 +1940,27 @@ impl<T: GuestStagePagingMode> HostVm<T, VmStateInitializing> {
         }
     }
 
+    /// Add PCI BAR pages to the host page tables.
+    pub fn add_pci_pages<I>(&mut self, to_addr: GuestPageAddr, pages: I)
+    where
+        I: ExactSizeIterator<Item = PciBarPage<ConvertedClean>>,
+    {
+        let page_tracker = self.inner.vm_pages.page_tracker();
+        // Unwrap ok since we've donate sufficient PT pages to map the entire address space up front.
```
dgreid, line 2001: nit s/donate/donated/

src/vm_pages.rs
```diff
@@ -730,17 +775,53 @@ impl<T: GuestStagePagingMode, S> VmPages<T, S> {
         Ok(())
     }
 
-    /// Locks `count` 4kB pages starting at `page_addr` for mapping in a region of type
-    /// `region_type`, returning a `VmPagesMapper` that can be used to insert (and measure, if
-    /// necessary) the pages.
-    pub fn map_pages(
+    // Locks `count` 4kB pages starting at `page_addr` for mapping in a region of type
+    // `region_type`, returning a `VmPagesMapper` that can be used to insert (and measure, if
+    // necessary) the pages.
```
dgreid, line 825: nit: the descriptions of the args is lost from the public docs, maybe move to map_zero_pages and reference it from the others?

dgreid (approved):
thanks, lgtm just a couple comment nits.

src/vm.rs
abrestic-rivos: Done.

src/vm_pages.rs
abrestic-rivos: Done.

PR #99, opened 2022-09-16
Attestation SBI: Implement the read and extend measurement functions

sameo:
This PR implements the `ExtendMeasurement` and `ReadMeasurement` functions for the attestation SBI extension.
This will allow for e.g. TVMs to dynamically extend the TCB with post finalisation measurements.

The first commit completes the TCG PCR index description and maps for our platform.
The second commit implements the measurement extension function while the third one implements the measurement read function.


sbi/src/error.rs
```diff
@@ -24,6 +24,8 @@ pub enum Error {
     AlreadyStarted = -7,
     /// Some of the given counters have already been stopped.
     AlreadyStopped = -8,
+    /// The buffer passed as a parameter is invalid.
+    InvalidBuffer = -9,
```
abrestic-rivos, line 28: From the name (and associated comment), I'd assume this was an alias for `InvalidAddress`, but it looks like you're using it for "insufficient buffer length". Maybe change the name to reflect that? Or just use `InvalidParam`?

sbi/src/attestation.rs
```diff
@@ -181,18 +194,29 @@ pub enum AttestationFunction {
         cert_size: u64,
     },
 
-    /// Extend the TCB measurement with an additional measurement.
-    /// TBD: Do we allow for a specific PCR index to be passed, or do we extend
-    /// one dedicated PCR with all runtime extended measurements?
+    /// Extend a measurement register with an additional measurement.
+    /// The first parameter is the address of the measurement buffer.
+    /// The second argument is the length of the measurement buffer, which must
+    /// be larger than the hash algorithm size reported by `GetCapabilities`.
```
abrestic-rivos, line 200: s/"larger than"/"the same as" ?

abrestic-rivos (approved):

sbi/src/attestation.rs
sameo: Fixed, thanks.

sbi/src/error.rs
sameo: I changed the name, I'd prefer to not piggy back all errors on InvalidParam and give more details through a more specific error.

PR #103, opened 2022-09-20
PageBox: Support sequential pages as backing

glg-rv:
First PR to check early I am on the right path.

This PR is the first step to being able to put a vCPU structure (that's already bigger than a page and expected to grow rather than shrink) inside a page box. 

1. Use SequentialPages as backing rather than a single page.
2. Add test cases for single and multiple pages.

### DIFF
--- a/page-tracking/src/collections/page_box.rs
+++ b/page-tracking/src/collections/page_box.rs
@@ -6,7 +6,9 @@ use core::fmt::{self, Display};
 use core::ops::{Deref, DerefMut};
 use core::ptr::NonNull;
 use core::{borrow, cmp};
-use riscv_pages::{InternalClean, InternalDirty, Page, PageAddr, PageSize, PhysPage, RawAddr};
+use riscv_pages::{
+    InternalClean, InternalDirty, Page, PageAddr, PageSize, PhysPage, RawAddr, SequentialPages,
+};
 
 use crate::PageTracker;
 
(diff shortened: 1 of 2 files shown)


page-tracking/src/collections/page_box.rs
```diff
@@ -49,52 +51,67 @@ use crate::PageTracker;
...
-    /// Consumes the `PageBox` and returns the page that was used to hold the data.
-    pub fn to_page(mut self) -> Page<InternalDirty> {
-        let page = unsafe {
+    /// Creates a `PageBox` with a single `page` to store it.
+    pub fn new_with_single_page(
+        data: T,
+        page: Page<InternalClean>,
+        page_tracker: PageTracker,
+    ) -> Self {
+        PageBox::new_with(data, SequentialPages::from(page), page_tracker)
```
abrestic-rivos, line 87: optional: given the `From` impl for `Page` to `SequentialPages` I'm not sure we need the separate `new_with_single_page()`.

page-tracking/src/collections/page_box.rs
```diff
@@ -321,4 +373,58 @@ mod tests {
...
+                    PageOwnerId::hypervisor(),
+                    TlbVersion::new()
+                )
+                .is_ok());
+            addr = addr.checked_add_pages_with_size(1, page_size).unwrap();
+        }
+    }
+
+    #[test]
+    #[should_panic(
```
abrestic-rivos, line 413: nit: I'd just use a plan `should_panic`; testing against strings is kinda brittle and given how simple this test is I'm not sure it's worth guarding against multiple possible panic paths.

abrestic-rivos (approved):
LGTM, thanks!

page-tracking/src/collections/page_box.rs
glg-rv: My reasoning was that "PageBox" as a type does strongly suggest that there's a single page behind a structure. And since it was the only use right now in the code I kind of wanted to keep the interface for it. No strong opinion, will remove it.

page-tracking/src/collections/page_box.rs
glg-rv: Sure, makes sense.

dgreid:
LGTM, but please squash the two commits.

Thanks!

glg-rv:
Squashed commits

PR #108, opened 2022-09-26
Clean up vCPU interface separation

abrestic-rivos:
Apply the same interface separation based on state to `VmCpu` that we've done with `Vm` and `VmPages`: `ActiveVmCpu` is the interface for running vCPUs, `IdleVmCpu` is for present-but-not-running vCPUs.

Patch 1 removes the now-unused virtual register interfaces, patch 2 separates the `ActiveVmCpu` interface from `VmCpu`, and patch 3 combines `ActiveVmCpu` with `RunningVmCpu` and defines a separate `IdleVmCpu` interface.

### DIFF
--- a/sbi/src/api/tsm.rs
+++ b/sbi/src/api/tsm.rs
@@ -4,9 +4,7 @@
 
 use crate::TeeFunction::*;
 use crate::{ecall_send, Error, Result, SbiMessage};
-use crate::{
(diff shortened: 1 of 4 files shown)


dgreid (approved):
thanks, that makes things much easier to understand.

atulkharerivos (approved):

PR #109, opened 2022-09-27
Rework vCPU shared memory register set enumeration

abrestic-rivos:
Instead of having single call that writes the layout to host memory and that the host must retry if they provide insufficient buffer space, take an approach similar to the SBI PMU interface where a one call returns the number of register sets (TvmCpuNumRegisterSets) and another returns the location of a specific register set (TvmCpuGetRegisterSet).

As part of this change, RegisterSetLocation is compressed to a single u32 so that it's guaranteed to fit in a single general-purpose register. The version field is dropped and instead the versioning is implicit from the RegisterSetId.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -684,6 +684,7 @@ dependencies = [
 name = "test_workloads"
 version = "0.1.0"
 dependencies = [
+ "arrayvec",
  "attestation",
  "der",
  "device_tree",
--- a/sbi/src/api/tsm.rs
+++ b/sbi/src/api/tsm.rs
@@ -143,16 +143,23 @@ pub fn add_page_table_pages(vmid: u64, page_addr: u64, num_pages: u64) -> Result
     Ok(())
 }
 
-/// Writes the layout used for the vCPU shared-memory state area for vCPUs of `vmid` to `layout`.
-/// Returns the number of entries written to `layout`.
-pub fn get_vcpu_mem_layout(vmid: u64, layout: &mut [RegisterSetLocation]) -> Result<usize> {
-    let msg = SbiMessage::Tee(TvmCpuGetMemLayout {
+/// Returns the number of register sets in the vCPU shared-memory state area for vCPUs of `vmid`.
+pub fn num_vcpu_register_sets(vmid: u64) -> Result<u64> {
+    let msg = SbiMessage::Tee(TvmCpuNumRegisterSets { guest_id: vmid });
+    // Safety: TvmCpuNumRegisterSets does not access host memory.
+    unsafe { ecall_send(&msg) }
+}
+
+/// Returns the location of the register set at `index` in the vCPU shared-memory state area for
+/// vCPUs of `vmid`.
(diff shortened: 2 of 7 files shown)


atulkharerivos (approved):
Nice...LGTM.

atishp04:
We are back to making multiple SBI calls from the host approach :). 
The rationale behind having a number of register sets in PMU extensions was the possible counter space is large (32) and allows logical counter indices to lower privilege mode.

Here we will have a small & fixed indices (via enums) for register set. Why can't guest directly call TvmCpuGetRegisterSet for each of the register set. Anything that is not supported will be returned not supported by the hypervisor.

abrestic-rivos:
> Here we will have a small & fixed indices (via enums) for register set. Why can't guest directly call TvmCpuGetRegisterSet for each of the register set. Anything that is not supported will be returned not supported by the hypervisor.

Yeah, I went back and forth on this a bit. The rationale for the index indirection here was to force the guest to enumerate the register sets in the shared memory area so that it can properly size the area and discover any potential compatibility issues. It's the same amount of overhead (still O(n) SBI calls), but one extra interface. Seemed worth it to guard against the potential compatibility issues. WDYT?

atishp04:
The host doesn't have to pre-allocate memory before TvmCpuGrtRegisterSet. Compatibility issues are also not a big deal as you got rid of the version. But I agree this interface is bit simpler to use at the cost of one Additional ecall. So lets stick with that. It would be good to add a note with the justification so that we don't have to answer this in the future. :)

abrestic-rivos:
> Compatibility issues are also not a big deal as you got rid of the version.

That was not my intent :). Instead (in order to fit everything in a `u32`), I'm saying that `RegisterSetId` should implicitly define the version. For example, if we had multiple versions of `HypervisorCsrs` the enum would look like this:

```
enum RegisterSetId {
    HypervisorCsrsV0,
    HypervisorCsrsV1,
}
```

> So lets stick with that. It would be good to add a note with the justification so that we don't have to answer this in the future. :)

Added this to the commit message and expanded the comments a bit.

Cargo.lock
```diff
@@ -684,6 +684,7 @@ dependencies = [
 name = "test_workloads"
```
dgreid, line 684: nit: in commit message, paragraph 2: "...the host will _call_ TvmCpuGetRegisterSet..."

dgreid (approved):

atishp04:
The host 

> That was not my intent :). Instead (in order to fit everything in a `u32`), I'm saying that `RegisterSetId` should implicitly define the version. For example, if we had multiple versions of `HypervisorCsrs` the enum would look like this:

Yes. That's what I was trying to say as well. With different RegisterSetId per version, it is simpler to manage compatibilities issues.

atishp04 (approved):

Cargo.lock
abrestic-rivos: Done.

PR #113, opened 2022-09-28
Minor VM exit cleanups

abrestic-rivos:
A series of minor cleanups around the various types relating to VM exits, mostly to do with removing redundant information. Patch 1 fixes an issue where vCPUs weren't properly marked as powered-off after taking a fatal exit (oops), patches 2-8 are the aforementioned minor changes, and patch 9 makes the `TvmCpuRun` return value indicate whether or not the vCPU is still runnable, which is something that came up when talking through the spec with @atulkharerivos.


dgreid (approved):
Thanks!

PR #114, opened 2022-09-29
Move MMIO and shared memory region registration to the guest

### LINKED ISSUE #104: Move MMIO region registration to the guest
As @atishp04 and I have found, the other confidential compute implementations (pKVM, TDX, SEV) have the TVM decide what's MMIO and what isn't -- pKVM registers the regions via hypercalls, TDX & SEV have a more convoluted scheme where a page fault gets reflected back to the TVM which it can then turn into a hypercall. So, let's make `tvm_add_emulated_mmio_region()` a guest-side SBI call in a separate TEE-Guest extension to be consistent and avoid having to introduce our own KVM `ioctl()` to register MMIO registers on the host side.

We should probably also do the same with shared memory. The host can tell the TVM via device-tree or whatever where it wants the shared memory region to be, and then the guest effectively "accepts" that as the shared memory region by making a `tvm_add_shared_memory_regino()` SBI call.

In both cases we can notify the host that the TVM has registered these regions by setting up the shared-memory block to look like an ECALL exit from the TVM.

abrestic-rivos:
Per the rationale described in #104, make MMIO and shared memory region registration a guest-side function.

Patch 1 adds the new TEE-Guest SBI extension with these functions, patch 2 wires them up, patches 3 & 4 convert existing users of the host-side equivalents, and patch 5 removes those host-side functions.

Fixes #104

### DIFF
--- a/sbi/src/api/tee_guest.rs
+++ b/sbi/src/api/tee_guest.rs
@@ -1,0 +1,38 @@
+// Copyright (c) 2022 by Rivos Inc.
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
+
+use crate::TeeGuestFunction::*;
+use crate::{ecall_send, Result, SbiMessage, TeeMemoryRegion};
+
+/// Registers an emulated MMIO region in a previously-unused range of guest physical address space.
+/// Future accesses in the specified address range will trap to the host, allowing it to emulate
+/// the access.
+pub fn add_emulated_mmio_region(addr: u64, len: u64) -> Result<()> {
+    let msg = SbiMessage::TeeGuest(AddMemoryRegion {
+        region_type: TeeMemoryRegion::EmulatedMmio,
+        addr,
+        len,
+    });
+    // Safety: AddMemoryRegion does not directly access our memory. The specified range of
+    // address space must have been previously inaccessible for the call to succeed, after which
+    // accesses to that range have well-defined behavior.
+    unsafe { ecall_send(&msg) }?;
+    Ok(())
+}
+
+/// Registers a shared memory region in a previously-unused range of guest physical address space.
+/// Future accesses in the specified address range will trap to to the host, allowing it to insert
+/// pages of shared memory.
+pub fn add_shared_memory_region(addr: u64, len: u64) -> Result<()> {
+    let msg = SbiMessage::TeeGuest(AddMemoryRegion {
+        region_type: TeeMemoryRegion::Shared,
+        addr,
+        len,
+    });
+    // Safety: AddMemoryRegion does not directly access our memory. The specified range of
+    // address space must have been previously inaccessible for the call to succeed, after which
+    // accesses to that range have well-defined behavior.
+    unsafe { ecall_send(&msg) }?;
+    Ok(())
+}
--- a/sbi/src/tee_guest.rs
+++ b/sbi/src/tee_guest.rs
@@ -1,0 +1,84 @@
+// Copyright (c) 2022 by Rivos Inc.
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
+
+use crate::error::*;
+use crate::function::*;
+use crate::TeeMemoryRegion;
+
+/// Functions provided by the TEE Guest extension to TVM guests.
+#[derive(Copy, Clone, Debug)]
+pub enum TeeGuestFunction {
+    /// Adds a memory region to the calling TVM at the specified range of guest physical address
+    /// space. Both `addr` and `len` must be 4kB-aligned and must not overlap with any
+    /// previously-added regions.
+    ///
+    /// Only `Shared` and `EmulatedMmio` regions may be added by the TVM.
+    ///
+    /// a6 = 0
+    AddMemoryRegion {
+        /// a0 = type of memory region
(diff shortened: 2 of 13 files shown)


atulkharerivos (approved):
LGTM; will update documentation unless anyone else has comments.

sbi/src/api/tee_guest.rs
```diff
@@ -0,0 +1,24 @@
+// Copyright (c) 2022 by Rivos Inc.
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
+
+use crate::TeeGuestFunction::*;
+use crate::{ecall_send, Result, SbiMessage};
+
+/// Registers an emulated MMIO region. Future accesses in the specified address range will trap
+/// to the host, allowing it to emulated the access.
```
dgreid, line 9: nit: s/emulated/emulate/

sbi/src/api/tee_guest.rs
```diff
@@ -0,0 +1,24 @@
...
+pub fn add_emulated_mmio_region(addr: u64, len: u64) -> Result<()> {
+    let msg = SbiMessage::TeeGuest(AddEmulatedMmioRegion { addr, len });
+    // Safety: AddEmulatedMmioRegion does not access our memory.
```
dgreid, line 12: Is this safe because passing any address that is already accessible to the guest will be failed with an error? Worth mentioning that as it isn't obvious what the behavior is if passed an address to an existing object.

sbi/src/api/tee_guest.rs
```diff
@@ -0,0 +1,24 @@
...
+    unsafe { ecall_send(&msg) }?;
+    Ok(())
+}
+
+/// Registers a shared memory region. Future accesses in the specified address range will trap to
+/// to the host, allowing it to insert pages of shared memory.
+pub fn add_shared_memory_region(addr: u64, len: u64) -> Result<()> {
+    let msg = SbiMessage::TeeGuest(AddSharedMemoryRegion { addr, len });
+    // Safety: AddSharedMemoryRegion does not access our memory.
```
dgreid, line 21: Same comment as above w.r.t. safety reasoning.

dgreid (approved):

test-workloads/src/bin/tellus.rs
```diff
@@ -637,6 +585,8 @@ extern "C" fn kernel_init(hart_id: u64, fdt_addr: u64) {
                             zero_pages_added += 1;
                         }
                         GUEST_SHARED_PAGES_START_ADDRESS..=GUEST_SHARED_PAGES_END_ADDRESS => {
```
atishp04, line 587: As discussed the TSM will pass the memory region information from run ECALL exit and the host can cache this information. I am thinking the host should use this information to distinguish between shared page fault and zero page fault ?

sbi/src/tee_guest.rs
```diff
@@ -0,0 +1,83 @@
...
+pub enum TeeGuestFunction {
+    /// Marks the specified range of guest physical address space in the calling TVM as used for
+    /// emulated MMIO. The region is unpopulated; attempts to access this region will cause a
+    /// guest page fault exit with the faulting address and (a transformed version of) the faulting
+    /// instruction reported to the TVM's host, allowing the host to emulate the faulting memory
+    /// access. Both `addr` and `len` must be 4kB-aligned and must not overlap with any
+    /// previously-added regions.
+    ///
+    /// a6 = 0
+    AddEmulatedMmioRegion {
```
atishp04, line 19: Do we really need separate functions ? One functions with additional parameter to distinguish between MMIO/shared will be useful ? It will also align the TEEH function.

//Allowed only for CONFIDENTIAL MEMORY REGION
sbi_teeh_add_memory_region

//Allowed only for SHARED/MMIO region
sbi_teeg_add_memory_region

atulkharerivos: Actually, I missed that one...it's indeed modeled as a single call for all region types (with an enum parameter) in the documentation (for the host), and as Atish noted, it should be similar when it's moved over to the guest side.

sbi/src/api/tee_guest.rs
abrestic-rivos: Done.

sbi/src/api/tee_guest.rs
abrestic-rivos: Yes. Updated the "Safety" and function-level comment.

sbi/src/api/tee_guest.rs
abrestic-rivos: Done.

test-workloads/src/bin/tellus.rs
abrestic-rivos: This is just a test program where it's easier to hard-code everything :) But yes, I've updated this to intercept the address of the regions added by the guest and use them for address-matching.

sbi/src/tee_guest.rs
abrestic-rivos: Right, done.

test-workloads/src/bin/tellus.rs
atishp04: I was not suggesting to modify tellus :). I was just clarifying the usage from host implementation perspective.
I see you have already modified the tellus code. Sorry for the extra work ;)

atishp04 (approved):
LGTM.

atulkharerivos (commented):
LGTM.

PR #117, opened 2022-09-30
Vm: partition memory into easily hugepaged sized chunks

stillson:
This is the first change in making salus handle huge pages. It partitions memory into regions that are exact multiples of page sizes in preparation of making all the pages as large as possible.


page-tracking/src/hw_mem_map.rs
```diff
@@ -146,6 +147,96 @@ impl HwMemRegion {
         let pages = self.size / PageSize::Size4k as u64;
         self.base.checked_add_pages(pages).unwrap()
     }
+
+    /// Returns the page size if this has been partitioned, or None if it hasn't
+    pub fn page_size(&self) -> Option<PageSize> {
+        self.page_size
+    }
+
+    /// break a `HwMemRegion` into multiple `HwMemRegion`s that can all fit into vm pages of the same size
```
atulkharerivos, line 156: Nit: Capitalize first letter here and elsewhere in the file.

page-tracking/src/hw_mem_map.rs
```diff
@@ -146,6 +147,96 @@ impl HwMemRegion {
...
+    pub fn partition(&self, page_size: PageSize) -> RegionVec {
+        let mut array_vec = RegionVec::default();
+
+        // degenerate case
+        if self.size == 0 {
+            return array_vec;
+        }
+
+        // current size to small to hold a page, go to a smaller page
```
atulkharerivos, line 165: s/to/too

page-tracking/src/hw_mem_map.rs
```diff
@@ -146,6 +147,96 @@ impl HwMemRegion {
...
+        if self.size < page_size as u64 {
+            return self.partition(page_size.down_size().expect("bad page size"));
+        }
+
+        // We have at lease one page. If this is aligned we are done
```
atulkharerivos, line 170: s/lease/least

page-tracking/src/hw_mem_map.rs
```diff
@@ -146,6 +147,96 @@ impl HwMemRegion {
...
+            let mut new_reg = *self;
+            new_reg.page_size = Some(page_size);
+            array_vec.push(new_reg);
+            return array_vec;
+        }
+
+        // if this isn't aligned and couldn't contain a page size (i.e it is less than 2 pages long)
+        // down_size the page and try again
+        if page_size.num_pages(self.size) < 2 && !page_size.is_aligned(self.base.bits()) {
+            return self.partition(page_size.down_size().expect("bad page size"));
```
atulkharerivos, line 181: Not sure of whether it's right in this context, but usually we avoid panics.

page-tracking/src/hw_mem_map.rs
```diff
@@ -146,6 +147,96 @@ impl HwMemRegion {
...
+        let mut array_vec = RegionVec::default();
+
+        // degenerate case
+        if self.size == 0 {
+            return array_vec;
+        }
+
+        // current size to small to hold a page, go to a smaller page
+        if self.size < page_size as u64 {
+            return self.partition(page_size.down_size().expect("bad page size"));
```
atulkharerivos, line 167: Not sure of whether it's right in this context, but usually we avoid panics.

riscv-pages/src/page.rs
```diff
@@ -39,6 +39,16 @@ impl PageSize {
         (val + PageSize::Size4k as u64 - 1) / (PageSize::Size4k as u64)
     }
 
+    /// Return the number of pages of this `PageSize` that will fit in `val`
```
atulkharerivos, line 42: I am slightly unclear on either the description of purpose of this function. My guess is that the values for the page size match the number of bits (sans the PFN) or something along those lines.

page-tracking/src/hw_mem_map.rs
stillson: Fixed

page-tracking/src/hw_mem_map.rs
stillson: I'll switch to unwrap, which also panics but at least doesn't also give a confusing message.

atulkharerivos: Ah OK...what I meant is that we usually propagate the error upstream (by using say a Result<> type), and let the caller deal with it. It's possible that in this case, we ***are***  always expecting callers to pass us correct parameters, so unwrap() is a way to discover that something was amiss.

riscv-pages/src/page.rs
stillson: It's the same thing as  val / PageSize

Doing with shifts is faster and you don't have to think about what happens with the remainder.  It's less confusing than trying to round to the nearest page, etc. Just shift.

it could also be written

```rust
match self {
    PageSize::Size4k => val >> 12;
    PageSize::Size2M => val >>  19,
    PageSize::Size1G => val >> 30,
    PageSize::Size512G => val >> 39,
}
```

page-tracking/src/hw_mem_map.rs
stillson: I'll switch to unwrap, which also panics but at least doesn't also give a confusing message.

riscv-pages/src/page.rs
atulkharerivos: It's fine the way as written, but wasn't sure of the exact intent. I think I understand it now, but will read the code a little more.

dgreid:
Thanks for putting this up and especially for having it so well covered with tests.

Do you have some more patches that show how the split partitions will be used?

abrestic-rivos (commented):
So reading through this I'm not sure `HwMemMap` is the best place to do the page-size partitioning: it can expand the number of regions (creating failures after regions are added) and doesn't account for regions that are added late. Since we're only using this information in one spot -- when building Salus' page tables -- can we instead do the partitioning then?

Also, IMO, it would make more sense to add huge-page support to `PageTable` first so that the PR adding this page-size partitioning can also show how it'll be used.

page-tracking/src/hw_mem_map.rs
```diff
@@ -229,6 +321,27 @@ impl HwMemMapBuilder {
         Ok(self)
     }
 
+    /// Go through all of the `HwMemRegion`s, and partition them into exactly pageable sizes
+    /// After running this function all `Available` memory will be split into regions
+    /// such that each region can be split exactly into pages and the `page_size` field
+    /// of the `HwMemRegion` will be set.
+    pub fn page_size_partition(mut self) -> Self {
+        let rv: RegionVec = self
```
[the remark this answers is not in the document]
stillson: 32 works ok so far, but I will bump it to 64 to be safe

page-tracking/src/hw_mem_map.rs
```diff
@@ -146,6 +147,96 @@ impl HwMemRegion {
...
+        if self.size == 0 {
+            return array_vec;
+        }
+
+        // Current size too small to hold a page, go to a smaller page
+        if self.size < page_size as u64 {
+            return self.partition(page_size.down_size().unwrap());
+        }
+
+        // We have at least one page. If this is aligned we are done
```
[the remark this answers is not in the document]
stillson: done

page-tracking/src/hw_mem_map.rs
```diff
@@ -474,46 +591,231 @@ mod tests {
...
+                HwReservedMemType::FirmwareReserved,
+                RawAddr::supervisor(0x7_0000_0000),
+                0x1000_0000,
+            )
+            .unwrap();
+
+        builder = builder.page_size_partition();
+        let mem_map = builder.build();
+
+        let expected = vec![
```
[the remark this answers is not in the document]
stillson: Comments added

page-tracking/src/hw_mem_map.rs
```diff
@@ -229,6 +321,27 @@ impl HwMemMapBuilder {
         Ok(self)
     }
 
+    /// Go through all of the `HwMemRegion`s, and partition them into exactly pageable sizes
+    /// After running this function all `Available` memory will be split into regions
+    /// such that each region can be split exactly into pages and the `page_size` field
+    /// of the `HwMemRegion` will be set.
+    pub fn page_size_partition(mut self) -> Self {
```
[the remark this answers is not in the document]
stillson: MMIO isn't a problem, unless the addresses overlay existing ram. Hmm.

page-tracking/src/hw_mem_map.rs
```diff
@@ -146,6 +147,96 @@ impl HwMemRegion {
...
+            size: end - start,
+            page_size: Some(page_size),
+            ..*self
+        }
+        .partition(page_size)
+        .into_iter()
+        .for_each(|m| array_vec.push(m));
+
+        // above
+        let above_len = (end - self.base.bits()) / 4096;
```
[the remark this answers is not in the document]
stillson: fixed

riscv-page-tables/src/page_table.rs
```diff
@@ -473,6 +474,7 @@ pub trait GuestStagePagingMode: PagingMode<MappedAddressSpace = GuestPhys> {
 }
 
 /// The internal state of a paging hierarchy.
+#[derive(Debug)]
```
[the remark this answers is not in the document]
stillson: I'll keep that in mind for the future

page-tracking/src/hw_mem_map.rs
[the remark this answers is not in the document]
abrestic-rivos: It won't overlap existing DRAM, but why wouldn't it be a problem? Presumably we'd want to map the 2GB+ of PCIe BAR space with hugepages?

page-tracking/src/hw_mem_map.rs
```diff
@@ -146,6 +147,96 @@ impl HwMemRegion {
...
+            let mut new_reg = *self;
+            new_reg.page_size = Some(page_size);
+            array_vec.push(new_reg);
+            return array_vec;
+        }
+
+        // If this isn't aligned and couldn't contain a page size (i.e it is less than 2 pages long)
+        // down_size the page and try again
+        if page_size.num_pages(self.size) < 2 && !page_size.is_aligned(self.base.bits()) {
+            return self.partition(page_size.down_size().unwrap());
```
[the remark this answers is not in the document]
stillson: fixed

atulkharerivos:
Believe the branch needs to be rebased before pushing. At present, it shows several unrelated commits in the pull request.

stillson:
Sorry, trying to rebase to keep the code up to date, forgot about the existing pr

Issue #112, opened 2022-09-28
Align Salus TEE FIDs with proposed SBI specification...


atulkharerivos:
The Salus TEE FIDs are currently unaligned with the proposed SBI specification, and should be reconciled with it.

PR #118, opened 2022-09-30
Renumber TEE Host FIDs to match SBI draft

atulkharerivos:
This aligns the FIDs with the draft in [1], and has no other functional changes.

[1]:
https://github.com/riscv-non-isa/riscv-ap-tee/pull/5/commits

### DIFF
--- a/sbi/src/tee_host.rs
+++ b/sbi/src/tee_host.rs
@@ -256,28 +256,102 @@ impl TeeMemoryRegion {
 /// Functions provided by the TEE Host extension.
 #[derive(Copy, Clone, Debug)]
 pub enum TeeHostFunction {
+    /// Writes up to `len` bytes of the `TsmInfo` structure to the non-confidential physical address
+    /// `dest_addr`. Returns the number of bytes written.
+    ///
+    /// a6 = 0
+    TsmGetInfo {
+        /// a0 = destination address of the `TsmInfo` structure
+        dest_addr: u64,
+        /// a1 = maximum number of bytes to be written
+        len: u64,
+    },
+    /// Converts `num_pages` of non-confidential memory starting at `page_addr`. The converted pages
+    /// remain non-confidential, and thus may not be assigned for use by a child TVM, until the
+    /// fence procedure, described below, has been completed.
+    ///
+    /// a6 = 1
+    TsmConvertPages {
+        /// a0 = base address of pages to convert
+        page_addr: u64,
+        /// a1 = page size
+        page_type: TsmPageType,
+        /// a2 = number of pages
+        num_pages: u64,
+    },
+    /// Reclaims `num_pages` of confidential memory starting at `page_addr`. The pages must not
+    /// be currently assigned to an active TVM.
+    ///
+    /// a6 = 2
+    TsmReclaimPages {
+        /// a0 = base address of pages to reclaim
+        page_addr: u64,
+        /// a1 = page size
+        page_type: TsmPageType,
+        /// a2 = number of pages
+        num_pages: u64,
+    },
+    /// Initiates a TLB invalidation sequence for all pages marked for conversion via calls to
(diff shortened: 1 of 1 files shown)


atulkharerivos:
Related to https://github.com/rivosinc/salus/issues/112.

abrestic-rivos:
Maybe I missed this discussion, but IMO it makes more sense to group functions by related operation/procedure than putting them in lifecycle order. For example, I would think the convert/reclaim/fence functions would belong together, as would the add measured/zero/shared pages.

atulkharerivos:
> Maybe I missed this discussion, but IMO it makes more sense to group functions by related operation/procedure than putting them in lifecycle order. For example, I would think the convert/reclaim/fence functions would belong together, as would the add measured/zero/shared pages.

I don't have strong feelings about arranging by lifetime vs. function. In the SBI document, it lifecycle looked a little more natural (just to avoid forward references), but since it's searchable, don't see that as sticking point. Will change once others chime in.

atishp04:
Yeah. That was the intention. I just suggested life cycle order for better reading experience. We can certainly group function based on operations as well. How about this order ? Don't feel very strongly about this but it will probably avoid bunch of forward references which was the original intention.

0. sbi_tee_host_get_tsm_info
1. sbi_tee_host_convert_pages
2. sbi_tee_host_reclaim_pages
3. sbi_tee_host_global_fence
4. sbi_tee_host_local_fence
5. sbi_tee_host_create_tvm
6. sbi_tee_host_finalize_tvm
7. sbi_tee_host_destroy_tvm
8. sbi_tee_host_add_tvm_memory_region
9. sbi_tee_host_add_tvm_page_table_pages
10. sbi_tee_host_add_tvm_measured_pages
11. sbi_tee_host_add_tvm_zero_pages
12. sbi_tee_host_add_tvm_shared_pages
13. sbi_tee_host_get_tvm_vcpu_num_register_sets
14. sbi_tee_host_get_tvm_vcpu_register_set
15. sbi_tee_host_create_tvm_vcpu
16. sbi_tee_host_run_tvm_vcpu

abrestic-rivos:
> Yeah. That was the intention. I just suggested life cycle order for better reading experience. We can certainly group function based on operations as well. How about this order ? Don't feel very strongly about this but it will probably avoid bunch of forward references which was the original intention.
> 
> 0. sbi_tee_host_get_tsm_info
> 1. sbi_tee_host_convert_pages
> 2. sbi_tee_host_reclaim_pages
> 3. sbi_tee_host_global_fence
> 4. sbi_tee_host_local_fence
> 5. sbi_tee_host_create_tvm
> 6. sbi_tee_host_finalize_tvm
> 7. sbi_tee_host_destroy_tvm
> 8. sbi_tee_host_add_tvm_memory_region
> 9. sbi_tee_host_add_tvm_page_table_pages
> 10. sbi_tee_host_add_tvm_measured_pages
> 11. sbi_tee_host_add_tvm_zero_pages
> 12. sbi_tee_host_add_tvm_shared_pages
> 13. sbi_tee_host_get_tvm_vcpu_num_register_sets
> 14. sbi_tee_host_get_tvm_vcpu_register_set
> 15. sbi_tee_host_create_tvm_vcpu
> 16. sbi_tee_host_run_tvm_vcpu

LGTM

atulkharerivos:
Updated per the above.

sbi/src/tee_host.rs
```diff
@@ -256,28 +256,103 @@ impl TeeMemoryRegion {
...
         len: u64,
     },
+    /// Moves a VM from the initializing state to the Runnable state
+    ///
+    /// a6 = 6
+    Finalize {
+        /// a0 = guest id
+        guest_id: u64,
+    },
+
```
abrestic-rivos, line 326: nit: spurious newline?

abrestic-rivos (approved):
LGTM, thanks!

PR #120, opened 2022-10-03
Allocate vCPU state on vCPU creation.

glg-rv:
This PR adds support for vCPUs state to be donated from the host at vCPU creation, not at TVM creation.

Lots of things going on here, in broad strokes:

1. TvmCpuCreate and TvmCreate interface are modified. TvmCreate doesn't need to know the number of vCPUs nor pass a global vcpu state address for all the future vcpus.
2. VmCpus is modified to hold a static array of MAX_VCPUS of what are essentially pointers to future vCPUs.
3. Given the previous point, the tvm state is now bigger than a page, so PageArc has to be modified to support SequentialPages as backing. This work is in line with what has been done to PageBox to hold the vCPU state.
4. A chunk allocator is added to Sequential Pages to allow to iterate on chunks of fixed number of pages at the time.

### DIFF
--- a/page-tracking/src/collections/page_arc.rs
+++ b/page-tracking/src/collections/page_arc.rs
@@ -8,7 +8,7 @@ use core::ops::Deref;
 use core::ptr::NonNull;
 use core::sync::atomic::{AtomicUsize, Ordering};
 
-use riscv_pages::{InternalClean, InternalDirty, Page, PageAddr, PageSize, PhysPage, RawAddr};
+use riscv_pages::{InternalClean, InternalDirty, PageAddr, PageSize, RawAddr, SequentialPages};
 
 use crate::collections::PageBox;
 use crate::PageTracker;
@@ -18,6 +18,7 @@ struct PageArcInner<T> {
     data: T,
     rc: AtomicUsize,
     page_size: PageSize,
+    page_count: u64,
     page_tracker: PageTracker,
 }
 
@@ -32,20 +33,41 @@ struct PageArcInner<T> {
 pub struct PageArc<T>(NonNull<PageArcInner<T>>);
 
 impl<T> PageArc<T> {
+    /// Try to create a `PageArc` that wraps the given data using
+    /// `pages` to store it, returning it to its previous owner on
+    /// `drop()` using `page_tracker`. Returns `None` if not enough
+    /// pages are supplied.
+    pub fn try_new(
+        data: T,
+        pages: SequentialPages<InternalClean>,
+        page_tracker: PageTracker,
+    ) -> Option<Self> {
+        if core::mem::size_of::<PageArcInner<T>>() <= pages.length_bytes() as usize {
+            Some(Self::new_with(data, pages, page_tracker))
+        } else {
+            None
+        }
+    }
+
     /// Creates a `PageArc` that wraps the given data using `page` as the backing store. The page is
     /// released back to the previous owner when the last reference is dropped using `page_tracker`.
-    pub fn new_with(data: T, page: Page<InternalClean>, page_tracker: PageTracker) -> Self {
-        let ptr = page.addr().bits() as *mut PageArcInner<T>;
+    pub fn new_with(
+        data: T,
+        pages: SequentialPages<InternalClean>,
+        page_tracker: PageTracker,
+    ) -> Self {
+        let ptr = pages.base().bits() as *mut PageArcInner<T>;
         assert!(!ptr.is_null()); // Explicitly ban pages at zero address.
         let inner = PageArcInner {
             data,
             rc: AtomicUsize::new(1),
-            page_size: page.size(),
+            page_size: pages.page_size(),
+            page_count: pages.len(),
             page_tracker,
         };
-        assert!(core::mem::size_of::<PageArcInner<T>>() <= page.size() as usize);
         unsafe {
             // Safe as the memory is totally owned and PageArcInner<T> fits in the page.
+            assert!(core::mem::size_of::<PageArcInner<T>>() <= pages.length_bytes() as usize);
             core::ptr::write(ptr, inner);
         }
         Self(NonNull::new(ptr).unwrap())
@@ -67,11 +89,18 @@ impl<T> PageArc<T> {
         this.inner().rc.load(Ordering::Acquire);
 
         let page_size = this.inner().page_size;
+        let page_count = this.inner().page_count;
         let page_tracker = this.inner().page_tracker.clone();
         // Safety: PageArcInner is repr(C) with data as the first field, therefore a pointer to data
         // is a page-aligned and properly initialized pointer to T.
-        let boxed =
-            unsafe { PageBox::from_raw(Self::as_ptr(&this) as *mut T, page_size, 1, page_tracker) };
+        let boxed = unsafe {
+            PageBox::from_raw(
+                Self::as_ptr(&this) as *mut T,
+                page_size,
+                page_count,
+                page_tracker,
+            )
+        };
         core::mem::forget(this);
         Ok(boxed)
     }
@@ -91,6 +120,23 @@ impl<T> PageArc<T> {
         // Safe since we were initialized to point to a valid PageArcInner in the constructor and
         // if we still hold a reference the structure we point to must still be alive.
         unsafe { self.0.as_ref() }
+    }
+
+    /// Returns the `SequentialPages` backing this `PageArc`.
+    ///
+    /// # Safety
+    ///
+    /// This method semantically takes ownership of the backing storage of this container without
+    /// preventing further usage. It is the caller's responsibility to ensure that this `PageArc`
+    /// is never used again.
+    unsafe fn take_pages(&mut self) -> SequentialPages<InternalDirty> {
+        // Unwrap ok, the backing pages must've been contiguous.
+        SequentialPages::from_mem_range(
+            PageAddr::new(RawAddr::supervisor(self.0.as_ptr() as u64)).unwrap(),
+            self.inner().page_size,
+            self.inner().page_count,
+        )
+        .unwrap()
     }
 }
 
@@ -124,17 +170,13 @@ impl<T> Drop for PageArc<T> {
         }
 
         let page_tracker = self.inner().page_tracker.clone();
-        let page_size = self.inner().page_size;
-        // Safe because we now have unique ownership of the page this PageArc was constructed with.
-        let page: Page<InternalDirty> = unsafe {
-            Page::new_with_size(
-                PageAddr::new(RawAddr::supervisor(self.0.as_ptr() as u64)).unwrap(),
-                page_size,
-            )
-        };
+        // Safe because we're in drop() and this PageArc can no longer be used.
+        let pages = unsafe { self.take_pages() };
 
-        // Unwrap ok: we have unique ownership of the page so we must be able to release it.
-        page_tracker.release_page(page).unwrap();
+        for p in pages {
+            // Unwrap ok: we have unique ownership of the page so we must be able to release it.
+            page_tracker.release_page(p).unwrap();
+        }
     }
 }
 
@@ -181,27 +223,24 @@ mod tests {
     use crate::TlbVersion;
     use riscv_pages::*;
 
-    fn stub_page() -> (PageTracker, Page<InternalClean>) {
+    fn stub_pages(n: usize) -> (PageTracker, SequentialPages<InternalClean>) {
         let (page_tracker, mut pages) = PageTracker::new_in_test();
-        let assigned_page = pages
-            .by_ref()
-            .take(1)
-            .map(|p| {
-                page_tracker
-                    .assign_page_for_internal_state(p, PageOwnerId::host())
-                    .unwrap()
-            })
-            .next()
-            .unwrap();
-        (page_tracker, assigned_page)
+        let assigned_pages = SequentialPages::from_pages(pages.by_ref().take(n).map(|p| {
+            page_tracker
+                .assign_page_for_internal_state(p, PageOwnerId::host())
+                .unwrap()
+        }))
+        .unwrap();
+        (page_tracker, assigned_pages)
     }
 
     #[test]
     fn lifecycle() {
-        let (page_tracker, backing_page) = stub_page();
-        let addr = backing_page.addr();
+        const TEST_PAGE_COUNT: usize = 1;
+        let (page_tracker, backing_pages) = stub_pages(TEST_PAGE_COUNT);
+        let addr = backing_pages.base();
         {
-            let arc = PageArc::new_with([5u8; 128], backing_page, page_tracker.clone());
+            let arc = PageArc::new_with([5u8; 128], backing_pages, page_tracker.clone());
             assert_eq!(PageArc::ref_count(&arc), 1);
             {
                 let copy = arc.clone();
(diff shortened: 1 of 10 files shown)


page-tracking/src/collections/page_arc.rs
```diff
@@ -32,20 +33,41 @@ struct PageArcInner<T> {
...
+        } else {
+            None
+        }
+    }
+
     /// Creates a `PageArc` that wraps the given data using `page` as the backing store. The page is
     /// released back to the previous owner when the last reference is dropped using `page_tracker`.
-    pub fn new_with(data: T, page: Page<InternalClean>, page_tracker: PageTracker) -> Self {
-        let ptr = page.addr().bits() as *mut PageArcInner<T>;
+    pub fn new_with(
```
dgreid, line 54: try_new seems like a better interface, and combining them will be allow the assert to be removed. Is there a reason to keep the asserting form `new_with`?

page-tracking/src/collections/page_box.rs
```diff
@@ -56,8 +56,22 @@ pub struct PageBox<T> {
 }
 
 impl<T> PageBox<T> {
-    /// Creates a `PageBox` that wraps the given data using `pages` to store it, returning it to its
-    /// previous owner on `drop()` using `page_tracker`.
+    /// Try to create a `PageBox` that wraps the given data using `pages` to store it, returning it to its
+    /// previous owner on `drop()` using `page_tracker`. Returns `None` if not enough pages are supplied.
+    pub fn try_new(
```
dgreid, line 61: Same comment as `PageArc`, can the panicy new_with be dropped?

riscv-pages/src/sequential_pages.rs
```diff
@@ -160,10 +160,38 @@ impl<S: State> SequentialPages<S> {
...
+        let addr = self.addr;
+        let page_size = self.page_size;
+        let count = if self.count > n { n } else { self.count };
+        self.addr = self
+            .addr
+            .checked_add_pages_with_size(count, page_size)
+            .unwrap();
+        self.count -= count;
+        // Safe because `pages` owns the memory, and we're adding to
+        // this sequence pages that have been removed from `self`.
```
dgreid, line 180: safety comment seems off. there isn't a `pages` in scope here.

riscv-pages/src/sequential_pages.rs
```diff
@@ -236,6 +264,43 @@ impl<S: State, I: Iterator<Item = Page<S>>> fmt::Debug for Error<S, I> {
...
 
+pub struct SeqPageChunkIter<S: State> {
+    pages: SequentialPages<S>,
+    n: u64,
+}
+
+impl<S: State> Iterator for SeqPageChunkIter<S> {
+    type Item = SequentialPages<S>;
+
+    fn next(&mut self) -> Option<SequentialPages<S>> {
```
dgreid, line 275: This code looks very similar to `take` above, can the iterator re-use that code?

sbi/src/api/tee_host.rs
```diff
@@ -73,24 +73,10 @@ pub fn reclaim_pages(addr: u64, num_pages: u64) -> Result<()> {
 ///
```
dgreid, line 73: stray (squash artifact?) comments in the commit message.

sbi/src/api/tee_host.rs
```diff
@@ -169,10 +155,16 @@ pub fn get_vcpu_register_set(vmid: u64, index: u64) -> Result<RegisterSetLocatio
 /// sufficient to hold the structure layout returned from `get_vcpu_mem_layout()`. Since memory
 /// within the shared-memory state area may be read or written by the TSM at any time, the caller
 /// must treat the memory as volatile for the lifetime of the TVM.
```
dgreid, line 157: safety comment needs to be updated to cover safe use of `state_addr`

src/vm_cpu.rs
```diff
@@ -1102,63 +1098,69 @@ pub enum VmCpuStatus {
...
+    pub fn add_vcpu(&self, vcpu_id: u64, vcpu_box: PageBox<Mutex<VmCpu>>) -> Result<()> {
+        let entry = self.vcpus.get(vcpu_id as usize).ok_or(Error::BadCpuId)?;
         let mut status = entry.status.write();
         if *status != VmCpuStatus::NotPresent {
             return Err(Error::VmCpuExists);
         }
-        entry.vcpu.lock().shared_area.call_once(|| shared_area);
         *status = VmCpuStatus::PoweredOff;
+
+        assert!(!entry.vcpu.is_completed());
```
dgreid, line 1144: why assert here instead of returning an error?

src/vm_pages.rs
```diff
@@ -1131,38 +1132,57 @@ impl<'a, T: GuestStagePagingMode> FinalizedVmPages<'a, T> {
...
+        let converted_pages = self.get_converted_pages(state_addr, VM_CPU_STATE_PAGES)?;
+        if !converted_pages.is_contiguous() {
+            return Err(Error::NonContiguousPages);
+        }
+
+        let new_owner = to.page_owner_id();
         let vcpu_pages =
-            SequentialPages::from_pages(self.assign_state_pages_for(vcpu_pages, id)).unwrap();
+            SequentialPages::from_pages(self.assign_state_pages_for(converted_pages, new_owner))
+                .unwrap();
```
dgreid, line 1178: Is this unwrap guaranteed never to panic? If so, why? Can it instead return an error?

riscv-pages/src/sequential_pages.rs
glg-rv: Agreed. I was in fact meant to do that but clearly forgot.

src/vm_cpu.rs
glg-rv: Because this is an invariant failure. The VmCpuExists error above should cover the case where a vCPU has already been added. This is paranoia added because call_once will silently not do anything in case something is there already.

If this assert is triggered, we would have a vCPU that was marked as non present but somehow had some vCPU state already allocated. This shouldn't happen, and it would be disastrous if by mistake it would. Safe place to have an assert. :)

src/vm_cpu.rs
```diff
@@ -1102,63 +1098,69 @@ pub enum VmCpuStatus {
...
+            num: RwLock::new(0),
+            vcpus,
         }
-        Ok(Self { inner })
     }
 
     /// Returns the number of vCPUs in this `VmCpus`.
     pub fn num_vcpus(&self) -> usize {
-        self.inner.len()
+        *self.num.read()
```
dgreid, line 1131: Does this re-define `num_vcpus` from number of possible to number added? Does that break `validate_imsic_addrs`?

src/vm_cpu.rs
dgreid: Can that invariant be represented with the type system? A type that indicates existence and holds the object instead of an object and a separate flag?

src/vm_cpu.rs
```diff
@@ -1102,63 +1098,69 @@ pub enum VmCpuStatus {
...
             return Err(Error::VmCpuExists);
         }
-        entry.vcpu.lock().shared_area.call_once(|| shared_area);
         *status = VmCpuStatus::PoweredOff;
+
+        assert!(!entry.vcpu.is_completed());
+        entry.vcpu.call_once(|| vcpu_box);
+
+        {
+            let mut n = self.num.write();
```
dgreid, line 1148: Same type system comment about the length and the storage of VCPUs being held separately. They should be bundled like Vec.

sbi/src/api/tee_host.rs
glg-rv: Yes.

src/vm_cpu.rs
glg-rv: I thought about it. Essentially what we'd need is a guaranteed fixed-space vector (where `with_capacity` does in fact guarantee to allocate precisely that).
I didn't want to use third party crates for something like this, given the nature of the project.
The idea of adding yet another type, and the question of where to add it -- I think `vm_cpu.rs` is already too long than it should be -- pushed me to attempt a C-style approach, and make the type that .guarantees this `VmCpus` itself.

I am open to suggestion on how to do this more in line with where this projects needs to be.

src/vm_pages.rs
glg-rv: I was going with the flow of the rest of Salus code.

Not a complete expert on the project yet, but since the pages comes from converted_pages _should_ not.

I think this has the same possibility of being wrong as the rest of the salus code.

The question of whether carefully selected page list passed by the TEE host might or might not crash salus -- we should make sure that it doesn't -- is definitely something that caught my eye.
I'd like to volunteer for that, but I think this is beyond the scope of this change.

riscv-pages/src/sequential_pages.rs
```diff
@@ -236,6 +264,43 @@ impl<S: State, I: Iterator<Item = Page<S>>> fmt::Debug for Error<S, I> {
...
+            .unwrap();
+        self.pages.count -= count;
+        // Safe because `pages` owns the memory, and we're adding to this sequence pages that have been removed.
+        unsafe { Some(SequentialPages::from_mem_range(addr, page_size, count).unwrap()) }
+    }
+
+    fn size_hint(&self) -> (usize, Option<usize>) {
+        let count = self.pages.count as usize;
+        (count, Some(count))
+    }
```
abrestic-rivos, line 299: Shouldn't `size_hint()` return the number of chunks remaining in the iterator, rather than the number of pages?

src/vm_cpu.rs
```diff
@@ -578,7 +574,7 @@ impl VmCpu {
     // Returns a reference to the shared-memory state area for this vCPU.
     fn shared_area(&self) -> &VmCpuSharedArea {
         // Unwrap ok: shared_area must've been initialized for this vCPU to have been activated.
```
abrestic-rivos, line 576: Unwrap comment can be dropped now.

src/vm_pages.rs
abrestic-rivos: It is guaranteed to succeed because of the `is_contiguous()` check above. The reason the check is separate from the creation of the `SequentialPages` is to avoid having to unwind the work done by `assign_state_pages_for()`. This deserves a comment, which I should've added when I initially introduced this pattern in `create_guest_vm()` (whoops)

src/vm_pages.rs
```diff
@@ -1131,38 +1132,57 @@ impl<'a, T: GuestStagePagingMode> FinalizedVmPages<'a, T> {
...
                 .unwrap();
 
-        let mut state_pages = self.assign_state_pages_for(state_pages, id);
-        let box_page = state_pages.next().unwrap();
-        let region_vec_pages = SequentialPages::from_pages(state_pages).unwrap();
+        let mut state_pages =
+            SequentialPages::from_pages(self.assign_state_pages_for(state_pages, id)).unwrap();
+        let vm_state_pages = state_pages
+            .take(TVM_VM_STATE_PAGES)
+            .ok_or(Error::NotEnoughMemory)?;
```
abrestic-rivos, line 1150: This leaks `state_pages` on error. We should check that we have enough pages up front.

src/vm_pages.rs
```diff
@@ -68,9 +69,11 @@ pub type InstructionFetchResult = core::result::Result<DecodedInstruction, Instr
 /// The number of pages for the `VmRegionList` vector.
 pub const TVM_REGION_LIST_PAGES: u64 = 1;
 
-/// The base number of state pages required to be donated for creating a new VM. For now, we just need
-/// one page to hold the VM state itself and whatever is required to hold the `VmRegionList`.
-pub const TVM_STATE_PAGES: u64 = 1 + TVM_REGION_LIST_PAGES;
+/// The number of pages to hold `Vm` state.
+pub const TVM_VM_STATE_PAGES: u64 = 3;
```
abrestic-rivos, line 73: Could we avoid hard-coding this constant with something like this?

```
pub const TVM_VM_STATE_PAGES: u64 = PageSize::num_4k_pages(core::mem::size_of::<Vm>());
```

src/vm_cpu.rs
abrestic-rivos: Something like this?

```
enum VmCpuState {
    NotPresent,
    PoweredOff(PageBox<Mutex<VmCpu>>),
    Runnable(PageBox<Mutex<VmCpu>>),
    Running(PageBox<Mutex<VmCpu>>),
}

struct VmCpuInner {
    vcpu: RwLock<VmCpuState>,
}
```

I think I tried this before the separate struct + flag approach that we have now and it was a little ugly. Maybe worth revisiting?

glg-rv: Oh yes, I think I was missing the point by mixing up with the num issue below.

riscv-pages/src/sequential_pages.rs
glg-rv: Yes.

src/vm_cpu.rs
glg-rv: Oh yes. In my defense a name such as `max_vcpus()` would be more appropriate. This returns, in fact, the number of CPUs. Permission to change it to `max_vcpus()` requested.

dgreid: sgtm, thanks!

However, if you modify the storage and allow looping over only present vCPUs, it might not be needed.

src/vm.rs
```diff
@@ -157,6 +162,7 @@ impl From<VmPagesError> for EcallError {
     fn from(error: VmPagesError) -> EcallError {
         match error {
             VmPagesError::PageFault(pf, e, addr) => EcallError::PageFault(pf, e, addr),
+            VmPagesError::NotEnoughMemory => EcallError::Sbi(SbiError::InvalidParam),
```
atishp04, line 165: This should also follow the same TODO about error code mapping form GuestTrackingError.

src/vm.rs
```diff
@@ -396,10 +410,10 @@ pub type InitializingVm<'a, T> = VmRef<'a, T, VmStateInitializing>;
 
 impl<'a, T: GuestStagePagingMode> InitializingVm<'a, T> {
     /// Adds a vCPU to this VM.
-    fn add_vcpu(&self, vcpu_id: u64, shared_area: VmCpuSharedArea) -> EcallResult<()> {
+    fn add_vcpu(&self, vcpu_id: u64, vcpu_box: PageBox<Mutex<VmCpu>>) -> EcallResult<()> {
```
atishp04, line 413: Need some rust lessons here :). What does PageBox represent ?

dgreid: ah, not a general rust thing. PageBox is a salus-ism and I'm to blame. It's a smart pointer like the standard rust `Box` but backed by a page or pages instead of a general purpose allocation.

atishp04: Thanks. I am reading the comments now.
https://github.com/rivosinc/salus/blob/main/page-tracking/src/collections/page_box.rs

src/vm_pages.rs
```diff
@@ -1131,38 +1132,57 @@ impl<'a, T: GuestStagePagingMode> FinalizedVmPages<'a, T> {
...
-            )
-            .map_err(Error::VmCreationFailed)?,
-            box_page,
-        ))
+        PageBox::try_new(
+            Mutex::new(VmCpu::new(vcpu_id, new_owner, shared_area)),
+            vcpu_pages,
+            self.inner.page_tracker.clone(),
+        )
+        .ok_or(Error::NotEnoughMemory)
```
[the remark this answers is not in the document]
glg-rv: Oh this is a very good point. I think then we should either keep the new_with() or return on error the sequential pages, so the caller can do the releasing.

I am more inclined on getting rid of the try_new(), as errors are easy to mishandle. And keep the caller to ensure that there are enough pages.

src/vm_pages.rs
glg-rv: Same as above.

src/vm_pages.rs
glg-rv: Yes.

glg-rv:
Closing this pull request as my approach on VmCpus handling has changed substantially from this PR. Sending a new out out soon.

PR #122, opened 2022-10-04
Add support for binding/unbding vCPUs to interrupt files

abrestic-rivos:
Define the necessary TEE-Host and TEE-Interrupt APIs for binding and unbinding a vCPU to guest interrupt file(s). The actual mechanics of saving/restoring state from an interrupt file are described in more detail in the AIA specification, but roughly the process looks like this:

Bind:
- Map in IMSIC interrupt files in the VM's page tables.
- Restore IMSIC state from software interrupt file.

Unbind:
- Invalidate IMSIC interrupt files in the VM's page tables.
- Save IMSIC state to software interrupt file.

`tvm_cpu_bind_imsic()` implements the former, and `tvm_cpu_unbind_imsic_{begin,end}()` implement the latter. As the unbind process requires a TLB flush, it is broken into `begin()` and `end()` calls between which the host must complete a TLB fence procedure for the TVM. The fence is initiated by a newly-introduced `tvm_initiate_fence()` call which will bump the TLB version for the target TVM much like `tsm_initiate_fence()` does for the calling VM. Instead of requiring the host to enter the TSM on every CPU however, it can complete the TVM fence by causing on exit on each vCPU that was running prior to the call to `tvm_initiate_fence()`.

Some other notes on the API / planned implementation:
 - No MRIFs for now since we don't have emulation support for those yet. We'll just save state to an MRIF-like software interrupt file.
 - No IMSIC emulation by the TSM. vCPUs must be bound to an interrupt file in order to run, and IPIs targeting an unbound vCPU will cause a guest page fault.
 - We protect for nested IMSIC virtualization by specifying an interrupt file mask in `tvm_cpu_bind_imsic()`. 
 - We'll eventually need a way for the host to inject interrupts (e.g. for virtio). That'll be through a separate API, and the TSM can choose the proper way to inject the interrupt depending on whether or not the vCPU is bound.

### DIFF
--- a/riscv-page-tables/src/page_table.rs
+++ b/riscv-page-tables/src/page_table.rs
@@ -31,8 +31,10 @@ pub enum Error {
     PageNotMapped,
     /// The requested range couldn't be removed from the page table.
     PageNotUnmappable,
-    /// Attempt to access a non-converted page as confidential.
+    /// The requested range isn't converted.
     PageNotConverted,
+    /// The requested range couldn't be unassigned.
+    PageNotUnassignable,
     /// Attempt to lock a PTE that is already locked.
     PteLocked,
     /// Attempt to unlock a PTE that is not locked.
@@ -176,6 +178,11 @@ impl<'a, T: PagingMode> InvalidatedPte<'a, T> {
     fn lock(self) -> LockedPte<'a, T> {
         self.pte.lock();
         LockedPte::new(self.pte, self.level)
+    }
+
+    /// Clears the PTE.
+    fn clear(self) {
+        self.pte.clear();
     }
 }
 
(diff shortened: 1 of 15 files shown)


sbi/src/tee_interrupt.rs
```diff
@@ -97,6 +97,55 @@ pub enum TeeInterruptFunction {
...
+    },
+    /// Begins the unbind process for the specified vCPU from its guest interrupt files. The
+    /// translations for the vCPU's virtualized IMSIC are invalidated, and a TLB flush sequence
+    /// for the TVM must be completed before calling `TvmCpuUnbindImsicEnd` to complete the unbind
+    /// process. Must be called on the physical CPU to which the vCPU is bound.
+    ///
+    /// Returns 0 on success.
+    ///
+    /// a6 = 5
+    TvmCpuUnbindImsicBegin {
```
dgreid, line 127: Does the unbind sequence look like this?

UnbindBegin
  TvmInitiateFence
    do needed IPIs
UnbindEnd

Did you consider automatically triggering the fence on UnbindBegin?

abrestic-rivos: > Does the unbind sequence look like this?
>
> UnbindBegin
TvmInitiateFence
do needed IPIs
UnbindEnd

Yes, that's right.

> Did you consider automatically triggering the fence on UnbindBegin?

I did. Two reasons I didn't go the implicit fence route:
1) It prevents the host from batching multiple invalidations in a single fence sequence
2) Since the fence sequence must be executed atomically, the host would need to hold a TVM-global lock around `unbind_begin()` + IPIs.

I was more concerned about (2) since `unbind_begin()` might be expensive. It has to: prepare the MRIF, unmap the IMSIC page, and issue IOMMU fence commands. 

Thoughts?

atulkharerivos: Off the top of my head, the batch interface definitely seems like the optimal approach. Most IOMMUs that I know support batched invalidations for similar reasons.

abrestic-rivos:
Updated the PR with the implementation of vCPU bind/unbind, PTAL. This should theoretically be enough to boot an SMP TVM, though obviously we're not at the point where we can test that yet :)

dgreid (approved):
thanks, looks great.

atulkharerivos (approved):

riscv-page-tables/src/page_table.rs
```diff
@@ -913,6 +920,46 @@ impl<T: PagingMode> GuestStagePageTable<T> {
         Ok(pages)
     }
 
+    /// Unassigns pages that were previously mapped in this page table if they were invalidated a
```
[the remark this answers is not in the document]
abrestic-rivos: Done.

PR #129, opened 2022-10-07
Add Salus-TSM stub

atulkharerivos:
This adds a new binary target for the Salus-TSM. The Salus-TSM binary will be loaded early by the platform-FW, and is intended for use with a HS-mode VMM.

At present, the binary is a stub intended for use prototyping purposes, and will evolve to provide the complete range of functionality.

1.  The first commit adds a new crate to contain functionality that will be shared between Salus binaries
2. The second commit adds the Salus-TSM binary
3. The third commit removes shared functionality from the Salus source.


dgreid:
Looks like a squashing error on commits 3 and 4?

Also can we have a slack discussion about splitting in to two vs adding an optional mode to the existing binary?

Cargo.toml
```diff
@@ -13,6 +13,15 @@ panic = "abort"
...
 salustsm = []
 
+[[bin]]
+name = "salus-tsm"
+path = "src/salus_tsm.rs"
+required-features = ["salustsm"]
+
+[[bin]]
+name = "salus"
+path = "src/main.rs"
```
abrestic-rivos, line 23: Instead of `main.rs` and `salus_tsm.rs`, can we put `main` in both names so that it's obvious that they're both the "main" modules of their respective binaries?

Makefile
```diff
@@ -141,6 +146,12 @@ run_debian: salus
 		-device e1000e,netdev=usernet \
 		$(EXTRA_QEMU_ARGS)
 
+run_salus_tsm: salus_tsm
+	$(QEMU_BIN) \
+		$(MACH_ARGS) \
+		-kernel $(RELEASE_BINS)salus-tsm \
+		$(EXTRA_QEMU_ARGS)
```
abrestic-rivos, line 153: We should probably have an internal discussion about how this Makefile target is expected to be used.

Cargo.toml
```diff
@@ -10,6 +10,9 @@ strip = "debuginfo"
 codegen-units = 1
 panic = "abort"
 
+[features]
+salustsm = []
```
abrestic-rivos, line 14: If there are separate binary targets, do we still need a feature flag? If we want a feature flag + separate binary targets, is there a way to automatically set the feature based on which binary we're building?

Also some bikeshedding: I feel like "tsm_only" is a better name than "salustsm", as it's more obvious what the flag is doing (stripping out the non-TSM components).

abrestic-rivos (commented):
Hm, I'm not sure separate binary targets + feature flag is the way to go, unless the binary targets can implicitly select the correct value of the feature flag.

I'm not a huge fan of the "Salus" vs "Salus-TSM" naming as they're both TSMs, but the former has the ability to load & run the host in VS while the latter is just a TSM that expects the host to be load & run independently. It would be better to name them after the deployment models they are to be used in (host-in-VS vs host-in-non-TEE-HS), but I don't have a good suggestion at the moment :)

Cargo.toml
atulkharerivos: Feature flags can be enabled by default, but it does require the flag to be defined AFAIK. The nomenclature was arbitrary, so perhaps, we can just see what feels best in the channel.

PR #131, opened 2022-10-08
crate: use toml based files for cargo and toolchain

luojia65:
This pull request includes a rename of `.cargo/config` and `rust-toolchain` to `.cargo/config.toml` and `rust-toolchain.toml`, it rewrites `rust-toolchain.toml` so that IDEs may apply syntax highlights according to file extensions. The `rust-toolchain.toml` file functions equal to the old file, and it allows to install build targets upon first cargo build.


dgreid (approved):
Thanks! Both of these are the standard now, so thanks for catching us up.

It'll need a quick rebase on top of some lint fixes that just landed.

dgreid:
Thanks, LGTM, merging.

Issue #125, opened 2022-10-05
Compile with vector support enabled by default


abrestic-rivos:
Looks like we don't need a separate target for this and can instead add `-Ctarget-feature=+v` to `.cargo/config`:

```
[target.riscv64gc-unknown-none-elf]
rustflags = [
    '-Clink-arg=-Tlds/qemu.lds',
    '-Ctarget-feature=+v',
    # new in 1.60 and generates false positives
    "-Aclippy::only_used_in_recursion",
]
````

This would allow us to get rid of the conditional compilation that's littered around for vector support, giving us better compile-time coverage.

Two issues I noticed when running with `+v` enabled (on a non-V enabled QEMU config):
- `check_vector_width()` runs unconditionally. It should be guarded based on whether we detected vector support.
- Tellus unconditionally exercises vector save/restore. It should check if vectors are supported first, preferably via DT.

cc @stillson

stillson:
I'm on this.

PR #133, opened 2022-10-10
vectors: convert code to check if vector extension present

stillson:
In order to clean up the make files, salus has been converted to always have the vector code compiled and can detect if the code should run. The only time the user needs to specify if vectors are used is when qemu needs to know whether it should enable vectors or not. This addresses https://github.com/rivosinc/salus/issues/125


dgreid (commented):
Thanks, that is much cleaner than the ifdefs.

src/main.rs
```diff
@@ -67,6 +66,9 @@ extern "C" {
     static _stack_end: u8;
 }
 
+// stash the vector state for use deeper in the vm_cpu code
+static VECTORS_ENABLED: Once<bool> = Once::new();
+
```
abrestic-rivos, line 71: Do we really need this? It only gets used once, so I'm not sure it's that useful of a shortcut vs `CpuInfo::get().has_vector()`

device-tree/src/fdt.rs
```diff
@@ -59,6 +61,28 @@ impl<'a> Fdt<'a> {
         self.get_module_node_region("multiboot,ramdisk")
     }
 
+    /// This return the property buf for the first property with the give name
```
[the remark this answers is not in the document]
stillson: fixed

src/vm_cpu.rs
```diff
@@ -552,8 +542,9 @@ impl VmCpu {
         let mut sstatus = LocalRegisterCopy::<u64, sstatus::Register>::new(0);
         sstatus.modify(sstatus::spp::Supervisor);
         sstatus.modify(sstatus::fs::Initial);
-        #[cfg(target_feature = "v")]
-        sstatus.modify(sstatus::vs::Initial);
+        if *crate::VECTORS_ENABLED.call_once(|| panic!("Unsure of vectors state")) {
+            sstatus.modify(sstatus::vs::Initial);
+        }
```
abrestic-rivos, line 547: See my previous comment -- I think using CpuInfo here would be much simpler.

abrestic-rivos (commented):
Thanks. This should probably be 2-3 commits though:

1) to remove the conditional compilation from Salus
2) to remove it from tellus/guestvm
3) update docs/makefile

stillson:
Do you want me to submit again 3 times? That seems a bit off since the steps in the middle won't really work.

src/main.rs
stillson: didn't realize that CPU_INFO was static. will fix

src/vm_cpu.rs
stillson: see above

abrestic-rivos (approved):
> Do you want me to submit again 3 times? That seems a bit off since the steps in the middle won't really work.

No, I'm suggesting splitting the single commit in this PR into 3 commits, still all in this same PR. See the other recent Salus PRs for examples. Limiting each commit to a single, logical change makes things easier to review and to narrow down bugs later on when bisecting. See also the guidance here: https://www.kernel.org/doc/html/v6.0/process/submitting-patches.html#separate-your-changes

I do realize the split I first suggested is not great because the intermediate state of different compilation flags for different binaries is annoying. This is a better split:
1. Fix the `check_vector_length()` check so that it's conditional on the CPU actually supporting vectors.
2. Fix tellus/guestvm to use the device-tree/bootarg to check if the CPU supports vectors.
3. Remove all the conditional compilation and update docs/makefiles.

1 and 2 actually fix pre-existing bugs when running a binary compiled with vector support on a QEMU without vector support. And then 3 is the change to always compile with vector support. 

Anyway, LGTM since this is such a small PR. It would be good practice to get into though.

PR #137, opened 2022-10-18
Fix host VM address space size calculation

abrestic-rivos:
This fixes a panic encountered when trying to boot a host VM with reduced 64-bit PCI BAR space.


dgreid (commented):
lgtm

PR #141, opened 2022-10-20
Remove page_type from TsmConvert/Reclaim pages

atulkharerivos:
The page_type parameter in the TsmConvertPages and TsmReclaimPages was hardcoded to PageSize4K for all intents and purposes. The consensus following an internal discussion about the APIs was that we should remove the page_type parameter in APIs that don't directly affect page tables.

The rationale behind the change is that the HW mechanism for protecting confidential memory is not necessarily page-table based (example: PMP). Furthermore, even for page-table based implementations, there may be a mismatch between the requested page size and what the implementation can reasonably provide, resulting in additional complexity in both the host and TSM. Note that the TSM implementation is still free to internally optimize conversion of greater than 4kB-sized ranges if HW
support is available.

### DIFF
--- a/sbi/src/tee_host.rs
+++ b/sbi/src/tee_host.rs
@@ -259,7 +259,7 @@ pub enum TeeHostFunction {
         /// a1 = maximum number of bytes to be written
         len: u64,
     },
-    /// Converts `num_pages` of non-confidential memory starting at `page_addr`. The converted pages
+    /// Converts `num_pages` of 4kB page-size non-confidential memory starting at `page_addr`. The converted pages
     /// remain non-confidential, and thus may not be assigned for use by a child TVM, until the
     /// fence procedure, described below, has been completed.
     ///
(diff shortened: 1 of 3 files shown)


abrestic-rivos (approved):

sbi/src/tee_host.rs
```diff
@@ -259,29 +259,25 @@ pub enum TeeHostFunction {
...
     ///
     /// a6 = 1
     TsmConvertPages {
         /// a0 = base address of pages to convert
         page_addr: u64,
-        /// a1 = page size
-        page_type: TsmPageType,
-        /// a2 = number of pages
+        /// a1 = number of pages
         num_pages: u64,
```
atishp04, line 271: Nit:
The comment here needs to be updated.
a1 = num_pages

atishp04 (approved):
Apart from the nit comment:
LGTM.

Issue #139, opened 2022-10-20
Print address in hex during the panic


atishp04:
I noticed one during a page fault panic
Unhandled page fault: InvalidAddress(2188476416)

As a general rule, we should print all the address in hex

atulkharerivos:
https://github.com/rivosinc/salus/commit/4ac65493ce0ef81ceb563e6fdcb610f1eb8de69a

PR #143, opened 2022-10-21
Add Display trait for MmioEmulationError

atulkharerivos:
This includes customizations like outputting the page fault address in hex. Closes https://github.com/rivosinc/salus/issues/139.

### DIFF
--- a/src/vm.rs
+++ b/src/vm.rs
@@ -1809,6 +1809,16 @@ enum MmioEmulationError {
     InvalidAddress(u64),
 }
 
+impl core::fmt::Display for MmioEmulationError {
+    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+        match self {
+            MmioEmulationError::FailedDecode(i) => write!(f, "FailedDecode {i:x}"),
+            MmioEmulationError::InvalidInstruction(i) => write!(f, "InvalidInstruction {i:?}"),
+            MmioEmulationError::InvalidAddress(i) => write!(f, "InvalidAddress {i:x}"),
+        }
+    }
+}
+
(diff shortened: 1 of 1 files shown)


atishp04 (commented):
Thanks for the quick fix. I am just curious why it is called MmioEmulationError ?
MmioEmulation is certainly one of the use case. 

The host might be accessing a confidential page by mistake. We will end up same fault path. Correct ?

src/vm.rs
```diff
@@ -1650,6 +1650,16 @@ enum MmioEmulationError {
     InvalidAddress(u64),
 }
 
+impl core::fmt::Display for MmioEmulationError {
```
abrestic-rivos, line 1653: I don't think this is sufficient; you'll need to also print the error using `{}` instead of `{:?}` to get the formatting implemented here.

abrestic-rivos (commented):
> I am just curious why it is called MmioEmulationError ?

Because the only type of page faults we can handle from the host are MMIO. So when we get a page fault, we always attempt to handle it as an MMIO fault.

src/vm.rs
atulkharerivos: Interesting...I wasn't aware of this subtlety until now...had assumed that {:?} would call either the default implementation or the override...fixed.

atishp04:
> > I am just curious why it is called MmioEmulationError ?
> 
> Because the only type of page faults we can handle from the host are MMIO. So when we get a page fault, we always attempt to handle it as an MMIO fault.

Saw the big comment in the code now :).

atishp04 (approved):
LGTM.

abrestic-rivos (approved):

PR #144, opened 2022-10-21
Some minor Vm/VmPages refactoring

abrestic-rivos:
While working on IMSIC bind/unbind support, I've found that the current pattern we have of implementing each "take pages from host to use in guest" as a unique `VmPages` API to get a bit messy since it introduces a bunch of otherwise unnecessary inter-dependencies. Move all of this coordination into `Vm` instead, making `VmPages` a bit more of an independent module.

Patch 1 is a purely cosmetic change to the logical organization of `vm.rs`, patch 2 simplifies VM creation a little bit by inlining the `VmRegionList` vector, and patch 3 does the actual refactor described above.

### DIFF
--- a/src/vm.rs
+++ b/src/vm.rs
@@ -11,7 +11,7 @@ use drivers::{
     CpuInfo,
 };
 use page_tracking::collections::PageBox;
-use page_tracking::{HypPageAlloc, PageList, PageTracker};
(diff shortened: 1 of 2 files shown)


atulkharerivos (approved):
LGTM.

dgreid (approved):

Issue #147, opened 2022-10-25
Update supported SBI version


dgreid:
Linux checks for version 1.0 or greater before checking if extensions such as debug console are supported.

Just uprevving the version causes issues with the fence extension though. Needs testing and debug.

Issue #146, opened 2022-10-25
Forward guest VM SBI console output to host


dgreid:
Salus shouldn't interleave SBI console output from guests with the host. Instead forward the guest SBI debug console ECALLs to the host.

This will depend on the output buffers being located in shared memory accessible from the host.

PR #148, opened 2022-10-25
Add debug console extension

### LINKED ISSUE #123: SBI DBCN Console in Salus
Can we implement SBI debug console extension in Salus ?
https://lists.riscv.org/g/tech-unixplatformspec/message/1815
The Linux patches can be found in the above email. Anup reports that it improves guest booting time by ~30%.
It should help boot time of Linux in Salus as well

dgreid:
fixes #123 

This still assumes that the console is owned by Salus. File #146 to address that.

We also need to uprev the supported SBI version, filed #147

### DIFF
--- a/s-mode-utils/src/sbi_console.rs
+++ b/s-mode-utils/src/sbi_console.rs
@@ -2,6 +2,7 @@
 // Licensed under the Apache License, Version 2.0, see LICENSE for details.
 // SPDX-License-Identifier: Apache-2.0
 
+use sbi::api::debug_console::console_puts;
 use sbi::SbiMessage;
 
 use crate::ecall::ecall_send;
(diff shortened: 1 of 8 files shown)


atishp04:
Did you notice any speed up while booting ? In my setup, I did not notice anything significant.
I tried booting both debian and a busybox rootfs.

dgreid:
I didn't see any speed up, but I'm pretty sure boot is still limited by PCI and DRAM page mapping. so ever a huge speedup would be in the noise for now.

s-mode-utils/src/sbi_console.rs
```diff
@@ -22,13 +21,7 @@ impl SbiConsole {
...
     /// Write an entire byte sequence to the SBI console.
     fn write_bytes(&self, bytes: &[u8]) {
-        for &b in bytes {
-            let message = SbiMessage::PutChar(b as u64);
-            // Safety: message doesn't contain pointers and the ecall doesn't touch memory so this
-            // is trivially safe.
-            unsafe {
-                ecall_send(&message).unwrap();
-            }
-        }
```
abrestic-rivos, line 32: Note that this will break early boot prints from Salus unless OpenSBI supports the DBCN. (There's a branch for that here: https://github.com/avpatel/opensbi/tree/riscv_sbi_dbcn_v1)

abrestic-rivos (approved):

atishp04:
Yeah. We reduced the PCI page mapping by limiting it to 2GB. DRAM Page mapping still takes a lot of time which can't also be avoided until Salus gets demand paging support.

dgreid:
added back the legacy console extension so salus can use it before probing the uart.

abrestic-rivos (approved):
thanks!

Issue #127, opened 2022-10-06
Publish docs on github pages


dgreid:
CI is now generating docs but throwing them away. It would be nice to push them to github pages so they can be easily browsed without locally generating them.

PR #152, opened 2022-11-01
docs: Add workflow to generate docs on checkin (issue #127)

stillson:
They will be hosted at https://rivosinc.github.io/salus/salus

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -1,0 +1,39 @@
+name: Docs
+
+on:
+  pull_request_target:
+    branches: ["main"]
+    types:
+      - closed
+  workflow_dispatch:
+
+env:
+  CARGO_TERM_COLOR: always
+
+permissions:
+  contents: read
+  pages: write
+  id-token: write
+
+jobs:
+  build:
+    environment:
+      name: github-pages
+      url: ${{ steps.deployment.outputs.page_url }}
+
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+    - name: Install riscv64 target
+      run: rustup target add riscv64gc-unknown-none-elf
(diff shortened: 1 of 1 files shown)


.github/workflows/docs.yml
```diff
@@ -0,0 +1,40 @@
...
+
+        runs-on: ubuntu-latest
+
+        steps:
+            - uses: actions/checkout@v3
+
+            - name: Install riscv64 target
+              run: rustup target add riscv64gc-unknown-none-elf
+            - name: Generate Docs
+              run: make doc
```
dgreid, line 31: Can the make doc step from the other flow be used here? Or, maybe the doc generation should be removed from the other flow as to avoid duplicating the work.

stillson: I think they are really two different things. One tests that the docs can be generated as part of CI, and the other is actually deploying the docs. I could possibly merge the files if that makes sense.

stillson: also, there are some differences (the environment, some of the actions imported) and it seems that while docs is nice, the rust workflow is essential and not worth messing with.

stillson: Ok, yes, that is a better way to do it. It's all in one workflow now.

dgreid (commented):
LGTM, can you squash to one commit before submitting?

stillson:
I'm changing this because we don't want to trigger doc rebuild on every PR, just on every accepted pr. This requires another workflow file.

stillson:
Now, docs are a separate workflow that only happens when the pull request is closed instead of on any pull request (which may make changes visible that aren't in the main branch). Can also be triggered manually.

atulkharerivos (approved):

dgreid (approved):

PR #154, opened 2022-11-08
vmem: huge page support

stillson:
- Extended SequentialPage testing to verify huge pages
- Created general version of 4k specific calls in page table
- Took out hugepage checks
- Function to compute expected page table size if using hugepages
- fixed some typos in comments

### DIFF
--- a/riscv-page-tables/src/lib.rs
+++ b/riscv-page-tables/src/lib.rs
@@ -195,7 +195,7 @@ mod tests {
                     page.size() as usize / mem::size_of::<u64>(),
                 );
                 slice[0] = 0xdeadbeef;
-                assert!(mapper.map_4k_addr(gpa, page.addr(), pte_fields).is_ok());
+                assert!(mapper.map_addr(gpa, page.addr(), pte_fields).is_ok());
             }
         }
     }
--- a/riscv-page-tables/src/page_table.rs
+++ b/riscv-page-tables/src/page_table.rs
@@ -19,12 +19,16 @@ pub enum Error {
     InsufficientPtePages,
     /// Attempt to access a middle-level page table, but found a leaf.
     LeafEntryNotTable,
+    /// Attempt to access a leaf page table, but found a middle-level.
+    TableEntryNotLeaf,
     /// Failure creating a root page table at an address that isn't aligned as required.
     MisalignedPages(SequentialPages<InternalClean>),
     /// Failure creating a root page table with an unsupported page size.
     UnsupportedPageSizePages(SequentialPages<InternalClean>, PageSize),
     /// The requested page size isn't (yet) handled by the hypervisor.
     PageSizeNotSupported(PageSize),
+    /// The requested size doesn't match what is found
+    PageSizeMismatch(PageSize, PageSize),
     /// Attempt to create a mapping over an existing one.
     MappingExists,
     /// The requested range isn't mapped.
@@ -45,6 +49,8 @@ pub enum Error {
     OutOfMapRange,
     /// The page cannot be shared
     PageNotShareable,
+    /// Address is not page aligned
+    AddressMisaligned(u64),
 }
 /// Hold the result of page table operations.
 pub type Result<T> = core::result::Result<T, Error>;
@@ -283,11 +289,6 @@ impl<'a, T: PagingMode> PageTable<'a, T> {
         }
     }
 
-    /// Returns the `PageTableLevel` this table is at.
-    fn level(&self) -> T::Level {
-        self.level
-    }
-
     /// Returns a mutable reference to the raw PTE at the given guest address.
     fn entry_mut(&mut self, index: PageTableIndex<T>) -> &'a mut Pte {
         let pte_addr =
@@ -482,6 +483,7 @@ pub trait GuestStagePagingMode: PagingMode<MappedAddressSpace = GuestPhys> {
 }
 
 /// The internal state of a paging hierarchy.
+#[derive(Debug)]
 struct PageTableInner<T: PagingMode> {
     root: SequentialPages<InternalClean>,
     table_type: PhantomData<T>,
@@ -524,18 +526,22 @@ impl<T: PagingMode> PageTableInner<T> {
     ///
     /// The caller must guarantee that `paddr` references a page uniquely owned by the root
     /// `GuestStagePageTable`.
-    unsafe fn map_4k_leaf(
+    unsafe fn map_leaf(
         &mut self,
         vaddr: PageAddr<T::MappedAddressSpace>,
         paddr: SupervisorPageAddr,
+        page_size: PageSize,
         perms: PteFieldBits,
     ) -> Result<()> {
         let entry = self.walk(RawAddr::from(vaddr));
         use TableEntryType::*;
         match entry {
             Locked(l) => {
-                if !l.level().is_leaf() {
-                    return Err(Error::PageSizeNotSupported(l.level().leaf_page_size()));
+                if l.level().leaf_page_size() != page_size {
+                    return Err(Error::PageSizeMismatch(
+                        page_size,
+                        l.level().leaf_page_size(),
+                    ));
                 }
                 l.map_leaf(paddr, perms);
                 Ok(())
@@ -548,13 +554,15 @@ impl<T: PagingMode> PageTableInner<T> {
 
     /// Locks the invalid leaf PTE mapping `vaddr`, filling in any missing intermediate page tables
     /// using `get_pte_page`.
-    fn lock_4k_leaf_for_mapping(
+    fn lock_leaf_for_mapping(
         &mut self,
         vaddr: PageAddr<T::MappedAddressSpace>,
+        page_size: PageSize,
         get_pte_page: &mut dyn FnMut() -> Option<Page<InternalClean>>,
     ) -> Result<()> {
         let mut table = PageTable::from_root(self);
-        while !table.level().is_leaf() {
+        // match the level to the page size
+        while table.level.leaf_page_size() != page_size {
             table = table.next_level_or_fill_fn(RawAddr::from(vaddr), get_pte_page)?;
         }
         let entry = table.entry_for_addr_mut(RawAddr::from(vaddr));
@@ -570,12 +578,12 @@ impl<T: PagingMode> PageTableInner<T> {
             }
             Locked(_) => Err(Error::PteLocked),
             Leaf(_) => Err(Error::MappingExists),
-            Table(_) => unreachable!(),
+            Table(_) => Err(Error::TableEntryNotLeaf),
         }
     }
 
     /// Unlocks the leaf PTE mapping `vaddr`.
-    fn unlock_4k_leaf(&mut self, vaddr: PageAddr<T::MappedAddressSpace>) -> Result<()> {
+    fn unlock_leaf(&mut self, vaddr: PageAddr<T::MappedAddressSpace>) -> Result<()> {
         let entry = self.walk(RawAddr::from(vaddr));
         use TableEntryType::*;
         match entry {
@@ -624,7 +632,6 @@ impl<T: PagingMode> PageTableInner<T> {
 
 /// A paging hierarchy for a given addressing type.
 /// This is used for HS and HU paging modes such as Sv48.
-///
 /// TODO: Support non-4k page sizes.
 pub struct FirstStagePageTable<T: FirstStagePagingMode> {
     inner: Mutex<PageTableInner<T>>,
@@ -658,16 +665,21 @@ impl<T: FirstStagePagingMode> FirstStagePageTable<T> {
         num_pages: u64,
         get_pte_page: &mut dyn FnMut() -> Option<Page<InternalClean>>,
     ) -> Result<FirstStageMapper<T>> {
-        if page_size.is_huge() {
-            return Err(Error::PageSizeNotSupported(page_size));
+        if !addr.is_aligned(page_size) {
+            return Err(Error::AddressMisaligned(addr.bits()));
         }
+        let iterator = addr
+            .iter_from_with_size(page_size)
+            .ok_or_else(|| Error::AddressMisaligned(addr.bits()))?;
+        addr.checked_add_pages_with_size(num_pages, page_size)
+            .ok_or(Error::OutOfMapRange)?;
 
         let mut inner = self.inner.lock();
-        for a in addr.iter_from().take(num_pages as usize) {
-            inner.lock_4k_leaf_for_mapping(a, get_pte_page)?;
+        for a in iterator.take(num_pages as usize) {
+            inner.lock_leaf_for_mapping(a, page_size, get_pte_page)?;
         }
 
-        Ok(FirstStageMapper::new(self, addr, num_pages))
+        Ok(FirstStageMapper::new(self, addr, page_size, num_pages))
     }
 }
 
@@ -677,6 +689,7 @@ impl<T: FirstStagePagingMode> FirstStagePageTable<T> {
 pub struct FirstStageMapper<'a, T: FirstStagePagingMode> {
     owner: &'a FirstStagePageTable<T>,
     vaddr: PageAddr<T::MappedAddressSpace>,
+    page_size: PageSize,
     num_pages: u64,
 }
 
@@ -685,11 +698,13 @@ impl<'a, T: FirstStagePagingMode> FirstStageMapper<'a, T> {
     fn new(
         owner: &'a FirstStagePageTable<T>,
         vaddr: PageAddr<T::MappedAddressSpace>,
+        page_size: PageSize,
         num_pages: u64,
     ) -> Self {
         Self {
             owner,
             vaddr,
+            page_size,
             num_pages,
         }
     }
@@ -700,19 +715,24 @@ impl<'a, T: FirstStagePagingMode> FirstStageMapper<'a, T> {
     /// # Safety
     ///
     /// Don't create aliases.
-    pub unsafe fn map_4k_addr(
+    pub unsafe fn map_addr(
         &self,
         vaddr: PageAddr<T::MappedAddressSpace>,
         paddr: PageAddr<SupervisorPhys>,
         pte_perms: PteFieldBits,
     ) -> Result<()> {
-        let end_vaddr = self.vaddr.checked_add_pages(self.num_pages).unwrap();
-        if vaddr < self.vaddr || vaddr >= end_vaddr {
-            return Err(Error::OutOfMapRange);
+        if !paddr.is_aligned(self.page_size) {
+            return Err(Error::AddressMisaligned(paddr.bits()));
         }
 
+        if vaddr < self.vaddr
+            || vaddr.bits() >= self.vaddr.bits() + self.num_pages * self.page_size as u64
+        {
+            return Err(Error::OutOfMapRange);
+        }
         let mut inner = self.owner.inner.lock();
-        inner.map_4k_leaf(vaddr, paddr, pte_perms)
+
+        inner.map_leaf(vaddr, paddr, self.page_size, pte_perms)
     }
 }
 
@@ -789,7 +809,7 @@ impl<T: PagingMode> GuestStagePageTable<T> {
         let mut mapper = GuestStageMapper::new(self, addr, 0);
         let mut inner = self.inner.lock();
         for a in addr.iter_from().take(num_pages as usize) {
-            inner.lock_4k_leaf_for_mapping(a, get_pte_page)?;
+            inner.lock_leaf_for_mapping(a, page_size, get_pte_page)?;
             mapper.num_pages += 1;
         }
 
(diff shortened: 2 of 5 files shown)


riscv-page-tables/src/page_table.rs
```diff
@@ -465,6 +460,10 @@ pub trait PagingMode {
     /// Calculates the number of PTE pages that are needed to map all pages for `num_pages` mapped
     /// pages for this type of page table.
     fn max_pte_pages(num_pages: u64) -> u64;
+
+    /// Used to calculate the number of PTE pages needed if all the pages are
+    /// not 4k
+    fn max_huge_pte_pages(num_pages: u64, page_size: PageSize) -> u64;
```
dgreid, line 466: What will `max_huge_pte_pages` be used for? If used to minimize the number of pages needed, it would limit the ability to break huge pages to 4k. That might be awkward.

riscv-page-tables/src/page_table.rs
```diff
@@ -518,6 +518,38 @@ impl<T: PagingMode> PageTableInner<T> {
...
 
+    /// Creates a translation for `vaddr` to `paddr` with the given permissions.
+    ///Like map_4k_leaf, but with an added page size parameter.
+    ///
+    /// # Safety
+    ///
+    /// The caller must guarantee that `paddr` references a page uniquely owned by the root
+    /// `GuestStagePageTable`.
+
+    unsafe fn map_general_leaf(
```
dgreid, line 529: Is there a reason to keep `map_4k_leaf`?

riscv-page-tables/src/page_table.rs
```diff
@@ -658,13 +698,13 @@ impl<T: FirstStagePagingMode> FirstStagePageTable<T> {
...
-        if page_size.is_huge() {
-            return Err(Error::PageSizeNotSupported(page_size));
-        }
-
         let mut inner = self.inner.lock();
-        for a in addr.iter_from().take(num_pages as usize) {
-            inner.lock_4k_leaf_for_mapping(a, get_pte_page)?;
+        for a in addr
+            .iter_from_with_size(page_size)
+            .unwrap()
```
dgreid, line 704: error instead of unwrap?

riscv-page-tables/src/page_table.rs
abrestic-rivos, line 466: This and `max_pte_pages()` should probably take a length rather than a number of (assumed to be) 4kB pages.

Also, `max_huge_pte_pages()` works just fine with a non-4kB size; `max_pte_pages_with_size()` would be a better name. Then you could do something like:

```
fn max_pte_pages_with_size(len: u64, page_size: PageSize) -> u64;

fn max_pte_pages(len: u64) -> u64 {
    Self::max_pte_pages_with_size(len, PageSize::Size4k)
}
```

(BTW -- where is this function used? I don't see it called anywhere in this patch.)

riscv-page-tables/src/page_table.rs
```diff
@@ -548,16 +580,23 @@ impl<T: PagingMode> PageTableInner<T> {
 
     /// Locks the invalid leaf PTE mapping `vaddr`, filling in any missing intermediate page tables
     /// using `get_pte_page`.
-    fn lock_4k_leaf_for_mapping(
+    /// Locks the invalid leaf PTE mapping `vaddr`, filling in any missing intermediate page tables
+    /// using `get_pte_page`.
+    fn lock_general_leaf_for_mapping(
```
dgreid, line 585: nit: drop the 'general', it's cleaner.

`lock_leaf_for_mapping`

riscv-page-tables/src/page_table.rs
abrestic-rivos, line 529: This is basically identical to `map_4k_leaf()`. How about updating `map_4k_leaf()` to take a `page_size` parameter (and renaming it to just `map_leaf()`) instead?

riscv-page-tables/src/page_table.rs
```diff
@@ -518,6 +518,38 @@ impl<T: PagingMode> PageTableInner<T> {
...
+        page_size: PageSize,
+        perms: PteFieldBits,
+    ) -> Result<()> {
+        let entry = self.walk(RawAddr::from(vaddr));
+        use TableEntryType::*;
+        match entry {
+            Locked(l) => {
+                if l.level().leaf_page_size() != page_size {
+                    // maybe not?
+                    return Err(Error::PageSizeNotSupported(l.level().leaf_page_size()));
```
abrestic-rivos, line 542: A `PageSizeMismatch` is probably a better fit in this case.

riscv-page-tables/src/page_table.rs
```diff
@@ -570,12 +609,12 @@ impl<T: PagingMode> PageTableInner<T> {
             }
             Locked(_) => Err(Error::PteLocked),
             Leaf(_) => Err(Error::MappingExists),
-            Table(_) => unreachable!(),
+            Table(_) => unreachable!(), // good test of problems with page sizes
```
abrestic-rivos, line 612: This is case is no longer unreachable in the presence of huge pages. Consider the case where a range of address space is already mapped with 4kB pages and the user tries to map in 2MB pages. We should return an error instead of blowing up here.

riscv-page-tables/src/page_table.rs
```diff
@@ -714,6 +754,34 @@ impl<'a, T: FirstStagePagingMode> FirstStageMapper<'a, T> {
...
+        &self,
+        vaddr: PageAddr<T::MappedAddressSpace>,
+        paddr: PageAddr<SupervisorPhys>,
+        pte_perms: PteFieldBits,
+        page_size: PageSize,
+    ) -> Result<()> {
+        let end_vaddr = self
+            .vaddr
+            .checked_add_pages_with_size(self.num_pages, page_size)
+            .unwrap();
```
dgreid, line 775: error instead of unwrap.

riscv-page-tables/src/page_table.rs
```diff
@@ -714,6 +754,34 @@ impl<'a, T: FirstStagePagingMode> FirstStageMapper<'a, T> {
...
     }
+
+    /// Maps `vaddr` to `paddr`, The caller must guarantee that paddr points to a page it is safe to
+    /// map in this page table.
+    ///
+    /// This functions works like map_4k_addr, but takes a page size parameter.
+    /// # Safety
+    ///
+    /// Don't create aliases.
+    pub unsafe fn map_general_addr(
```
dgreid, line 765: Should `map_4k_addr` be changed to call `map_general_addr` with PageSize4k specified? Also, `map_addr` instead of `map_general_addr`.

riscv-page-tables/src/page_table.rs
```diff
@@ -658,13 +698,13 @@ impl<T: FirstStagePagingMode> FirstStagePageTable<T> {
...
     ) -> Result<FirstStageMapper<T>> {
-        if page_size.is_huge() {
-            return Err(Error::PageSizeNotSupported(page_size));
-        }
-
         let mut inner = self.inner.lock();
-        for a in addr.iter_from().take(num_pages as usize) {
-            inner.lock_4k_leaf_for_mapping(a, get_pte_page)?;
+        for a in addr
+            .iter_from_with_size(page_size)
```
abrestic-rivos, line 703: Need to check that `addr` is aligned to `page_size` first.

riscv-pages/src/page.rs
```diff
@@ -38,6 +38,22 @@ impl PageSize {
...
+            PageSize::Size4k => 12,
+            PageSize::Size2M => 21,
+            PageSize::Size1G => 30,
+            PageSize::Size512G => 39,
+        }
+    }
+
+    /// Return the number of pages of this `PageSize` that will fit in `val`.
+    /// This rounds down the number of pages, unlike num_4k_pages() which rounds up.
+    pub fn num_pages_contained_in(&self, val: u64) -> u64 {
```
dgreid, line 54: nit: would val be better names size_bytes? It took me a minute to figure out what it meant.

riscv-pages/src/page.rs
```diff
@@ -38,6 +38,22 @@ impl PageSize {
     pub const fn num_4k_pages(val: u64) -> u64 {
         (val + PageSize::Size4k as u64 - 1) / (PageSize::Size4k as u64)
     }
+    /// Return the number of bits to shift to get one page.
+    /// i.e. 4096 == 2^12 == 1 << 12
+    pub fn page_shift(&self) -> u64 {
```
dgreid, line 43: This duplicates some code in PageTable::addr_shift, would be good to have that function use this if possible.

riscv-page-tables/src/page_table.rs
```diff
@@ -714,6 +754,34 @@ impl<'a, T: FirstStagePagingMode> FirstStageMapper<'a, T> {
...
+            .checked_add_pages_with_size(self.num_pages, page_size)
+            .unwrap();
+
+        if vaddr < self.vaddr || vaddr >= end_vaddr {
+            return Err(Error::OutOfMapRange);
+        }
+
+        let mut inner = self.owner.inner.lock();
+
+        inner.map_general_leaf(vaddr, paddr, page_size, pte_perms)
```
abrestic-rivos, line 783: Need to check alignment of `vaddr` and `paddr` here (or within `map_leaf()`).

dgreid (commented):
Thanks for adding the testing up front, that makes it much easier to reason about.

Couple comments.

Also it is probably worth breaking up to a few commits. any time there is a bullet list in a commit message, it's a good indication it can be broken down further.

riscv-page-tables/src/sv48x4.rs
```diff
@@ -95,4 +95,29 @@ impl PagingMode for Sv48x4 {
         let num_l4_pages = 4;
         num_l1_pages + num_l2_pages + num_l3_pages + num_l4_pages
     }
+
+    fn max_huge_pte_pages(num_pages: u64, page_size: PageSize) -> u64 {
```
abrestic-rivos, line 99: I'm having a hard time making sense of this. Is `num_pages` in units of `page_size`, or is it always 4kB? I'm skeptical of the math either way -- take for example 4MB of 4kB pages (`num_pages = 1024`). This looks like it will return 4 PTE pages, which is most definitely incorrect.

This would be a lot easier to follow if we specified the amount of address space in number of bytes (or start + end) rather than number of pages (see comment above).

riscv-pages/src/page.rs
```diff
@@ -38,6 +38,22 @@ impl PageSize {
...
+            PageSize::Size1G => 30,
+            PageSize::Size512G => 39,
+        }
+    }
+
+    /// Return the number of pages of this `PageSize` that will fit in `val`.
+    /// This rounds down the number of pages, unlike num_4k_pages() which rounds up.
+    pub fn num_pages_contained_in(&self, val: u64) -> u64 {
+        val >> self.page_shift()
+    }
```
abrestic-rivos, line 56: Don't see this used anywhere?

riscv-pages/src/sequential_pages.rs
```diff
@@ -39,7 +39,7 @@ pub struct SequentialPages<S: State> {
 }
```
abrestic-rivos, line 39: I'm all for more tests, but these changes are completely unrelated from the rest of the commit that deals with mapping hugepages in a page table and so should go in their own commit.

riscv-page-tables/src/page_table.rs
```diff
@@ -283,11 +283,6 @@ impl<'a, T: PagingMode> PageTable<'a, T> {
         }
```
abrestic-rivos, line 283: Commit message needs some work. Good commit messages should have prose describing the change being made and why, along with possibly highlighting any particularly tricky or notable implementation details.

riscv-page-tables/src/page_table.rs
stillson: gone

riscv-page-tables/src/page_table.rs
stillson: fixed

riscv-page-tables/src/page_table.rs
stillson: fixed

riscv-page-tables/src/page_table.rs
stillson: fixed

riscv-pages/src/page.rs
stillson: fixed. It's not the right function name, but I'm not sure what to call it.

riscv-page-tables/src/page_table.rs
stillson: fixed.

riscv-page-tables/src/page_table.rs
stillson: fixed

riscv-page-tables/src/page_table.rs
stillson: Got rid of the function. May come back later.

riscv-page-tables/src/page_table.rs
stillson: done

riscv-page-tables/src/page_table.rs
stillson: Fixed

riscv-page-tables/src/page_table.rs
stillson: fixed

riscv-page-tables/src/page_table.rs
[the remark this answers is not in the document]
stillson: fixed

riscv-page-tables/src/page_table.rs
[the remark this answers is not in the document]
stillson: done

riscv-page-tables/src/page_table.rs
```diff
@@ -727,10 +727,10 @@ impl<'a, T: FirstStagePagingMode> FirstStageMapper<'a, T> {
         page_size: PageSize,
```
abrestic-rivos, line 727: Please update the commit message. The subject is currently the same as the first patch; something like "SequentialPages: Add huge page tests" would be more appropraite.

riscv-page-tables/src/page_table.rs
```diff
@@ -624,8 +632,7 @@ impl<T: PagingMode> PageTableInner<T> {
 
 /// A paging hierarchy for a given addressing type.
 /// This is used for HS and HU paging modes such as Sv48.
-///
-/// TODO: Support non-4k page sizes.
```
abrestic-rivos, line 628: Not sure this TODO is ready to come out yet -- this patch adds support for _mapping_ huge-pages, but we still need to support unmap and invalidation.

riscv-page-tables/src/page_table.rs
```diff
@@ -658,16 +665,21 @@ impl<T: FirstStagePagingMode> FirstStagePageTable<T> {
...
+            return Err(Error::AddressMisaligned(addr.bits()));
         }
 
-        let mut inner = self.inner.lock();
-        for a in addr.iter_from().take(num_pages as usize) {
-            inner.lock_4k_leaf_for_mapping(a, get_pte_page)?;
+        let iterator = addr
+            .iter_from_with_size(page_size)
+            .ok_or_else(|| Error::AddressMisaligned(addr.bits()))?;
+
```
abrestic-rivos, line 677: nit: unnecessary newline

riscv-page-tables/src/page_table.rs
```diff
@@ -658,16 +665,21 @@ impl<T: FirstStagePagingMode> FirstStagePageTable<T> {
         num_pages: u64,
         get_pte_page: &mut dyn FnMut() -> Option<Page<InternalClean>>,
     ) -> Result<FirstStageMapper<T>> {
-        if page_size.is_huge() {
-            return Err(Error::PageSizeNotSupported(page_size));
+        let mut inner = self.inner.lock();
+
+        if !addr.is_aligned(page_size) {
+            return Err(Error::AddressMisaligned(addr.bits()));
         }
```
abrestic-rivos, line 672: In general it's better to put these sanity checks first, before acquiring a lock, if possible.

riscv-page-tables/src/page_table.rs
```diff
@@ -700,19 +722,33 @@ impl<'a, T: FirstStagePagingMode> FirstStageMapper<'a, T> {
...
+            return Err(Error::AddressMisaligned(vaddr.bits()));
+        }
+        if !paddr.is_aligned(self.page_size) {
+            return Err(Error::AddressMisaligned(paddr.bits()));
+        }
+
+        if let Some(end_vaddr) = self
+            .vaddr
+            .checked_add_pages_with_size(self.num_pages, self.page_size)
+        {
```
abrestic-rivos, line 741: This check is now redundant.

riscv-page-tables/src/page_table.rs
abrestic-rivos: Getting better, though "huge page support" is still overly generic. Something like "riscv-page-tables: Add support for mapping huge pages" would be more appropriate.

riscv-page-tables/src/page_table.rs
```diff
@@ -45,6 +49,8 @@ pub enum Error {
     OutOfMapRange,
     /// The page cannot be shared
     PageNotShareable,
+    /// Invalid address, should be page aligned
```
abrestic-rivos, line 52: nit: "Address is not page-size aligned" would be more clear

riscv-page-tables/src/page_table.rs
```diff
@@ -685,13 +698,22 @@ impl<'a, T: FirstStagePagingMode> FirstStageMapper<'a, T> {
...
         num_pages: u64,
-    ) -> Self {
-        Self {
+    ) -> Result<Self> {
+        if vaddr
+            .checked_add_pages_with_size(num_pages, page_size)
+            .is_none()
+        {
+            return Err(Error::OutOfMapRange);
+        }
```
abrestic-rivos, line 709: Thinking about this more, there's a subtle problem with returning an error here -- it leaves the PTEs locked, effectively leaking them. I think it's probably best to do the overflow check in the caller, before we start locking PTEs.

abrestic-rivos (commented):
Thanks, this is getting closer.

riscv-page-tables/src/page_table.rs
```diff
@@ -624,7 +632,6 @@ impl<T: PagingMode> PageTableInner<T> {
 
 /// A paging hierarchy for a given addressing type.
 /// This is used for HS and HU paging modes such as Sv48.
-///
 /// TODO: Support non-4k page sizes.
```
glguida, line 635: Can this TODO be removed or there's still some non-4k pages support to be added?

glg-rv, line 635: Can this TODO be removed now or there still work to be done to support non-4k pages?

riscv-page-tables/src/lib.rs
```diff
@@ -195,7 +195,7 @@ mod tests {
                     page.size() as usize / mem::size_of::<u64>(),
```
abrestic-rivos, line 195: you can remove the bulleted list from the commit message now

riscv-page-tables/src/page_table.rs
abrestic-rivos: I actually asked for this to stay as there's quite a lot left -- we still need to support invalidating & unmapping hugepages, and then there's the whole interaction with `PageTracker`, which assumes everything is 4kB.

riscv-page-tables/src/page_table.rs
abrestic-rivos: not done

riscv-page-tables/src/page_table.rs
```diff
@@ -685,13 +703,15 @@ impl<'a, T: FirstStagePagingMode> FirstStageMapper<'a, T> {
     fn new(
         owner: &'a FirstStagePageTable<T>,
         vaddr: PageAddr<T::MappedAddressSpace>,
+        page_size: PageSize,
         num_pages: u64,
-    ) -> Self {
-        Self {
+    ) -> Result<Self> {
```
abrestic-rivos, line 708: no longer needs to return a Result

riscv-page-tables/src/page_table.rs
```diff
@@ -658,16 +665,26 @@ impl<T: FirstStagePagingMode> FirstStagePageTable<T> {
...
-            inner.lock_4k_leaf_for_mapping(a, get_pte_page)?;
+        let iterator = addr
+            .iter_from_with_size(page_size)
+            .ok_or_else(|| Error::AddressMisaligned(addr.bits()))?;
+        if addr
+            .checked_add_pages_with_size(num_pages, page_size)
+            .is_none()
+        {
+            return Err(Error::OutOfMapRange);
+        }
```
abrestic-rivos, line 682: can be written more concisely as:

```
addr.checked_add_pages_with_size(num_pages, page_size).ok_or(Error::OutOfMapRnage)?;
```

riscv-page-tables/src/page_table.rs
glg-rv: Yes, just realised watching the history of reviewing.

abrestic-rivos (approved):

Issue #91, opened 2022-09-08
IMSIC virtualization support


abrestic-rivos:
Tracking bug for supporting IMSIC virtualization in TVMs. So far we have support for making IMSIC interrupt files confidential (and then reclaiming them). We will also need support for:
- [x] Binding interrupt files to TVM vCPUs
- [x] Migrating TVM vCPUs between physical CPUs, which includes saving/restoring IMSIC state to a SW interrupt file.
- [x] Injecting interrupts into a TVM vCPU's IMSIC.
- [ ] SGEI virtualization (i.e. `hgeie`/`hgeip` handling and injecting SGEI)

abrestic-rivos:
The IMSIC part of this is done -- at least enough to get virtio working between a host and TVM guest.

The only large remaining item is virtualization of supervisor guest external interrupts, which requires emulating the `hgeie`/`hgeip`/`hie` CSRs, and injecting an SGEI into the host VM when one occurs. I've been working on this a bit and it gets pretty awkward with our current exit reporting structure (e.g. `scause` being per guest vCPU). I might go and move us to something more NACL-like where H-CSRs are in shared memory per physical CPU and `scause`/`stval` are written directly.

PR #155, opened 2022-11-08
Rework vCPU locking

abrestic-rivos:
Patch 1 reworks locking within VmCpu in preparation for interrupt injection (#91), while patch 2 is a trivial change to prevent accidental sharing of per-CPU state between threads.

### DIFF
--- a/src/vm_cpu.rs
+++ b/src/vm_cpu.rs
@@ -12,7 +12,7 @@ use riscv_page_tables::GuestStagePagingMode;
 use riscv_pages::{GuestPhysAddr, GuestVirtAddr, PageOwnerId, PageSize, RawAddr};
 use riscv_regs::*;
 use sbi::{self, SbiMessage, SbiReturnType};
-use spin::{Mutex, MutexGuard, Once, RwLock, RwLockReadGuard};
+use spin::{Mutex, MutexGuard, Once, RwLock};
 
 use crate::smp::PerCpu;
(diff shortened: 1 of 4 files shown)


src/vm_cpu.rs
```diff
@@ -1026,14 +987,14 @@ impl<T: GuestStagePagingMode> Drop for ActiveVmCpu<'_, '_, '_, T> {
...
-            .get(self.vcpu.vcpu_id as usize)
-            .unwrap();
-        let entry = once_entry.get().unwrap();
-        let mut status = entry.status.write();
+        // Avoid the lock order iversion by manually dropping the VmCpu::arch lock before taking
+        // the VmCpu::status lock (although it is safe from deadlock since status is checked before
+        // acquiring arch).
+        //
+        // Safety: self.arch is valid for dropping and we do not access it beyond this point.
+        unsafe { ManuallyDrop::drop(&mut self.arch) };
```
dgreid, line 995: I'll send you a patch to consider to avoid this unsafe. Because drop order is guaranteed to be the order listed in the struct, we can add a guard type to do the status set below and avoid manually drop.

It makes the code harder to read if you're not used to following drop chains though so feel free to ignore it.

abrestic-rivos: thanks, i prefer that to `ManuallyDrop` -- applied

dgreid (approved):
Thanks!

PR #159, opened 2022-11-10
Pin nightly toolchain

dgreid:
This will make sure everyone is building with the same toolchain, allowing there to be a common set of nightly features that need to be declared and avoiding having to get everyone to upgrade when features are stabilized.

Signed-off-by: Dylan Reid <dgreid@rivosinc.com>

### DIFF
--- a/rust-toolchain.toml
+++ b/rust-toolchain.toml
@@ -1,5 +1,5 @@
 [toolchain]
-channel = "nightly"
+channel = "nightly-2022-11-10"
 components = []
 targets = ["riscv64gc-unknown-none-elf"]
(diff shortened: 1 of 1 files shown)


dgreid:
> Who will keep this up to date?

One assumption here is that keeping up to date isn't important. And when it is, we'll explicitly update this file instead of reacting.

When a new version is required, someone can update this and fix new warning/errors etc atomically.

dgreid:
> Looks like it needs to be updated to a nightly that has stable `int_log`?

Done, and rebased on stillson's change for int_log.

PR #160, opened 2022-11-10
Support external interrupt injection

abrestic-rivos:
This series adds support for secure external interrupts by introducing the TEE-Guest calls to allow TVMs to manipulate their per-vCPU allowlist of injectable interrupts, and the TEE-Host call to actually inject the external interrupt from the host side. Guests start with injection of external interrupts disabled by default.

Patch 1 defines the new TEE calls, patches 2-4 are cleanups / prep-work to support interrupt injection, patch 5 implements the allow-listing, and patch 6 implements the interrupt injection itself. This is all (minimally) exercised in patch 7 by having GuestVm allow an external interrupt and then having Tellus inject it. While we don't yet have a trap handler in GuestVm, I've verified by manually inspecting the IMSIC CSRs that the injected interrupt was recorded.

This covers the interrupt injection support part of #91

### DIFF
--- a/drivers/src/imsic/core.rs
+++ b/drivers/src/imsic/core.rs
@@ -17,6 +17,9 @@ use super::geometry::*;
 use super::sw_file::SwFile;
 use crate::{CpuId, CpuInfo, MAX_CPUS};
 
+/// The maximum number of IMSIC interrupt IDs, as per the AIA specification.
+pub const MAX_INTERRUPT_IDS: usize = 2048;
+
 const MAX_GUEST_FILES: usize = 7;
 const MAX_MMIO_REGIONS: usize = 8;
 
@@ -265,7 +268,7 @@ pub struct Imsic {
     per_cpu: Mutex<ImsicCpuState>,
     mmio_regions: ArrayVec<SupervisorPageRange, MAX_MMIO_REGIONS>,
     geometry: ImsicGeometry<SupervisorPhys>,
-    interrupt_ids: u32,
+    interrupt_ids: usize,
     phandle: u32,
 }
 
@@ -315,8 +318,11 @@ impl Imsic {
         let interrupt_ids = imsic_node
             .props()
             .find(|p| p.name() == "riscv,num-ids")
-            .and_then(|p| p.value_u32().next())
+            .and_then(|p| p.value_u32().next().map(|v| v as usize))
             .ok_or(Error::MissingProperty("riscv,num-ids"))?;
+        if interrupt_ids == 0 || interrupt_ids > MAX_INTERRUPT_IDS {
+            return Err(Error::InvalidInterruptIds(interrupt_ids));
+        }
 
         // We must have at least one guest interrupt file to start the host.
         let guest_index_bits = imsic_node
@@ -521,6 +527,11 @@ impl Imsic {
         .unwrap()
     }
 
+    /// Returns the number of implemented external interrupt IDs.
+    pub fn interrupt_ids(&self) -> usize {
+        self.interrupt_ids
+    }
+
     /// Takes ownership over the guest interrupt file pages for `cpu`, returning an iterator over
     /// the pages.
     pub fn take_guest_files(&self, cpu: CpuId) -> Result<ImsicGuestPageIter> {
@@ -579,7 +590,7 @@ impl Imsic {
 
     // Returns the number EIE/EIP registers used by the IMSIC.
     fn num_ei_regs(&self) -> usize {
-        (self.interrupt_ids as usize + 63) / 64
+        (self.interrupt_ids + 63) / 64
     }
 
     // Returns an ImsicGuestCsrAccess struct providing scoped access to guest_file's CSRs.
@@ -704,7 +715,7 @@ impl Imsic {
         imsic_node.add_prop("riscv,ipi-id")?.set_value_u32(&[1])?;
         imsic_node
             .add_prop("riscv,num-ids")?
-            .set_value_u32(&[self.interrupt_ids])?;
+            .set_value_u32(&[self.interrupt_ids as u32])?;
         if self.geometry.group_index_bits() > 0 {
             // These only matter when we have multiple groups.
             imsic_node
--- a/drivers/src/imsic/core.rs
+++ b/drivers/src/imsic/core.rs
@@ -550,16 +550,20 @@ impl Imsic {
         Ok(iter)
     }
 
-    /// Returns the IMSIC location specifier of the supervisor level interrupt file for the given
-    /// CPU.
-    pub fn supervisor_file_location(&self, cpu: CpuId) -> Result<ImsicLocation> {
+    /// Returns the IMSIC location specifier of the interrupt file on the specified physical CPU.
+    pub fn phys_file_location(&self, cpu: CpuId, file: ImsicFileId) -> Result<ImsicLocation> {
         let cpus = self.per_cpu.lock();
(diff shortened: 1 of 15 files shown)


drivers/src/imsic/core.rs
```diff
@@ -550,37 +550,45 @@ impl Imsic {
...
-        Ok(ImsicLocation::new(
-            pcpu.group,
-            pcpu.hart,
-            ImsicFileId::supervisor(),
-        ))
+        Ok(ImsicLocation::new(pcpu.group, pcpu.hart, file))
+    }
+
+    /// Returns the address of the interrupt file on the specified physical CPU.
+    pub fn phys_file_addr(&self, cpu: CpuId, file: ImsicFileId) -> Result<SupervisorPageAddr> {
```
dgreid, line 561: Does phys_file_addr need to be pub?

drivers/src/imsic/core.rs
```diff
@@ -550,37 +550,45 @@ impl Imsic {
...
 
-    /// Sends an IPI to the specified CPU.
-    pub fn send_ipi(&self, cpu: CpuId) -> Result<()> {
-        // If the CPU is valid, then its IMSIC location must be valid.
-        let addr = self
-            .geometry
-            .location_to_addr(self.supervisor_file_location(cpu)?)
-            .unwrap();
+    /// Sends an IPI with the raw `id` to the specified CPU and interrupt file.
+    pub fn send_ipi_raw(&self, cpu: CpuId, file: ImsicFileId, id: u32) -> Result<()> {
```
dgreid, line 575: Is this limited to IPI interrupts? Or can it inject any imsic interrupt id? Should a type other than u32 be used to indicate that?

src/vm_interrupts.rs
```diff
@@ -25,11 +27,56 @@ enum BindStatus {
...
+        self.bits.iter_mut().for_each(|i| *i = 0u64);
+    }
+}
+
 /// Virtual external interrupt state for a vCPU.
 pub struct VmCpuExtInterrupts {
     bind_status: BindStatus,
     imsic_location: ImsicLocation,
     sw_file: SwFile,
+    allow_list: AllowList,
```
dgreid, line 84: naming nit: allowed_ids? something that indicates what is held as opposed to how it's implemented.

src/vm_interrupts.rs
```diff
@@ -136,4 +184,24 @@ impl VmCpuExtInterrupts {
...
+    pub fn allow_interrupt(&mut self, id: usize) -> Result<()> {
+        self.allow_list.allow_id(id)
+    }
+
+    /// Allows injection of all external interrupts.
+    pub fn allow_all_interrupts(&mut self) {
+        self.allow_list.allow_all();
+    }
+
+    /// Removes `id` from the list of inejctable external interrupts.
```
dgreid, line 203: sp. injectable

dgreid (approved):
awesome, just nits. thanks!

sbi/src/api/tee_guest.rs
```diff
@@ -36,3 +36,21 @@ pub fn add_shared_memory_region(addr: u64, len: u64) -> Result<()> {
     unsafe { ecall_send(&msg) }?;
     Ok(())
 }
+
+/// Allows injection of the specified external interrupt ID by the host to the calling CPU. Passing
+/// -1 enables injection of all external interrupts.
+pub fn allow_external_interrupt(id: i64) -> Result<()> {
```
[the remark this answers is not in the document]
abrestic-rivos: Done.

sbi/src/api/tee_interrupt.rs
```diff
@@ -85,3 +85,16 @@ pub fn unbind_vcpu_imsic_end(tvm_id: u64, vcpu_id: u64) -> Result<()> {
     unsafe { ecall_send(&msg) }?;
     Ok(())
 }
+
+/// Injects an external interrupt into the specified vCPU. The interrupt ID must have been
+/// allowlisted by the guest vCPU.
```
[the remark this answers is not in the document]
abrestic-rivos: Done.

sbi/src/tee_interrupt.rs
```diff
@@ -146,6 +146,23 @@ pub enum TeeInterruptFunction {
         /// a1 = vCPU ID
         vcpu_id: u64,
     },
+    /// Injects an external interrupt with the given `interrupt_id` into the specified vCPU. If the
+    /// vCPU is presently bound to an IMSIC guest interrupt file, the interrupt is immediately
+    /// injected by writing to the interrupt file. If it is not bound, the interrupt is recorded
+    /// in software and will be injected once the vCPU becomes bound. The specified interrupt must
+    /// be valid and have been allowlisted by the vCPU.
```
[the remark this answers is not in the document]
abrestic-rivos: Done.

drivers/src/imsic/core.rs
abrestic-rivos: Nope, removed the `pub`

src/vm_interrupts.rs
abrestic-rivos: Sure, done.

src/vm_interrupts.rs
abrestic-rivos: Done.

drivers/src/imsic/core.rs
abrestic-rivos: A CPU writing to the memory-mapped `seteipnum` is sort of by definition an IPI since it's software running on the CPU explicitly triggering an interrupt. I used `u32` here since MSI writes are always 4 bytes.

dgreid: sgtm

sbi/src/tee_guest.rs
```diff
@@ -24,6 +24,27 @@ pub enum TeeGuestFunction {
...
+    /// Denies injection of the specified external interrupt ID into the calling TVM vCPU. Passing
+    /// an ID of -1 denies injection of all external interrupts.
+    ///
+    /// Returns an error if the specified external interrupt ID is invalid.
+    ///
+    /// a6 = 2
+    DenyExternalInterrupt {
+        /// a0 = interrupt ID
+        id: i64,
+    },
```
atishp04, line 47: Do we need two APIs ? We can just set a policy of DenyAll by default.

abrestic-rivos: Then there would be no way to deny an interrupt after allowing it, making deny -> allow a one way transition. Given what we've been through with shared memory, I'm not sure that's a good strategy :)

Also, while we don't know what this will look like on the Linux side yet, it's seems plausible that we'd be allowing interrupts in the `request_irq()` path. If that's the case, what would we do in `free_irq()`?

atishp04: Do you mean modifying drivers or some irqchip hook up in the request_irq() path ? I am leaning towards doing the latter as modifying drivers is tricky. I will have more details once I start implementing the guest side. I am currently doing the host side stuff.

Looking at the TDX implementation, they seems to leave it to the guest to make sure that it can gracefully handle malicious external interrupts.

"However, TDX module does not allow VMM injecting interrupt vectors in range 0-30 via posted-interrupt mechanism, which drastically reduces the exposed attack surface towards the untrusted VMM. The rest of above interrupts are considered controlled by the host and therefore the guest kernel code that handles them must be audited and fuzzed as any other code that receives malicious host input."

abrestic-rivos: Yeah, in the generic `request_irq()` path ideally. Indeed TDX (and SNP?) do opt out rather than opt in, which seems backwards... Maybe @rsahita has more context.

PR #161, opened 2022-11-10
Tell docs workflow to use "main"

stillson:
Added parameter so that docs workflow builds from main

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -1,10 +1,11 @@
 name: Docs
 
 on:
-  pull_request_target:
-    branches: ["main"]
+  pull_request:
     types:
       - closed
+    branches:
+      - 'main'
   workflow_dispatch:
 
 env:
@@ -17,6 +18,7 @@ permissions:
 
 jobs:
   build:
+    if: github.event.pull_request.merged == true
     environment:
       name: github-pages
       url: ${{ steps.deployment.outputs.page_url }}
@@ -24,6 +26,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v3
+      with:
+        ref: "main"
     - name: Install riscv64 target
       run: rustup target add riscv64gc-unknown-none-elf
     - name: Generate Docs


.github/workflows/docs.yml
```diff
@@ -24,6 +24,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v3
+      with:
+        repository: "main"
```
abrestic-rivos, line 28: Hm, I don't think this is right. I believe it'll look for a repository called "main" not a branch; looks like if you want a branch you need to use `ref:`.

But reading through the github workflow docs a bit, I think we're keying off the wrong event. Seems like we should be using `pull_request` rather than `pull_request_target`, see: https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#running-your-workflow-when-a-pull-request-merges.

.github/workflows/docs.yml
```diff
@@ -17,13 +18,16 @@ permissions:
...
+    if: github.event.pull_request.merged == true
     environment:
       name: github-pages
       url: ${{ steps.deployment.outputs.page_url }}
 
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v3
+      with:
+        ref: "main"
```
abrestic-rivos, line 30: probably not necessary since we only ever merge PRs to main

stillson: True, but here I think it has value to document what's happening. Also, it ensures that the github.io docs are only generated from main.

abrestic-rivos: is it not also redundant with the `branches: main` above?

stillson: yes, it's unneeded. but it's not going to slow anything down. It would be better if there was a single variable for the same place, I'm checking now if yaml can do that, but I feel safer (probably irrationally) that having both in place could prevent issues in the future.

stillson: Anchors don't work. I can remove it, but I would rather leave it in.

stillson: By anchors, i mean the yaml feature that would automatically repeat a value, so main is in this twice.

abrestic-rivos (approved):

PR #162, opened 2022-11-11
Fix hardcoded device

glg-rv:
Current code mapping fixed reset page at 0x10_0000 was wrong, as it did not account the reset device's PTEs.

This patch adds a super simple "ResetDriver", that adds to the memory map the device. Could be changed to a generic device for future (if any) hardcoded device. But for now it's the only "special" platform device we have, so calling it the Reset Driver seems sane.


dgreid (approved):
Thanks!

glg-rv:
Rebased. Map changes were in conflict.

PR #163, opened 2022-11-12
Added debugging to workflow, fixed workflow_dispatch

stillson:
This adds some debugging to the workflow, need this to debug problems

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -8,9 +8,6 @@ on:
       - 'main'
   workflow_dispatch:
 
-env:
-  CARGO_TERM_COLOR: always
-
 permissions:
   contents: read
   pages: write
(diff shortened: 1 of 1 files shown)


.github/workflows/docs.yml
```diff
@@ -8,36 +8,35 @@ on:
       - 'main'
   workflow_dispatch:
 
-env:
-  CARGO_TERM_COLOR: always
```
dgreid, line 12: is this deletion intentional?

Also the easiest way to test this might be to use a branch, changing the ref to enable it on that branch. You could even do this with the main branch in your personal fork github.com/stillson/salus if that's easier to test merge flows etc.

stillson: I've done that, and it works for me on my branch, but it doesn't work on the rivos branch. That's basically how I've been working on it. This should be close to working, but now it dumps a bunch of variables in the log so that I can check out why it's not working. But, on my branch it just works.

stillson: Also, I turned of colors in case the escape strings interfered with something. I will add them back, eventually, if everything else works.

dgreid: If you push to the `main` branch of your fork do you get the same behavior as the working branch or does it fail like it does on `main` of rivosinc?

stillson: i developed it in my main branch, then copy it to a fork for the PR. It won't work on branch that isn't main

dgreid: assuming nothing depends on stillson/main matching rivosinc/main, does pushing to stillson/main work? That's how I test CI changes, and it's worked pretty well lets me commit intentionally broken commits etc. Then force push it back to rivosinc/main when I'm done

stillson: yes, it does

dgreid: Got it, that maybe means we've got something misconfigured on the rivosinc repo.

Can you eliminate the whitespace changes so it's more clear what's added here and can be reverted later? Then I'm good merging this. Thanks!

Issue #153, opened 2022-11-03
Dynamic shared memory conversion


abrestic-rivos:
Our current approach of having the guest statically declare shared memory regions at start up does not align well with the direction taken by other CC implementations, which hook into the `set_encrypted_memory()` / `set_decrypted_memory()` kernel APIs to convert ranges of caller-chosen address space. To avoid having to hack up swiotlb just for our implementation, we should probably add support for dynamic shared <-> confidential memory conversion like everyone else.

Proposal:

Add two new TG calls, `tg_share_mem_range()` and `tg_unshare_mem_range()` which take an address + length pair and perform the requested conversion by:
 - verifying the requested memory range is a subset of an existing shared/confidential memory region
 - invalidating the pages mapped in the region (if any)
 - blocking the calling vCPU until the host has completed a TVM fence operation
Upon completion, the host may insert pages of the requested type in the region and the vCPU becomes runnable.

We can reuse the page unassignment flow that was added for IMSIC unbind to do the invalidation. The only tricky thing here is the blocking of the vCPU until the invalidation is complete. We could do this by marking the vCPU as "not runnable until TLB version is > N".

PR #166, opened 2022-11-15
Basic dynamic memory sharing support

abrestic-rivos:
Add support for runtime, guest-initiated conversion between shared <-> confidential memory when the region to be converted is unpopulated.

This adds two new TEE-Guest calls -- `share_memory()` and `unshare_memory()` -- which initiate the confidential -> shared and shared -> confidential conversion, respectively. For now, this consists of updating the `VmRegionList` with the new region type and verifying that the range is unpopulated. We'll add support for handling populated regions in a follow-on PR.

Patch 1 fixes an issue that prevent sharing of already-shared pages.
Patch 2 defines the new TEE-Guest calls.
Patches 3-5 are prep work for handling the runtime shared <-> confidential conversion.
Patch 6 implements the new TEE-Guest calls.
Patch 7 updates Tellus/GuestVM to use the new calls.
Patches 8&9 remove now-obsolete TEE calls.

This should be enough to enable `swiotlb` in Linux TVM guests without restoring to pre-declared shared memory pools.

Partially addresses #153; still need to handle the already-populated case.

### DIFF
--- a/src/vm_pages.rs
+++ b/src/vm_pages.rs
@@ -989,12 +989,6 @@ pub type AnyVmPages<'a, T> = VmPagesRef<'a, T, VmStateAny>;
 pub type FinalizedVmPages<'a, T> = VmPagesRef<'a, T, VmStateFinalized>;
 
 impl<'a, T: GuestStagePagingMode> FinalizedVmPages<'a, T> {
-    /// Adds a shared memory region of `len` bytes starting at `page_addr` to this VM's address
-    /// space.
-    pub fn add_shared_memory_region(&self, page_addr: GuestPageAddr, len: u64) -> Result<()> {
-        self.do_add_region(page_addr, len, VmRegionType::Shared)
-    }
-
     /// Adds an emulated MMIO region of `len` bytes starting at `page_addr` to this VM's address
     /// space.
     pub fn add_mmio_region(&self, page_addr: GuestPageAddr, len: u64) -> Result<()> {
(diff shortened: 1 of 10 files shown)


src/vm_pages.rs
```diff
@@ -341,6 +337,7 @@ pub struct VmPagesMapper<'a, T: GuestStagePagingMode, M> {
     vm_pages: &'a VmPages<T>,
     mapper: GuestStageMapper<'a, T>,
     _mapper_type: PhantomData<M>,
```
glg-rv, line 445: IMHO would look slightly better to move phantom data at the end of the struct.

src/vm_pages.rs
```diff
@@ -368,6 +366,7 @@ impl<'a, T: GuestStagePagingMode, M> VmPagesMapper<'a, T, M> {
             vm_pages,
             mapper,
             _mapper_type: PhantomData,
```
glg-rv, line 474: (as above)

src/vm_pages.rs
```diff
@@ -331,6 +402,41 @@ impl VmRegionList {
     }
 }
 
+// A reference to a `VmRegion` that is in the process of being updated. The referenced region
+// is converted back into its previous region type if this object is dropped before `finish()`
+// is called.
```
glg-rv, line 407: nit: is "Reserved" a bit too generic as a name? It _is_ reserved as it is not accessible, but it doesn't describe really the state is in. Naming is hard, so won't insist on this.

abrestic-rivos: Yeah, fair. I renamed this to `VmRegionUpdater`, and changed the temporary region name to `Updating`. WDYT?

atishp04:
LGTM.

PR #167, opened 2022-11-17
Make PageTable interface more generic

abrestic-rivos:
To prepare for dynamic (un)sharing of populated regions of a guest's address space, make the `PageTable` APIs more generic and remove (most of) their dependency on `PageTracker` states. This puts the burden of coordinating page states on the caller, and avoids having to expand the `PageTable` API for the slight variations of page invalidation / unassignment needed for dynamic (un)sharing.

The general theme is that the `PageTable` APIs take a virtual address range, returning an iterator that performs the requested action only if a predicate function returns true for the entire address range. The `PageTable` lock is held within the iterator, allowing the caller to still execute page state transitions atomically. This also allows us to handle concurrent page sharing properly.

- Patches 1&2 make `invalidate_range()` and `unassign_range()` more generic, respectively.
- Patch 3 introduces `get_mapped_pages()` and uses it to implement page sharing.
- Patch 4 removes the need to exclusively lock pages in `PageTracker` when sharing them.
- Patch 5 introduces `get_invalidated_pages()` and uses it to implement assignment of converted pages.


glg-rv (commented):
The last commit message should say "iterate" rather than iterator.

riscv-page-tables/src/page_table.rs
```diff
@@ -1098,3 +1064,37 @@ impl<'a, T: PagingMode> Drop for GuestStageMapper<'a, T> {
...
+}
+
+impl<'a, T: PagingMode> Iterator for InvalidateIter<'a, T> {
+    type Item = SupervisorPageAddr;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        if self.vaddr == self.end_vaddr {
+            return None;
+        }
+        let pte = self.inner.get_mapped_4k_leaf(self.vaddr).ok()?.invalidate();
```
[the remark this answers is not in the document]
abrestic-rivos: This can't happen; we already checked the entire range can be invalidated and we've held an exclusive lock the whole time. Changed it to `unwrap()` with a comment.

abrestic-rivos:
I went ahead and updated this with the rest of the dynamic page sharing support, PTAL.

riscv-page-tables/src/page_table.rs
```diff
@@ -1103,3 +1110,28 @@ impl<'a, T: PagingMode> Iterator for InvalidateIter<'a, T> {
...
+type UnmapIter<'a, T> = PageTableIter<'a, T, UnmapPages>;
+
+impl<'a, T: PagingMode> Iterator for UnmapIter<'a, T> {
+    type Item = SupervisorPageAddr;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        while self.vaddr != self.end_vaddr {
+            let pte = self.inner.get_invalidated_4k_leaf(self.vaddr);
+            // Unwrap ok: We checked for overflow prior to construction.
+            self.vaddr = self.vaddr.checked_add_pages(1).unwrap();
```
dgreid, line 1134: Should there be a check in `new` that `end_vaddr` >= `vaddr`?

dgreid: Actually are these iterator types useful? I see that the Unmap vs Invalidate vs Map type safety is nice, but nothing else is using those type explicitly. What if we embraced iterator chaining and avoided the unwraps all together? Here is a patch, `run_tellus` tested only that might work and could be used for the other iters too.
The types are still different so no safety lost there, they are just implicit, unreadable types combining an iterator and a closure:)
```
diff --git a/riscv-page-tables/src/page_table.rs b/riscv-page-tables/src/page_table.rs
index 86e57c8..33ea268 100644
--- a/riscv-page-tables/src/page_table.rs
+++ b/riscv-page-tables/src/page_table.rs
@@ -889,17 +889,17 @@ impl<T: PagingMode> GuestStagePageTable<T> {
     /// Verifies the entire virtual address range is invalidated (or unpopulated) and that `pred`
     /// returns true for each invalidated page, returning an iterator that yields the pages after
     /// clearing their PTEs.
-    pub fn unmap_range<F>(
-        &self,
+    pub fn unmap_range<'a, F>(
+        &'a self,
         vaddr: PageAddr<T::MappedAddressSpace>,
         len: u64,
         mut pred: F,
-    ) -> Result<UnmapIter<T>>
+    ) -> Result<impl Iterator<Item = SupervisorPageAddr> + 'a>
     where
         F: FnMut(SupervisorPageAddr) -> bool,
     {
         let num_pages = PageSize::num_4k_pages(len);
-        let end = vaddr
+        vaddr
             .checked_add_pages(num_pages)
             .ok_or(Error::AddressOverflow)?;
         let mut inner = self.inner.lock();
@@ -924,7 +924,18 @@ impl<T: PagingMode> GuestStagePageTable<T> {
                 }
             }
         }
-        Ok(UnmapIter::new(inner, vaddr, end))
+
+        Ok(vaddr
+            .iter_from()
+            .take(num_pages as usize)
+            // Skip over non-invalidated PTEs -- we verified there were no mapped or locked PTEs
+            // prior to construction.
+            .filter_map(move |va| {
+                let pte = inner.get_invalidated_4k_leaf(va).ok()?;
+                let paddr = pte.page_addr();
+                pte.clear();
+                Some(paddr)
+            }))
     }

     /// Verifies the entire virtual address range is mapped and that `pred` returns true for
@@ -1119,31 +1130,6 @@ impl<'a, T: PagingMode> Iterator for InvalidateIter<'a, T> {
     }
 }

-/// An iterator that clears invalidated PTEs as it yields the mapped physical addresses.
-pub enum UnmapPages {}
-
-type UnmapIter<'a, T> = PageTableIter<'a, T, UnmapPages>;
-
-impl<'a, T: PagingMode> Iterator for UnmapIter<'a, T> {
-    type Item = SupervisorPageAddr;
-
-    fn next(&mut self) -> Option<Self::Item> {
-        while self.vaddr != self.end_vaddr {
-            let pte = self.inner.get_invalidated_4k_leaf(self.vaddr);
-            // Unwrap ok: We checked for overflow prior to construction.
-            self.vaddr = self.vaddr.checked_add_pages(1).unwrap();
-            // Skip over non-invalidated PTEs -- we verified there were no mapped or locked PTEs
-            // prior to construction.
-            if let Ok(p) = pte {
-                let paddr = p.page_addr();
-                p.clear();
-                return Some(paddr);
-            }
-        }
-        None
-    }
-}
-
 /// An iterator that yields the mapped physical addresses for a populated virtual address range.
 pub enum MappedPages {}
```

dgreid (commented):
just got through the iterators will try to read the rest of the commits ASAP.

dgreid (approved):
Thanks, this looks a lot less painful than I worried it would!

atulkharerivos (approved):

riscv-page-tables/src/lib.rs
```diff
@@ -152,10 +152,16 @@ mod tests {
             assert!(mapper.map_page(gpa, mappable).is_ok());
         }
         let version = TlbVersion::new();
-        guest_page_table
-            .invalidate_range::<Page<Invalidated>>(gpa_base, PageSize::Size4k, 2)
-            .unwrap()
-            .for_each(|invalidated| page_tracker.convert_page(invalidated, version).unwrap());
+        let invalidated = guest_page_table
```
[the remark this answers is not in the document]
abrestic-rivos: Done.

riscv-page-tables/src/page_table.rs
abrestic-rivos: Ah good point -- the `filter_map()` makes things cleaner when skipping over PTEs too. I've taken this change and updated the other iterator-returning functions accordingly.

PR #171, opened 2022-11-29
Zero the BSS

abrestic-rivos:
Turns out we weren't doing this. Whoops.

Not an issue currently since we're running under QEMU and only OpenSBI is running before us, but it's still the right thing to do.

Patch 1 is trivial fix to the trap handler, patches 2&3 handle the BSS zeroing.

### DIFF
--- a/src/start.S
+++ b/src/start.S
@@ -14,18 +14,20 @@ _start:
 .option norelax
     la gp, _global_pointer
 .option pop
-    la sp, _stack_end
     csrw sstatus, zero
     csrw sie, zero
 
-    la t1, kernel_init
-    la ra, 1f
-    jr t1
-
-    j 1f
+    // Clear the BSS
+    la   a3, _bss_start
+    la   a4, _bss_end
 1:
-    wfi
-    j 1b
+    sd   zero, (a3)
+    addi a3, a3, 8
+    blt  a3, a4, 1b
+
+    la   sp, _stack_end
+    call kernel_init
+    j    wfi_loop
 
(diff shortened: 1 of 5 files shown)


glg-rv (approved):
New to RISCV-64 asm and had to read way too many PDFs to review this, but LGTM.

src/start.S
```diff
@@ -14,18 +14,22 @@ _start:
...
+    sd   zero, (a3)
+    addi a3, a3, 8
+    blt  a3, a4, 1b
+
+    la   sp, _stack_end
+    call kernel_init
+2:
     wfi
-    j 1b
+    j 2b
```
rajnesh-kanwal, line 32: 2b on line 49?

>     la t1, secondary_init
>     la ra, 1b
>     jr t1

test-workloads/src/start.S
```diff
@@ -14,18 +14,22 @@ _start:
...
+
+    la   sp, _stack_end
+    call kernel_init
+2:
     wfi
-    j 1b
+    j 2b
 
 // The entry point for secondary CPUs.
 .global _secondary_start
```
rajnesh-kanwal, line 35: Same as above.
2b on line 49?

rajnesh-kanwal (approved):
Minor comment otherwise LGTM.

src/start.S
abrestic-rivos: Whoops, good catch! Ended up just giving the wfi loop it's own label to make this more clear.

PR #172, opened 2022-11-29
Per Cpu Page Tables

glg-rv:
Three things going on in this PR (see commits).

1. Rename `VmState` to `State`, and allow to allocate state page for both host vm and hypervisor. Could create a different state for hypervisor, but wanted to go for the simpler option first.
2. Use the new ability to allocate hypervisor internal state to allocate page tables via the page allocator rather than by reserving hardware memory map. This allows to separate a bit better that fragile moment between turning paging on and starting using the allocator. Fixes an inconsistency that was due to the fragility of allocator and paging initialization.
3. Create per-cpu pagetables by copying the top level pagetable.

With this patch, the hypervisor page tables will be shared, and only the top level page table will be per-cpu.
This is much better than it sounds, as even when we'll have user space mapping we will go towards a static and not modifiable hypervisor page tables for all VCPUs, so we will not worry about pte reference counting.


page-tracking/src/page_tracker.rs
```diff
@@ -606,24 +606,42 @@ impl HypPageAlloc {
...
         count: usize,
         align: u64,
     ) -> SequentialPages<InternalClean> {
         let assignable_pages = self.take_pages(count, align);
         SequentialPages::from_pages(assignable_pages.into_iter().map(|p| {
             let page_info = self.pages.get_mut(p.addr()).unwrap();
-            page_info
-                .assign(PageOwnerId::host(), PageState::VmState)
-                .unwrap();
+            page_info.assign(owner, PageState::State).unwrap();
```
abrestic-rivos, line 622: The pages returned by `take_pages()` are already hypervisor-owned, so in the `take_pages_for_hyp_state()` case you end up pushing the hypervisor as the owner again. What I think you want instead is to make `take_pages()` more generic w.r.t. the state it sets for the newly-claimed pages.

riscv-page-tables/src/page_table.rs
```diff
@@ -649,6 +659,16 @@ impl<T: FirstStagePagingMode> FirstStagePageTable<T> {
         })
     }
 
+    /// Create a new page table root from the privided `root`, with the same constraint as
+    /// `FirstStagePageTable::new()`. Copy the content of the root page table from `src`.
+    ///
+    /// Note: The PTEs will be shared between `src` and the new page table created.
```
abrestic-rivos, line 665: This seems incredibly unsafe to me. You're basically requiring that any existing non-root PTs and PTEs remain static after this point, otherwise there's unguarded shared mutability. I get that you plan on using this in a safe way -- by putting the per-CPU mappings far away from everything else in VA space -- but this still seems like a massive foot-gun.

Unless you want to go down the route of somehow enforcing that this shared range of VA space doesn't get modified (e.g. PTE locks, or "sealing" mappings), I'd just make a deep copy like in your previous iteration for now.

page-tracking/src/page_info.rs
```diff
@@ -34,8 +34,8 @@ pub enum PageState {
     Shared(u64),
 
     /// Page is used to store hypervisor-internal state for the current owner, e.g. to back per-VM
-    /// data structures or as a page-table page for the VM.
-    VmState,
+    /// data structures or as a page-table page for the VM or hypervisor.
+    State,
```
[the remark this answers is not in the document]
glg-rv: Yes, thank you. I was essentially fishing for names. :-)

page-tracking/src/page_tracker.rs
glg-rv: Okay, makes sense.

page-tracking/src/page_tracker.rs
```diff
@@ -606,24 +606,42 @@ impl HypPageAlloc {
         dirty_pages.clean()
     }
 
-    /// Same as above, but the returned pages are assigned for internal use.
-    pub fn take_pages_for_host_state_with_alignment(
+    // Same as `take_pages`, but returned page are state pages for either hypervisor or host.
+    //
+    // Safety:
+    // `owner` must be either `PageOwnerId::host()` or `PageOwnerId::hypervisor()`.
```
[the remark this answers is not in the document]
glg-rv: Yes, I just wanted to understand our use of unsafe. Will change to a requirement of the call. But actually most likely this particular function will go away.

riscv-page-tables/src/page_table.rs
glg-rv: Yes, I'd like to go towards enforcing this. In general there's an advantage to be had in having page table extremely static, set once. I think that mapping page tables read-only once set is a decent way to go.
I don't think salus has any advantage in having dynamic paging.

glg-rv: Just to get back at this discussion, I think a better solution here is to create some form of type safety (something similar to `VmPages`, with an `InitializingVmPages` state that allows wider changes, and a state of `FinalizedVm` that prevents that). Once this is done, it becomes irrelevant whether we have a a full copy of page tables or a shared one.

page-tracking/src/page_info.rs
atulkharerivos, line 38: Unrelated, but typo (onwner) in commit message.

src/main.rs
```diff
@@ -509,11 +495,19 @@ extern "C" fn kernel_init(hart_id: u64, fdt_addr: u64) {
...
-
     // Set up per-CPU memory and boot the secondary CPUs.
     PerCpu::init(hart_id, &mut mem_map);
 
+    // Create an allocator for the remaining pages. Anything that's left over will be mapped
+    // into the host VM.
+    let mut hyp_mem = HypPageAlloc::new(&mut mem_map);
+
+    // We are about to set up paging based on the hardware memory map. It can't be modified from now on.
+    let mem_map = mem_map; // Remove mutability.
```
atulkharerivos, line 526: Nit: The "lousily" in the commit message can be dropped given that this is an accepted Rust idiomatic practice to convert a mutable reference to mutable, and the enforcement is assured by the compiler.

page-tracking/src/page_info.rs
glg-rv: Thanks!

src/main.rs
glg-rv: Okay. Still the reality of a computer doesn't stop at the compiler. I consider it lousy because something somewhere can still write to it.

src/main.rs
```diff
@@ -266,28 +258,22 @@ fn hyp_map_region(
...
-            PageSize::Size4k,
-            num_pte_pages,
-        )
-        .unwrap()
-        .clean()
-        .into_iter();
+fn setup_hyp_paging(mem_map: &HwMemMap, hyp_mem: &mut HypPageAlloc) {
     // Create empty sv48 page table
-    let root_page = pte_pages.next().unwrap();
+    // Unwrap okay: we expect to have at least one page free or not much will happen anyway.
```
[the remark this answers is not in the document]
glg-rv: Heh. Okay! Will remove the personal touch! :-)

riscv-page-tables/src/page_table.rs
glg-rv: Created HypMap in new version. Also switched to full copies to avoid confusing rust's model.

src/main.rs
glg-rv: Done. Not lousily.

src/main.rs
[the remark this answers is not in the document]
glg-rv: Done.

page-tracking/src/page_tracker.rs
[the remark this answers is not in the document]
glg-rv: Removed

page-tracking/src/page_info.rs
glg-rv: Should be fxied

src/hyp_map.rs
```diff
@@ -0,0 +1,111 @@
...
+
+// Adds an identity mapping to the given Sv48 table for the specified address range.
+fn hyp_map_region(
+    sv48: &FirstStagePageTable<Sv48>,
+    base: PageAddr<SupervisorPhys>,
+    size: u64,
+    perms: PteLeafPerms,
+    get_pte_page: &mut dyn FnMut() -> Option<Page<InternalClean>>,
+) {
+    let region_page_count = PageSize::Size4k.round_up(size) / PageSize::Size4k as u64;
```
abrestic-rivos, line 45: nit: `PageSize::num_4k_pages()`

src/hyp_map.rs
```diff
@@ -0,0 +1,111 @@
...
+use riscv_pages::{InternalClean, Page, PageAddr, PageSize, RawAddr, SupervisorPhys};
+use riscv_regs::{satp, LocalRegisterCopy, SatpHelpers};
+
+// Returns the base, size, and permission pair for the given region if that region type should be
+// mapped in the hypervisor's virtual address space.
+fn hyp_map_params(r: &HwMemRegion) -> Option<(PageAddr<SupervisorPhys>, u64, PteLeafPerms)> {
+    match r.region_type() {
+        HwMemRegionType::Available => {
+            // map available memory as rwx - unser what it'll be used for.
+            Some((r.base(), r.size(), PteLeafPerms::RWX))
```
abrestic-rivos, line 16: (Not related to your patch and no action requested, but we need to fix these permissions. Unused, random memory should _not_ have X perms.)

src/main.rs
```diff
@@ -204,12 +201,15 @@ fn find_available_region(mem_map: &HwMemMap, size: u64) -> Option<SupervisorPage
...
-    // Store the SATP value for other CPUs. They load from the global in start_secondary.
-    SATP_VAL.call_once(|| satp);
-    // Install the page table in satp
-    CSR.satp.set(satp);
+    // Create per-cpu page tables.
+    let cpu_info = CpuInfo::get();
+    for i in 0..cpu_info.num_cpus() {
+        let page_table = hyp_map.new_page_table(hyp_mem);
+        PerCpu::set_cpu_page_table(CpuId::new(i), page_table);
+    }
```
abrestic-rivos, line 209: Instead of setting up all the secondary CPU PTs on the boot CPU, could we stick `hyp_map` in a global and let each secondary CPU set up its own PT? That would (theoretically) allow us to set up PTs in parallel and avoids having to reach into the `PerCpu` struct from the boot CPU (obviously fine here, but a little ugly).

abrestic-rivos (commented):
Looking good, thanks!

src/main.rs
glg-rv: I could. Could be ugly in other forms. `hyp_mem` must also be stored somewhere. `hyp_mem` is currently consumed by the host vm loader.
Now in the current code we wait for the host_vm to be finalized, so we could just stick the `hyp_mem` in a global, and pass it only as a mutable reference. And impose that no one uses it after. I think that letting the loader consume it is much better than this scenario.
Alternatively we could re-introduce `pte_page_count`, _over_-allocate the pages on BSP for all cpus in a sequential page and using that -- in a thread-safe way. Which means that `new_page_table` will ascept a SequentialPages.

I am open to what you think it's best, I just think that this current code is the slighlty better version.

Please also note that I would like to keep the cpu_page_table field to map per-cpu mapping in the future.

src/hyp_map.rs
glg-rv: Yeah, having hypervisor image as RWX is a bit scary too. I need to look up at how salus is loaded.

src/hyp_map.rs
glg-rv: ack.

src/main.rs
abrestic-rivos: > Now in the current code we wait for the host_vm to be finalized, so we could just stick the hyp_mem in a global, and pass it only as a mutable reference. And impose that no one uses it after. I think that letting the loader consume it is much better than this scenario.

Yeah ok, that's a fair point. What you have now SGTM then.

glg-rv:
I pushed the wrong branch by mistake. Please use this link to see the difference between this version and the previously reviewed.

https://github.com/rivosinc/salus/compare/de62e7c6fee13e079b4f3d3fa265a31790ef13d0..e32f364bf17beddeffbad0fbf375627ee59f7c40

src/hyp_map.rs

page-tracking/src/page_tracker.rs
```diff
@@ -553,12 +554,17 @@ impl HypPageAlloc {
...
+    // system map. Sets the hypervisor as the owner and `page_state` as the state, for the returned
+    // pages and for any pages consumed for alignment purposes.
+    // Returns a pair of supervisor addresses containing the first and the last page address of the
+    // allocated range.
+    fn alloc_pages(
+        &mut self,
+        count: usize,
+        align: u64,
+        page_state: PageState,
+    ) -> (SupervisorPageAddr, SupervisorPageAddr) {
```
abrestic-rivos, line 567: Use `SupervisorPageRange` instead of a pair of addresses?

abrestic-rivos (approved):
LGTM, thanks!

page-tracking/src/page_tracker.rs
glg-rv: Oh, much better. Yes!

src/hyp_map.rs
glg-rv: I think @rajnesh-kanwal is doing this as part of his changes.

page-tracking/src/page_tracker.rs
glg-rv: Updated. I thought about using `SupervisorPageRange` for the assignment iteration as well, but the iterator is destructive and I would've had to create two.

PR #177, opened 2022-12-07
HypMap: Fixes and cleanup for hypervisor mappings.

glg-rv:
In preparation for user mappings, create `HypMapRegion, a structure that abstracts mapping in kernels.

At the moment, only Hw Memory Map regions can be consumed to create a hyp map region.

Also, fix the pte permissions for hardware regions. Although more work needs to be done for the hypervisor image, host kernel and initrd are mapped read-only, all others, except for hypervisor image, read/write. 

Also, as PTEs are now allocated by the page allocator, the definition of `HypervisorPtes` is removed.


src/hyp_map.rs
```diff
@@ -2,63 +2,96 @@
...
-            // No need to map regions reserved for firmware use
+/// Maximum number of regions that will be mapped in the hypervisor.
+const MAX_HYPMAP_REGIONS: usize = 32;
+
+/// Represents a region of the virtual address space of the hypervisor.
+pub struct HypMapRegion {
+    vaddr: PageAddr<SupervisorVirt>,
+    paddr: PageAddr<SupervisorPhys>,
+    page_size: PageSize,
+    page_count: usize,
```
abrestic-rivos, line 21: So I'm not sure encoding the page size in the region is a good idea -- consider the possible ways to represent a region that is, for example, 5MB and not necessarily 2MB aligned. IMO the easiest way to go about mapping a contiguous region with huge pages would be for `map()` below to pick the appropriate page size based on the alignment and remaining bytes in the region.

page-tracking/src/hw_mem_map.rs
```diff
@@ -88,9 +88,6 @@ pub enum HwReservedMemType {
     /// The hypervisor per-CPU memory area.
     HypervisorPerCpu,
 
-    /// Sv48 page tables for the hyeprvisor running in HS mode.
-    HypervisorPtes,
-
```
abrestic-rivos, line 93: Removal of `HypervisorPtes` should probably be in its own commit.

src/hyp_map.rs
glg-rv: My expectation was that PageAddr would be aligned to the particular size. So that rather than pushing the decision to use big pages in some inner code would've gone a bit more at higher level.
But can change this.

src/hyp_map.rs
```diff
@@ -2,63 +2,96 @@
...
-        .iter_from()
-        .zip(base.iter_from())
-        .take(region_page_count as usize)
-    {
-        // Safe as we will create exactly one mapping to each page and will switch to
-        // using that mapping exclusively.
-        unsafe {
-            mapper.map_addr(virt, phys, pte_fields).unwrap();
+    /// Map this region into a page table.
+    pub fn map(
```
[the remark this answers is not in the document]
glg-rv: No

page-tracking/src/hw_mem_map.rs
glg-rv: Ack.

src/hyp_map.rs
abrestic-rivos: So the issue with that is, each region you add to the list could actually end up being 3 (or more!) separate regions. Since the list is a small-ish fixed size -- and its perfectly possible to do the page-size fitting in the inner loop -- it seems better to avoid dividing regions up by page size in the regions list.

src/hyp_map.rs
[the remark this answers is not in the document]

src/hyp_map.rs
glg-rv: Yes I understand. Actually makes the whole thing less architectural so it's probably a good thing.

page-tracking/src/hw_mem_map.rs
glg-rv: Done.

abrestic-rivos (approved):

PR #178, opened 2022-12-09
Introduce umode binary and map it in salus user space address.

glg-rv:
This PR includes a `umode` and `libuser` library (taken from Dylan's branch). These are just stubs at the moment, but they do create a risc-v elf binary that can be used as a basis for future userspace work.

It also introduces riscv-elf. A super simple ELF library tailored to our needs, with no memory allocation and focused on loading elf. Care has been taken in checking the validity of file offset before accessing them.

It then proceed to extend `HypMap` to allow a new type of region to be loaded. Specifically, it's a region that needs to be allocated and populated with existing data.

With this patch, the ELF is loaded in userspace, and there's a copy of the whole elf for each CPU. Current abstraction can be changed though to share `.text` and `.rodata`  across CPUs. Although the big batch of memory (`.data` and `.stack`) will always be per-cpu.

### DIFF
--- a/lds/umode.lds
+++ b/lds/umode.lds
@@ -1,0 +1,47 @@
+OUTPUT_ARCH( "riscv" )
+
+ENTRY( _start )
+
+PHDRS
+{
+    text PT_LOAD;
+    rodata PT_LOAD;
+    data PT_LOAD;
+    stack PT_LOAD;
+}
+
+SECTIONS
+{
+    . = 0xffffffff00000000;
+
+    .text ALIGN(4096) : {
+        *(.text.start)
+        *(.text.init) *(.text .text.*)
+    } :text
+
+    .rodata ALIGN(4096) : {
+        *(.rodata .rodata.*)
+    } :rodata
+
+    .data ALIGN(4096) : {
+        *(.data .data.*)
+
+        . = ALIGN(8);
+        PROVIDE(__global_pointer$ = .);
+        *(.sdata .sdata.*)
+
+        *(.sbss .sbss.*) *(.bss .bss.*)
+    } :data
+
+    . += 4096;
+
+    .stack ALIGN(4096) (NOLOAD) : {
+    PROVIDE(_stack_start = .);
+    . += 4096;
+    PROVIDE(_stack_end = .);
+    } :stack
+
+    /DISCARD/ : {
+        *(.eh_frame)
+    }
+}
--- a/libuser/src/task_start.S
+++ b/libuser/src/task_start.S
@@ -1,0 +1,23 @@
+// Copyright (c) 2021 by Rivos Inc.
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
+
+.option norvc
+
+.section .text.start
+
+// The entry point for a task.
+.global _start
+_start:
+
+.option push
+.option norelax
+    la gp, __global_pointer$
+.option pop
+    la sp, _stack_end
+
+    call task_main
+
+    // ecall to exit
+    ecall
+
--- a/riscv-elf/src/lib.rs
+++ b/riscv-elf/src/lib.rs
@@ -1,0 +1,570 @@
+// Copyright (c) 2022 by Rivos Inc.
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
+
+#![no_std]
+
+//! RiscV ELF loader library for salus
+
+// For testing use the std crate.
+#[cfg(test)]
+#[macro_use]
+extern crate std;
+
+use arrayvec::ArrayVec;
+use core::{fmt, result};
+
+// Maximum size of Program Headers supported by the loader.
+const ELF_SEGMENTS_MAX: usize = 8;
+
+/// Elf Offset Helper
+///
+/// An Elf Offset. A separate type to be sure to never used it
+/// directly, but only through `slice_*` functions.
+#[repr(packed, C)]
+#[derive(Copy, Clone)]
+pub struct ElfOffset64 {
+    inner: u64,
+}
+
+impl ElfOffset64 {
+    fn as_usize(&self) -> usize {
+        // We're 64-bit. u64 fits in a usize.
+        self.inner as usize
+    }
+
+    fn usize_add(self, other: usize) -> Option<ElfOffset64> {
+        let inner = self.inner.checked_add(other as u64)?;
+        Some(Self { inner })
+    }
+}
+
+impl From<usize> for ElfOffset64 {
+    fn from(val: usize) -> Self {
+        Self { inner: val as u64 }
+    }
+}
+
+fn slice_check_offset(bytes: &[u8], offset: ElfOffset64) -> bool {
+    bytes.len() > offset.as_usize()
+}
+
+fn slice_check_range(bytes: &[u8], offset: ElfOffset64, size: usize) -> bool {
+    if size < 1 {
+        return false;
+    }
+
+    if let Some(last) = offset.usize_add(size - 1) {
+        slice_check_offset(bytes, last)
+    } else {
+        false
+    }
+}
+
+fn slice_get_range(bytes: &[u8], offset: ElfOffset64, len: usize) -> Option<&[u8]> {
+    if slice_check_range(bytes, offset, len) {
+        let start = offset.as_usize();
+        Some(&bytes[start..start + len])
+    } else {
+        None
+    }
+}
+
+/// ELF64 Program Header Table Entry
+#[repr(packed, C)]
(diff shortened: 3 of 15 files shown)


src/hyp_map.rs
```diff
@@ -109,17 +200,27 @@ impl HypPageTable {
...
-    regions: ArrayVec<HypMapRegion, MAX_HYPMAP_REGIONS>,
+    supervisor_regions: ArrayVec<HypMapFixedRegion, MAX_HYPMAP_SUPERVISOR_REGIONS>,
+    user_regions: ArrayVec<HypMapPopulatedRegion, MAX_HYPMAP_USER_REGIONS>,
 }
 
 impl HypMap {
     /// Create a new hypervisor map from a hardware memory mem map.
-    pub fn new(mem_map: HwMemMap) -> HypMap {
-        let regions = mem_map
+    pub fn new(mem_map: HwMemMap, elf_map: ElfMap) -> HypMap {
```
abrestic-rivos, line 209: s/elf_map/user_map (or umode_map)?

src/hyp_map.rs
```diff
@@ -93,6 +97,93 @@ impl HypMapRegion {
...
+/// Represents a virtual address region that must be allocated and populated to be mapped.
+struct HypMapPopulatedRegion {
+    // The address space where this region starts.
+    vaddr: PageAddr<SupervisorVirt>,
+    // Number of bytes of the VA area
+    size: u64,
+    // PTE bits for the mappings.
+    pte_fields: PteFieldBits,
+    // Data to be populated in the VA area
+    data: Vec<u8>,
```
abrestic-rivos, line 109: Can we avoid this extra `memcpy()` from the ELF to this temporary buffer?

src/main.rs
```diff
@@ -459,8 +449,34 @@ extern "C" fn kernel_init(hart_id: u64, fdt_addr: u64) {
     let guest_phys_size = mem_map.regions().last().unwrap().end().bits()
         - mem_map.regions().next().unwrap().base().bits();
 
+    // Parse the user-mode ELF containing the user-mode task.
+    let user_elf = include_bytes!("../target/riscv64gc-unknown-none-elf/release/umode");
```
abrestic-rivos, line 453: Ugh, this is gross but I don't immediately have a better suggestion.

src/hyp_map.rs
glg-rv: You're describing my Thursday night. Originally I had kept the reference from `'elf` to here. It is sane because afterall the elf is `'static`. But turns out that at least I couldn't understand how to let Rust understand in a non convoluted way that there were to lifetimes here, one of the underlying data structure and the other of the `user_map`.

In general the structure needs to have a life time, the code becomes much much harder to follow and should essentially hold the ElfMap in it to keep the reference lifetime valid. Just over complication for a language that clearly doesn't like these sort of things.

This would also remove generality to the whole concept of a populated region.

In general, much more complicated code, to save a few kbs of RAM.

src/hyp_map.rs
glg-rv: if I have to chose (and not chose `elf_map`, i'll go for `user_map`. :-)

riscv-elf/src/lib.rs
```diff
@@ -0,0 +1,379 @@
...
+#[repr(packed, C)]
+#[derive(Copy, Clone)]
+pub struct ElfOffset64 {
+    inner: u64,
+}
+
+impl ElfOffset64 {
+    fn as_usize(&self) -> Option<usize> {
+        self.inner.try_into().ok()
+    }
```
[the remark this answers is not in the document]
glg-rv: Sure. I went for paranoia in this code.

riscv-elf/src/lib.rs
```diff
@@ -0,0 +1,379 @@
...
+
+    /// Return an iterator containings loadable segments of this ELF file.
+    pub fn segments(&'elf self) -> impl Iterator<Item = &'elf ElfSegment> {
+        self.segments.iter()
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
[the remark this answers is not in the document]
glg-rv: I actually have a fuzzing infrastructure for this (super simple). But yeah, I can just take a random thing just to show some basic things in case someone changes these in the future.

lds/umode.lds
```diff
@@ -0,0 +1,49 @@
...
+
+        *(.sdata .sdata.*)
+
+        *(.sbss .sbss.*) *(.bss .bss.*)
+        PROVIDE(_bss_end = .);
+    } >ram AT>ram :data
+
+    .stack : ALIGN(4096) {
+        PROVIDE(_stack_start = .);
+        . += 4096;
```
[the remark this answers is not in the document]
glg-rv: No but can be changed quite easily. This is just for the stub.

lds/umode.lds
```diff
@@ -0,0 +1,49 @@
+OUTPUT_ARCH( "riscv" )
+
+ENTRY( _start )
+
+MEMORY
+{
+    ram  (wxa) : ORIGIN = 0xffffffff00000000, LENGTH = 128M
```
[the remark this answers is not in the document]
glg-rv: Yes, I agree. Just need to look out to change the assumpion in salus that VA = PA. Which is a hard one to get rid of.

src/main.rs
glg-rv: Yes, it is indefensible but also the best thing we can do now. Double-checked with @dgreid about this as well.

src/hyp_map.rs
abrestic-rivos: I think you had your lifetimes backwards in `ElfMap::segments()`. With that fixed, this becomes pretty straightforward, see https://github.com/abrestic-rivos/salus/tree/for-glg/user_mode

glg-rv: Whops. That's a few hours of my life I'll never get back. Thank you!

lds/umode.lds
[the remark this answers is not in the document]
dgreid: We should file a github issue for moving salus address space. I think we'll want to enalbe paging earlier than we currently do, which will be complicated.

src/main.rs
dgreid: yes, it's just a hack until we figure out a better build system. I'm hoping to let zach and tom figure out blaze for hubris tasks then copy that.

riscv-elf/src/lib.rs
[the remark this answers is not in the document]
glg-rv: Changed. as_usize() returns usize. (with a comment).

riscv-elf/src/lib.rs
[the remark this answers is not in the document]
glg-rv: Wrote a few tests to check some (but definitely not all) of the cases of corrupted ELFs.

src/hyp_map.rs
glg-rv: Changed to user_map.

src/hyp_map.rs
glg-rv: Changed to your code. Thanks again.

lds/umode.lds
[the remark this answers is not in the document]
glg-rv: Filed https://github.com/rivosinc/salus/issues/180Filed

glg-rv:
Simplified `HypMapPopulatedRegion`. As segments are aligned to 4k we can eliminate the offset. This in turns makes everything simpler.

libuser/src/task_start.S
```diff
@@ -0,0 +1,27 @@
...
+    la gp, __global_pointer$
+.option pop
+    la sp, _stack_end
+
+    la t1, task_main
+    la ra, 1f
+    jr t1
+
+    j 1f
+1:
```
abrestic-rivos, line 24: Just `call task_main` is sufficient.

lds/umode.lds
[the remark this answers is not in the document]
abrestic-rivos: Ok, can we put some space between the bss and the stack so that we get a page fault when we inevitably hit that case?

abrestic-rivos (approved):
LGTM, thanks!

lds/umode.lds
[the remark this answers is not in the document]
glg-rv: As per inevitably, once we get an actual binary doing something the linking would've been changed.

Updated. Got rid of MEMORY command as it influences the way `ld` finds the next available address when assigning a section to a memory region.

Now there's an empty 4k section between the end of BSS and stack.

libuser/src/task_start.S
glg-rv: Ah good catch. Had a note somewhere when I saw your patches to change this too. Thanks!

Fixed.

glg-rv:
Added a fix to load Program Headers that have no data in file.

PR #179, opened 2022-12-09
Switch to per-host-CPU shared memory areas

### LINKED ISSUE #124: Make CSR shared memory interface be per-physical-CPU
Currently, TvmCpuNumRegisterSets/TvmCpuGetRegisterSet interface is defined per VM. 
It is defined as per VM in case TSM need to provide different layouts to different VMs. This provides more flexibility for TVM owners to selectively disable/enable extensions->shared memory layout in the future.

However, every VM has to make at least 4~5 ECALLS to retrieve the shared memory layout. This may affect the boot time of each VM. We may not want to penalize every VM creation for a corner case. We can always define special VMID value (-1) or additional flags to indicate the host wants global layout for all TVMs.

There are pros and cons in both methods. This is just a place holder issue created to address this issue in the future.

abrestic-rivos:
In order to better align with the proposed nested acceleration ("NACL") extension, reduce memory overhead, and make modification of HS-level CSRs that aren't directly relevant to a specific TVM less awkward, turn the per-vCPU shared memory area into a per-host-CPU shared memory area. The layout of this area matches that of the one in the proposed NACL extension, which notably includes space for all possible HS/VS-level CSRs. The layout of the scratch space, and the rules for how the structure is written by the TSM, are of course different however.

I've tried to split up the changes as cleanly as possible, but as we're changing the ABI here some of the patches have a larger surface area in order to make the changes atomic.

- Patch 1 makes `tvm_finalize()` set the entry point of the TVM directly, rather than latching from the shared memory area.
- Patches 2-4 are some refactors to prepare for per-host-CPU shared-memory areas.
- Patch 5 removes the dynamic vCPU shared-memory struct layout, going back to a static layout.
- Patch 6 adds a few missing CSR defines.
- Patch 7 switches to a per-host-CPU shared memory area (`TsmShmemArea`) and introduces a new call to register it (`tsm_set_shmem`).
- Patch 8 makes SCAUSE and STVAL be reported via CSR rather than shared memory.
- Patch 9 switches `TsmShmemArea` over to the "NACL" shared memory layout.

We should probably re-organize the TEE-Host FIDs given these changes; I'll do that if this series looks okay to everyone.

Fixes #124


sbi/src/api/tee_host.rs
```diff
@@ -106,6 +106,33 @@ pub fn get_info() -> Result<TsmInfo> {
...
 
+/// Registers the host <-> TSM shared memory area for the calling CPU. `shmem_page_addr` must
+/// be page-aligned and refer to a sufficient number of contiguous non-confidential pages to
+/// hold a `TsmShmemArea` struct. The pages are "pinned" in the non-confidential state until
+/// the shared-memory area is unregistered by calling `unregister_shmem()`.
+///
+/// # Safety
+///
+/// The caller must own the pages referenced by `shared_page_addr`, for the number of pages
+/// sufficient to hold the `TsmShmemArea` strucutre. Since memory within the shared-memory
```
dgreid, line 106: sp. structure

dgreid (approved):
thanks, looks cleaner to me.

src/host_vm.rs
```diff
@@ -4,7 +4,7 @@
```
glg-rv, line 4: Aren't we overloading the term "Host" with another meaning (as 'host the host' in the commit message implies). I.e., not only host vm but also host_context, meaning essentially the generalized parent of a vcpu?

sbi/src/api/tee_host.rs
```diff
@@ -165,9 +212,9 @@ pub fn get_vcpu_register_set(vmid: u64, index: u64) -> Result<RegisterSetLocatio
 /// # Safety
 ///
 /// The caller must own the pages referenced by `shared_page_addr`, for the number of pages
-/// sufficient to hold the structure layout returned from `get_vcpu_mem_layout()`. Since memory
-/// within the shared-memory state area may be read or written by the TSM at any time, the caller
-/// must treat the memory as volatile for the lifetime of the TVM.
+/// sufficient to hold the `TvmCpuSharedState` strucutre. Since memory within the shared-memory state
```
glg-rv: typo: `strucutre` instead of `structure`

sbi/src/api/tee_host.rs
abrestic-rivos: Done.

sbi/src/api/tee_host.rs
abrestic-rivos: Done.

PR #181, opened 2022-12-14
Run umode stub

glg-rv:
This PR introduces ability for Salus to run user mode code.

This PR is focused on the hypervisor side.

The first patch fixes a silly bug in printing registers on an unexpected salus trap. The second patch is a preparation step.

The third patch adds ability to reset the umode at the initial state, by saving the physical addresses of that particular page table of the writable VA mappings. This avoids using slow implementations of userva accesses .

The fourth patch creates the basic infrastructure for umode life time in salus.

The fifth adds some basic assembler and run the stub until the first ecall.

### DIFF
--- a/src/hyp_map.rs
+++ b/src/hyp_map.rs
@@ -232,7 +232,7 @@ pub struct HypMap {
 
 impl HypMap {
     // Create a new hypervisor map from a hardware memory mem map and a umode ELF.
-    pub fn new(mem_map: HwMemMap, umode_elf: ElfMap<'static>) -> Result<HypMap, Error> {
+    pub fn new(mem_map: HwMemMap, umode_elf: &ElfMap<'static>) -> Result<HypMap, Error> {
         // All shared mappings come from the HW Memory Map.
         let shared_regions = mem_map
             .regions()
--- a/src/hyp_map.rs
+++ b/src/hyp_map.rs
@@ -21,6 +21,15 @@ type PrivateRegionsVec = ArrayVec<PrivateRegion, MAX_PRIVATE_REGIONS>;
 // Shared regions vector.
 type SharedRegionsVec = ArrayVec<SharedRegion, MAX_SHARED_REGIONS>;
 
+/// Errors returned by HypMap.
+#[derive(Debug)]
+pub enum Error {
+    /// U-mode ELF segment is not page aligned.
+    ElfUnalignedSegment,
+    /// U-mode ELF segment is not in U-mode VA area.
+    ElfInvalidAddress,
+}
+
 // Represents a virtual address region of the hypervisor that will be the same in all pagetables.
 struct SharedRegion {
     vaddr: PageAddr<SupervisorVirt>,
@@ -101,6 +110,23 @@ impl SharedRegion {
     }
 }
 
+// U-mode mappings start here.
+const UMODE_VA_START: u64 = 0xffffffff00000000;
+// Size in bytes of the U-mode VA area.
+const UMODE_VA_SIZE: u64 = 128 * 1024 * 1024;
+// U-mode mappings end here.
+const UMODE_VA_END: u64 = UMODE_VA_START + UMODE_VA_SIZE;
+
+// Returns true if `addr` is contained in the U-mode VA area.
+fn is_umode_addr(addr: u64) -> bool {
+    (UMODE_VA_START..UMODE_VA_END).contains(&addr)
+}
+
+// Returns true if (`addr`, `addr` + `len`) is a valid non-empty range in the VA area.
+fn is_valid_umode_range(addr: u64, len: usize) -> bool {
+    len != 0 && is_umode_addr(addr) && is_umode_addr(addr + len as u64 - 1)
+}
+
 // Represents a virtual address region that will point to different physical page on each pagetable.
 struct PrivateRegion {
     // The address space where this region starts.
@@ -115,7 +141,7 @@ struct PrivateRegion {
 
 impl PrivateRegion {
     // Creates a per-pagetable region from an U-mode ELF segment.
-    fn from_umode_elf_segment(seg: &ElfSegment<'static>) -> Option<Self> {
+    fn from_umode_elf_segment(seg: &ElfSegment<'static>) -> Result<Self, Error> {
         // Sanity check for segment alignments.
         //
         // In general ELF might have segments overlapping in the same page, possibly with different
@@ -125,9 +151,13 @@ impl PrivateRegion {
         //
         // The following check enforces that the segment starts at a 4k page aligned address. Unless
         // the linking is completely corrupt, this also means that it starts at a different page.
-        // Assert is okay. This is a build error.
-        assert!(PageSize::Size4k.is_aligned(seg.vaddr()));
-
+        if !PageSize::Size4k.is_aligned(seg.vaddr()) {
+            return Err(Error::ElfUnalignedSegment);
+        }
+        // Sanity check for VA area of the segment.
+        if !is_valid_umode_range(seg.vaddr(), seg.size()) {
+            return Err(Error::ElfInvalidAddress);
+        }
         let pte_perms = match seg.perms() {
             ElfSegmentPerms::ReadOnly => PteLeafPerms::UR,
             ElfSegmentPerms::ReadWrite => PteLeafPerms::URW,
@@ -136,7 +166,7 @@ impl PrivateRegion {
         // Unwrap okay. `seg.vaddr()` has been checked to be 4k aligned.
         let vaddr = PageAddr::new(RawAddr::supervisor_virt(seg.vaddr())).unwrap();
         let pte_fields = PteFieldBits::leaf_with_perms(pte_perms);
-        Some(Self {
+        Ok(Self {
             vaddr,
             size: seg.size(),
             pte_fields,
@@ -201,8 +231,8 @@ pub struct HypMap {
 }
 
 impl HypMap {
-    /// Create a new hypervisor map from a hardware memory mem map and a umode ELF.
-    pub fn new(mem_map: HwMemMap, umode_elf: ElfMap<'static>) -> HypMap {
+    // Create a new hypervisor map from a hardware memory mem map and a umode ELF.
+    pub fn new(mem_map: HwMemMap, umode_elf: ElfMap<'static>) -> Result<HypMap, Error> {
         // All shared mappings come from the HW Memory Map.
         let shared_regions = mem_map
             .regions()
@@ -211,12 +241,12 @@ impl HypMap {
         // All private mappings come from the U-mode ELF.
         let private_regions = umode_elf
             .segments()
-            .filter_map(PrivateRegion::from_umode_elf_segment)
-            .collect();
-        HypMap {
+            .map(PrivateRegion::from_umode_elf_segment)
+            .collect::<Result<_, _>>()?;
+        Ok(HypMap {
             shared_regions,
             private_regions,
-        }
+        })
     }
 
     /// Create a new page table based on this memory map.
--- a/src/hyp_map.rs
+++ b/src/hyp_map.rs
@@ -11,21 +11,26 @@ use riscv_pages::{
 };
 use riscv_regs::{satp, LocalRegisterCopy, SatpHelpers};
 
-/// Maximum number of supervisor regions.
-const MAX_HYPMAP_SUPERVISOR_REGIONS: usize = 32;
-/// Maximum number of user regions.
-const MAX_HYPMAP_USER_REGIONS: usize = 32;
+// Maximum number of regions unique to every pagetable (private).
+const MAX_PRIVATE_REGIONS: usize = 32;
+// Maximum number of regions shared across all pagetables.
+const MAX_SHARED_REGIONS: usize = 32;
 
-/// Represents a virtual address region of the hypervisor with a fixed VA->PA Mapping.
-pub struct HypMapFixedRegion {
+// Private regions vector.
+type PrivateRegionsVec = ArrayVec<PrivateRegion, MAX_PRIVATE_REGIONS>;
+// Shared regions vector.
+type SharedRegionsVec = ArrayVec<SharedRegion, MAX_SHARED_REGIONS>;
+
+// Represents a virtual address region of the hypervisor that will be the same in all pagetables.
+struct SharedRegion {
     vaddr: PageAddr<SupervisorVirt>,
     paddr: PageAddr<SupervisorPhys>,
     page_count: usize,
     pte_fields: PteFieldBits,
 }
 
-impl HypMapFixedRegion {
-    // Create a supervisor VA region from a Hw Memory Map entry.
+impl SharedRegion {
+    // Create a shared region from a Hw Memory Map entry.
     fn from_hw_mem_region(r: &HwMemRegion) -> Option<Self> {
         let perms = match r.region_type() {
             HwMemRegionType::Available => {
@@ -87,8 +92,8 @@ impl HypMapFixedRegion {
             .zip(self.paddr.iter_from())
             .take(self.page_count)
         {
-            // Safe as we will create exactly one mapping to each page and will switch to
-            // using that mapping exclusively.
+            // Safety: all shared regions come from the HW memory map. we will create exactly one
+            // mapping for each page and will switch to using that mapping exclusively.
             unsafe {
                 mapper.map_addr(virt, phys, self.pte_fields).unwrap();
             }
@@ -96,8 +101,8 @@ impl HypMapFixedRegion {
     }
 }
 
-/// Represents a virtual address region that must be allocated and populated to be mapped.
-struct HypMapPopulatedRegion {
+// Represents a virtual address region that will point to different physical page on each pagetable.
+struct PrivateRegion {
     // The address space where this region starts.
     vaddr: PageAddr<SupervisorVirt>,
     // Number of bytes of the VA area
@@ -108,14 +113,14 @@ struct HypMapPopulatedRegion {
     data: Option<&'static [u8]>,
 }
 
-impl HypMapPopulatedRegion {
-    // Creates an user space virtual address region from a ELF segment.
-    fn from_user_elf_segment(seg: &ElfSegment<'static>) -> Option<Self> {
-        // Sanity Check for segment alignments.
+impl PrivateRegion {
+    // Creates a per-pagetable region from an U-mode ELF segment.
+    fn from_umode_elf_segment(seg: &ElfSegment<'static>) -> Option<Self> {
+        // Sanity check for segment alignments.
         //
         // In general ELF might have segments overlapping in the same page, possibly with different
         // permissions. In order to maintain separation and expected permissions on every page, the
-        // linker script for user ELF creates different segments at different pages. Failure to do so
+        // linker script for umode ELF creates different segments at different pages. Failure to do so
         // would make `map_range()` in `map()` fail.
         //
         // The following check enforces that the segment starts at a 4k page aligned address. Unless
@@ -131,7 +136,7 @@ impl HypMapPopulatedRegion {
         // Unwrap okay. `seg.vaddr()` has been checked to be 4k aligned.
         let vaddr = PageAddr::new(RawAddr::supervisor_virt(seg.vaddr())).unwrap();
         let pte_fields = PteFieldBits::leaf_with_perms(pte_perms);
-        Some(HypMapPopulatedRegion {
+        Some(Self {
             vaddr,
             size: seg.size(),
             pte_fields,
@@ -165,8 +170,9 @@ impl HypMapPopulatedRegion {
             .zip(pages.base().iter_from())
             .take(page_count as usize)
         {
-            // Safe because these pages are mapped into user mode and will not be accessed in
-            // supervisor mode.
+            // Safety: all per-pagetable regions are user mappings. User mappings are not considered
+            // aliases because they cannot be accessed by supervisor mode directly (sstatus.SUM needs
+            // to be 1).
             unsafe {
                 mapper.map_addr(virt, phys, self.pte_fields).unwrap();
             }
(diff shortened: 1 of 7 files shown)


src/hyp_map.rs
```diff
@@ -224,17 +291,23 @@ impl HypMap {
...
         for r in &self.supervisor_regions {
             r.map(&sv48, &mut || {
                 hyp_mem.take_pages_for_hyp_state(1).into_iter().next()
             });
         }
-        // Map user regions.
-        for r in &self.user_regions {
-            r.map(&sv48, hyp_mem);
+        // Map umode regions and create umode reset areas.
+        let mut umode_reset = ArrayVec::<UmodeResetArea, MAX_HYPMAP_UMODE_REGIONS>::new();
```
stillson, line 301: can ArrayVec::<UmodeResetArea, MAX_HYPMAP_UMODE_REGIONS> and the non-umode version be type aliases?

src/hyp_map.rs
```diff
@@ -96,8 +96,44 @@ impl HypMapFixedRegion {
...
+    // The size of the memory mapped for this region. A PageSize::Size4k multiple, bigger than `size`.
+    mapped_size: u64,
+    // The size in bytes of this region.
+    size: usize,
+    // Data to be populated at the beginning of the area.
+    data: Option<&'static [u8]>,
+}
+
+impl UmodeResetArea {
+    fn reset(&self) {
```
stillson, line 112: If you are resetting memory, do you have to flush the cache too?

src/main.rs
```diff
@@ -495,6 +497,21 @@ extern "C" fn kernel_init(hart_id: u64, fdt_addr: u64) {
...
+    // Unwrap okay: it has been called after `Umode::init()`
+    PerCpu::this_cpu().set_umode(UMODE_TASK.get().unwrap().new_cpu_state());
+
+    {
+        PerCpu::this_cpu()
+            .umode()
+            .activate()
+            .expect("Could not start U-mode")
+            .run()
+            .unwrap();
```
stillson, line 512: Does this have any dependency on the umode executable? If so, crashing seems a bit too extreme.

src/main.rs
```diff
@@ -533,9 +550,21 @@ extern "C" fn secondary_init(_hart_id: u64) {
...
 
+    {
+        PerCpu::this_cpu()
+            .umode()
+            .activate()
+            .expect("Could not start U-mode")
+            .run()
+            .unwrap();
+    }
+
```
stillson, line 567: see above

src/umode.S
```diff
@@ -0,0 +1,161 @@
+// Copyright (c) 2022 by Rivos Inc.
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
```
stillson, line 3: do you have to save floating point and vector registers?

glg-rv: There's been some discussion about that. Floating point is disabled at this point, so umode -- which should be really seen as a a form of lesser hypervisor code -- doesn't use it. Same for vector.
In general we might enable them in the future, and then yes, we'd have to restore and save them on entry and exit.

src/main.rs
glg-rv: This executes a stub umode. The umode is also considered part of build -- which at the moment isn't great, see previous PR. This code is strictly linked to the basic testing of the feature. On future PRs in the queue, that will add an API and in general more functional umode, this code will be removed.

src/main.rs
glg-rv: Likewise.

src/hyp_map.rs
glg-rv: As a matter of personal preference, I prefer to expose at least a simple combination of basic types, rather than having to look it up all the time. It's verbose, but such is rust and anything camel case.

Not dying on the hill of a typing fight, though.

src/hyp_map.rs
glg-rv: Risc-V caches are a bit of a cloudy area, at least to me. Are you implying some form of virtual address tagging in the cache?
If this is for RVWMO (thanks @abrestic-rivos) then I shouldn't have to because this memory is strictly per-cpu.

src/hyp_map.rs
```diff
@@ -96,8 +96,44 @@ impl HypMapFixedRegion {
...
+        let dest = self.paddr.bits() as *mut u8;
+        // Clear memory first.
+        // Safety: `dest` has `mapped_size` of memory mapped. The memory is owned by the hypervisor, by
+        // construction.
```
dgreid, line 116: either `reset` needs to be unsafe, or a `UmodeResetArea::new` needs to be added as unsafe and creation of this struct must be restricted to that function. Adding a new is cleaner as it better contains the unsafe code within one location of the user.

As is, safe code can create a `UmodeResetArea` with any pointer, then call reset to scribble on it.

It might require this struct to be its own rs file.

src/hyp_map.rs
```diff
@@ -96,8 +96,44 @@ impl HypMapFixedRegion {
     }
 }
 
-/// Represents a virtual address region that must be allocated and populated to be mapped.
-struct HypMapPopulatedRegion {
+// Represents an area that must be reset to restore U-mode original state.
```
abrestic-rivos, line 99: So, why do we need to "reset" the task every time we want to run it? That's a pretty big runtime penalty to pay in order to use U-mode. If we consider the first candidate use case -- certificate parsing -- how many kBs of memory are we copying/zeroing on each attempt to use U-mode?

Instead, how about a program flow that looks something like this:

```
fn main() {
    init();
    let mut result = 0
    loop {
        let task = ecall_get_next_task(result);
        result = handle_task(task);
    }
}
```

Basically the U-mode task runs in a loop with an ECALL that returns the result of the task and suspends execution until we have another task to give it.

I don't disagree that we'd eventually want the ability to reset U-mode, e.g. in the event the U-mode task is totally borked, but doing it every time we use it seems excessive.

src/hyp_map.rs
```diff
@@ -165,21 +211,42 @@ impl HypMapPopulatedRegion {
...
+        } else {
+            None
+        }
     }
 }
 
 /// A page table that contains hypervisor mappings.
 pub struct HypPageTable {
     inner: FirstStagePageTable<Sv48>,
+    umode_reset: ArrayVec<UmodeResetArea, MAX_HYPMAP_UMODE_REGIONS>,
```
abrestic-rivos, line 239: AFAICT the reason these are stored in the per-CPU `HypPageTable` is so that we can use the supervisor-level mappings from the identity map to access the U-mode memory, and those addresses will obviously differ between CPUs.

Any reason we can't use the (user) virtual addresses instead? The commit message says that using the supervisor mapping is faster, but is it really that big of a difference? We should only need to flip `sstatus.SUM` and then copying should "just work", e.g.

```
CSR.sstatus.modify(sstatus::sum.val(1));
core::ptr::copy(data.as_ptr(), dest_va, len);
CSR.sstatus.modify(sstatus::sum.val(0));
```

We know the mappings exist here, so we don't need to do the PF fixup tricks like we do when copying to/from guest memory.

And if you do end up using the user VA, there's no need to keep these reset areas around per-CPU; keeping the U-mode `ElfMap` in a single (immutable) global should be sufficient.

src/hyp_map.rs
abrestic-rivos: FWIW, I think the generic types for `ArrayVec` will get inferred here, so there's no need to spell them out explicitly. i.e. `ArrayVec::new()` should just work.

src/umode.rs
```diff
@@ -0,0 +1,106 @@
...
+        PerCpuUmode {
+            entry: self.entry,
+            arch: Mutex::new(UmodeCpuArchState::default()),
+        }
+    }
+}
+
+/// Per-CPU U-mode state.
+pub struct PerCpuUmode {
+    entry: u64,
```
abrestic-rivos, line 240: Is this ever different from the global `UMODE_TASK.entry`?

src/umode.rs
```diff
@@ -0,0 +1,106 @@
...
+    arch: Mutex<UmodeCpuArchState>,
```
abrestic-rivos, line 241: As we were talking about yesterday, have a `Mutex` protect a per-CPU struct is kinda pointless :). You can use a `RefCell` here instead to get interior mutability, like we do already for `PerCpu::vmid_tracker`.

src/umode.rs
```diff
@@ -101,6 +265,26 @@ impl<'active> Drop for ActiveUmode<'active> {
...
+            _ => Err(Error::UnexpectedTrap),
+        }
+    }
+
+    /// Run until it exits
+    fn run_to_exit(&mut self) {
+        unsafe {
+            // Safe to run umode code as it only touches memory assigned to it through umode mappings.
+            _run_umode(&mut *self.arch as *mut UmodeCpuArchState);
+        }
```
abrestic-rivos, line 285: Where are you setting up `sstatus`? At minimum, you'll need to make sure `sstatus.SPP` is set to 1 before you `sret`.

src/hyp_map.rs
glg-rv: The original discussion was on the following decision branch (_simplified_):

1. Single task with multiple functionalities or multiple tasks with single functionalities.

Single task was decided.

2. If single task, how do we ensure / at what level of paranoia we guarantee isolation inside it:
 a. We keep the umode stateful, and then just restart it if things go wrong.
 b. We keep the umode stateless, and we reset the state of umode as of its creation every time we run, providing context through an input area to be mapped/copied in when we run it -- keeping an eye on performance.

Consensus was somewhat on 'b', so this is what this code implement. 
'a' could be a viable option at this point (not requiring much change to the above code). And would possibly make it easier to just switch in the future to multiple task.

src/hyp_map.rs
glg-rv: Okay. This is how I implemented it originally. Then I had a look at how guest mem is copied and realised that we're extremely precise on the PC that might cause an exception. So I went and implemented copy_to/from_user (and a bzero) in the same way. Then realised that in the lack of optimization that would be a huge performance penalty. And came up with this.

What I did was consuming the `ElfMap` in the `Umode::init`, as parsing it everytime we reset is a bit pointless.

Will do.

src/umode.rs
glg-rv: Agreed. :-) Will do.

src/umode.rs
glg-rv: My understanding is that I shouldn't set it to one?

> The SPP bit indicates the privilege level at which a hart was executing before entering supervisor
mode. When a trap is taken, SPP is set to 0 if the trap originated from user mode, or 1 otherwise.
When an SRET instruction (see Section 3.2.2) is executed to return from the trap handler, the
privilege level is set to user mode if the SPP bit is 0, or supervisor mode if the SPP bit is 1; SPP
is then set to 0.

abrestic-rivos: As discussed offline -- you are absolutely right :) I got that backwards.

src/hyp_map.rs
glg-rv: In the end, had to create a type for the vec in hypmap because they became so long they had to be indented.

Here I used `ArrayVec::new()`.

src/hyp_map.rs
glg-rv: Changed in current code to use VA (see `umode_mem.rs`). The structure is created only if the address is in VA area. And the caller must guarantee the U-mode area is mapped (page tables are static, so you need to call it from the right place). In case it isn't, though, what we get is a page fault, rather than memory corruption.

src/hyp_map.rs
glg-rv: I restructured the code in `umode.rs` to make clear a few points. In general, we don't have to reset every time we run, but every time we end the lifetime of an active task. Even in the previous code I could hold a active task for some time and execute `run()` multiple times.

This new iteration hopefully makes things clearer.

For example, we could have a `PerCpu::` field that holds a `Option<UmodeActiveTask>`. This will allow us to reset on `panic` and to keep a stateful `umode` without changing anything in this code.

Essentially this discussion must be had, but I hope to have made it orthogonal to this PR.

src/hyp_map.rs
glg-rv: Changed. See `umode_mem.rs`.

src/umode.rs
glg-rv: It was a shot at avoiding a lifetime. :-) Changed in current code.

src/umode.rs
glg-rv: Changed to `RefCell`.

src/umode.rs
```diff
@@ -0,0 +1,205 @@
...
+static UMODE: Once<Umode> = Once::new();
+
+// Maximum number of reset areas.
+const MAX_RESET_AREAS: usize = 2;
+
+/// Represents the U-mode global state.
+pub struct Umode {
+    entry: u64,
+    reset_areas: ArrayVec<UmodeResetArea, MAX_RESET_AREAS>,
+}
```
abrestic-rivos, line 262: Isn't this information already available to us in `HypMap` and `ElfMap`? Do we need to duplicate it here?

src/umode_mem.rs
```diff
@@ -18,3 +20,80 @@ pub fn is_umode_addr(addr: u64) -> bool {
 pub fn is_valid_umode_range(addr: u64, len: usize) -> bool {
     len != 0 && is_umode_addr(addr) && is_umode_addr(addr + len as u64 - 1)
 }
+
+/// Errors returned by `umode_mem`.
+#[derive(Debug)]
+#[allow(clippy::enum_variant_names)]
```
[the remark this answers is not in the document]
glg-rv: Unhappy by the repetition of Invalid as a prefix. I either had to change it into Address, Offset, or become creative with adjectives.

src/hyp_map.rs
```diff
@@ -11,21 +11,26 @@ use riscv_pages::{
...
-pub struct HypMapFixedRegion {
+// Maximum number of regions unique to every pagetable.
+const MAX_PER_PAGETABLE_REGIONS: usize = 32;
+// Maximum number of regions shared across all pagetables.
+const MAX_SHARED_REGIONS: usize = 32;
+
+// Per-pagetable regions vector.
+type PerPagetableRegionsVec = ArrayVec<PerPagetableRegion, MAX_PER_PAGETABLE_REGIONS>;
+// Shared regions vector.
+type SharedRegionsVec = ArrayVec<SharedRegion, MAX_SHARED_REGIONS>;
```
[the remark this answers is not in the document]
glg-rv: `Shared` to `Global` yes, I was between the two. I looped through `shared` and `private`, `global` and `percpu`. Didn't chosse `PerCpu` in the end as this is really per page table. And if we have plan in the future to have multiple page table, this better signals the reality of what the code does.
At the moment a pagetable is created for each cpu, which is why is per cpu. But it's not defined in this code.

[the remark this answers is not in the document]
abrestic-rivos: Ok, maybe private to be more concise then? :)

src/hyp_map.rs
```diff
@@ -190,33 +196,33 @@ impl HypPageTable {
...
+            .filter_map(SharedRegion::from_hw_mem_region)
             .collect();
-        // All user regions come from the ELF segment.
-        let user_regions = user_map
+        // All per-pagetable mappings come from the U-mode ELF.
+        let per_pagetable_regions = umode_elf
             .segments()
-            .filter_map(HypMapPopulatedRegion::from_user_elf_segment)
+            .filter_map(PerPagetableRegion::from_umode_elf_segment)
             .collect();
```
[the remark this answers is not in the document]
abrestic-rivos: Oh whoops, probably. Definitely something that can be left as a future optimization.

src/umode_mem.rs
```diff
@@ -18,3 +20,80 @@ pub fn is_umode_addr(addr: u64) -> bool {
...
+        }
+    }
+
+    pub fn len(&self) -> usize {
+        self.len
+    }
+
+    /// Copy from hypervisor to the beginning of this memory range.
+    /// Caller must ensure that the U-mode memory range is mapped.
+    pub fn copy_to_umode(&self, data: &[u8]) {
```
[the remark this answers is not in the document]
abrestic-rivos: Clarification: while we check that it's in the U-mode VA range, there's nothing here that guarantees (or can reasonably check) that the VA actually maps user memory (as opposed to supervisor memory, or nothing at all). So the constructor probably should have an `unsafe` with the caller promising that the address range does in fact map user memory.

src/hyp_map.rs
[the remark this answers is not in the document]
glg-rv: Haha. Okay. :-)

src/umode.rs
glg-rv: Hm. Okay. The rational I would guess is to be able to have a different `sstatus` when running umode and generic supervisor code. I can see why in XS and FS these could be needed in the future.

src/umode.rs
```diff
@@ -0,0 +1,205 @@
...
+
+    /// Return this CPU's runnable task.
+    pub fn active_task(&self) -> Result<UmodeActiveTask, Error> {
+        PerCpu::this_cpu().umode_task().activate()
+    }
+}
+
+/// Represents a U-mode state with its running context.
+pub struct UmodeTask<'umode> {
+    umode: &'umode Umode,
```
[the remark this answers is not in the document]
glg-rv: Just wanted to maintain some truth in the dependency of the type. Not needed in this case, but in some experimental branches I have I use that. Can remove for this PR.

src/hyp_map.rs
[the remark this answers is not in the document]
glg-rv: Yes it is on my list. In general the way I put HypMap is to maintain things clear now, so can be made more efficient in the future. I am also unease with the generosity we have with pages at the moment.

src/umode_mem.rs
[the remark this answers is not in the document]
glg-rv: Okay, that makes sense. Thanks!

src/umode.rs
glg-rv: Yes, you hinted this at the previous iteration. `entry` is definitely there and doesn't have to be copied. Creating the reset areas (a small numbers) help with some basic sanity check and preparation of data for having simpler reset code. I want to avoid iterating over all ELF segments, selecting the writable ones, rounding up the page of the size and cutting data to size every time.

glg-rv: I think a word about the current design that might have been missed is necessary for context.

ELFs can be a pain to parse. Not because the format is complicated, but because there are a lot of things to check every time we want to use it in a safe way. Every offset, every value. The way I went in `riscv-elf` is to create a simple library that does all these checks once, creating a data structure that is essentially a sanitized image of the ELF file parts we're interested in, easy to use from the salus code. While the data is references is in the ELF file directly, this structure is not static. Rather, gets consumed later on.

This data structure, `ElfMap`, is then passed around to more specialized code (`HypMap` and `Umode`) to create the data it needs. The advantage of that is that `ElfMap` can be maintained simple and generic, and the other code can just parse this information to something it's much easier to handle. More importantly, this structure is consumed, and dropped by the last place that uses it (was `HypMap`, here becomes `umode`).

Now, based on the whole `unsafe` discussion, I am going to remove `UmodeMemRange`, which removes the need to create specialized reset areas structures in Umode. Furthermore, I agree that having a single point for mapping information is much better than disseminating it across the code.

In my next iteration, the direction I am thinking of going is to have `HypMap` being able to return mapping information it got from the `ElfMap`, and use it in umode on reset.

src/main.rs
glg-rv: In new update, removed the unwrap and just print the result. Again, this specific code will be removed pretty soon.

src/hyp_map.rs
glg-rv: Pushing implementation of reset in this iteration has clearly been a distracting issue. I have removed it from these changes. This PR is now strictly about running Umode until the first e-call.

Decision on when to reset the state is one that I am preparing (writing document as we speak, although possibly can be had in the github issue).

src/hyp_map.rs
glg-rv: Reset implementation has been removed. `umode_mem.rs` as well.

src/umode.rs
glg-rv: Set and restored `sstatus`, just to be extremely sure that we're not entering with the wrong `SPP` bit.

src/hyp_map.rs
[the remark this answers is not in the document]

src/umode_mem.rs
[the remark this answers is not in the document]
glg-rv: Reset has been removed.

src/umode.rs
glg-rv: Reset not implemented in this iteration. But most likely still going with what I said above.

src/main.rs
```diff
@@ -495,6 +497,18 @@ extern "C" fn kernel_init(hart_id: u64, fdt_addr: u64) {
...
+    UmodeTask::init(umode_elf);
+    // Setup U-mode task for this CPU.
+    UmodeTask::setup_this_cpu();
+
+    // Simple test: run U-mode until the first `ecall`.
+    let res = UmodeTask::get()
+        .activate()
+        .expect("Could not activate U-mode")
+        .run();
+    println!("umode result: {:?}", res);
```
abrestic-rivos, line 510: left over debug print?

src/main.rs
```diff
@@ -536,6 +550,15 @@ extern "C" fn secondary_init(_hart_id: u64) {
...
     me.set_online();
 
+    // Setup U-mode task for this CPU.
+    UmodeTask::setup_this_cpu();
+    // Simple test: run U-mode until the first `ecall`.
+    let res = UmodeTask::get()
+        .activate()
+        .expect("Could not activate U-mode")
+        .run();
+    println!("umode result: {:?}", res);
```
abrestic-rivos, line 560: ditto

src/umode.rs
```diff
@@ -0,0 +1,108 @@
...
+    }
+
+    /// Return this CPU's task. Must be call after `Self::setup_this_cpu()`.
+    pub fn get() -> &'static UmodeTask {
+        PerCpu::this_cpu().umode_task()
+    }
+
+    /// Activate this umode in order to run it.
+    pub fn activate(&self) -> Result<UmodeActiveTask, Error> {
+        let arch = self.arch.try_borrow_mut().map_err(|_| Error::TaskBusy)?;
```
abrestic-rivos, line 262: Since these are per-CPU shouldn't failure here be impossible, short of nested calls to `activate()`?

src/umode.rs
```diff
@@ -46,9 +51,175 @@ struct UmodeCpuArchState {
     trap_csrs: TrapRegs,
 }
 
+impl UmodeCpuArchState {
+    fn print(&self) {
```
abrestic-rivos, line 55: Instead of defining your own `print()` function, how about implementing `Display` or `Debug` instead?

src/umode.rs
```diff
@@ -0,0 +1,108 @@
...
+
+    /// Activate this umode in order to run it.
+    pub fn activate(&self) -> Result<UmodeActiveTask, Error> {
+        let arch = self.arch.try_borrow_mut().map_err(|_| Error::TaskBusy)?;
+        Ok(UmodeActiveTask { arch })
+    }
+}
+
+/// Represents a U-mode that is running or runnable. Not at initial state.
+pub struct UmodeActiveTask<'act> {
```
abrestic-rivos, line 268: bikeshed: I think `ActiveUmodeTask` sounds a little more natural

abrestic-rivos (approved):
LGTM, just a couple minor comments. Thanks!

src/main.rs
glg-rv: likewise.

src/umode.rs
glg-rv: A umode task is active when it is not at initial state. In the case of long term active state (no reset any time) I want to check that there's only *ONE* active task per cpu. Essentially this should be an unwrap(), but I am going for the style of letting the caller panic over it.

src/umode.rs
glg-rv: Good point.

src/umode.rs
glg-rv: Ack. My preference for prefixes is leaking. :-)

src/main.rs
glg-rv: Hm. This is just temporary code to check that things work. The unwrap was reviewed as a bit extreme in previous iteration (not by your review). So at least I'd have a print.

glg-rv: Switched back to `unwrap()`.

src/main.rs
glg-rv: Switched back to `unwrap()`.

src/umode.rs
glg-rv: No changes from this.

src/umode.rs
glg-rv: Done.

src/umode.rs
glg-rv: Done.

abrestic-rivos (approved):

dgreid (approved):

PR #182, opened 2022-12-15
Guest external interrupt virtualization

### LINKED ISSUE #91: IMSIC virtualization support

abrestic-rivos:
This series implements guest external interrupt virtualization for the host VM by trapping-and-emulating a (subset of) the interrupt-related HS CSRs like HIE and HGEIE, and then injecting SGEI when appropriate using HVICTL. We also need to make use of HGEI to trap on host VM external interrupts when running a guest TVM. The end result is a (mostly) complete interrupt virtualization implementation that should be compatible with LInux/KVM.

- Patch 1 makes `host_context` an enum in order to distinguish between host-VM (non-nested) and guest-TVM (nested) vCPUs.
- Patch 2 fixes timer interrupt virtualization by trapping guest TVMs if the host VM has a pending timer interrupt.
- Patch 3 exercises guest trapping on pending host timers in Tellus.
- Patches 4&5 are prep work for trapping-and-emulating the relevant HS-CSRs for the host VM.
- Patch 6 implements virtualization of HIE/HIP (SGIE only) and HGEIE/HGEIP for the host VM.
- Patch 7 tests the virtualize of HGEIE/HGEIP in Tellus/GuestVm.
- Patch 8 adds the relevant defines for the HVICTL CSR.
- Patch 9 makes use of SGEI to emulate SGEI for the host VM, and to trap guest TVM's on a host VM external interrupt.
- Patch 10 exercise SEI and SGEI in Tellus.

Note that we still don't attempt to emulate VS-level interrupts (VSEI, VSTI, VSSI) for the host VM. These are of limited usefulness for the users we care about (Linux/KVM) and the TSM -- regardless of deployment model -- is going to
forcibly delegate them to VS when running a TVM anyway.

Fixes #91


src/vm_cpu.rs
```diff
@@ -526,6 +531,15 @@ impl<'vcpu, 'pages, 'host, T: GuestStagePagingMode> ActiveVmCpu<'vcpu, 'pages, '
     pub fn run(&mut self) -> VmCpuTrap {
         self.complete_pending_mmio_op();
 
+        // Set up our host's timer at HS level so that we trap if our host's timer expires.
+        if let VmCpuParent::HostVm(ref host_vcpu) = self.host_context {
+            let host_sie = LocalRegisterCopy::new(host_vcpu.vs_csrs().vsie);
+            if host_sie.read(sie::stimer) != 0 {
+                CSR.stimecmp.set(host_vcpu.vs_csrs().vstimecmp);
+                CSR.sie.read_and_set_field(sie::stimer);
+            }
```
atishp04, line 590: This will trigger timer before you actually want to. vstimecmp is time w.r.t host in VS mode.
However, stiemcmp should be programmed with vstimecmp + htdelta.

atishp04: We can also do a micro optimization here by first checking if the vstimecmp + htdelta > current_time.
In that case, the timer is already expired and host can be scheduled again.

abrestic-rivos: Ah good catch on `htimedelta`! Fixed. I've left the latter as a TODO for now while I think of a clean way to avoid duplicating the SGEI vs SEI detection logic.

src/vm_interrupts.rs
```diff
@@ -87,17 +87,19 @@ pub struct VmCpuExtInterrupts {
     imsic_location: ImsicLocation,
     sw_file: SwFile,
     allowed_ids: AllowList,
+    num_guests: usize,
 }
 
 impl VmCpuExtInterrupts {
     /// Creates a new `VmCpuExtInterrupts` to track the external interrupt state of a vCPU with
     /// `imsic_location` as the location of the vCPU's virtualized IMSIC.
```
dgreid, line 95: nit: num_guests is pretty self explanatory, but would be nice to have in the doc comment .

dgreid (approved):
That's a big step forward in functionality. Thanks.

src/vm_interrupts.rs
abrestic-rivos: Done.

test-workloads/src/bin/guestvm.rs
```diff
@@ -319,6 +320,41 @@ fn test_memory_sharing() {
...
+    // EITHRESHOLD = 0
+    CSR.siselect.set(0x72);
+    CSR.sireg.set(0);
+    // EIE0[3] = 1
+    CSR.siselect.set(0xc0 + (INTERRUPT_ID / 64));
+    CSR.sireg.read_and_set_bits(1 << (INTERRUPT_ID % 64));
+
+    tee_guest::allow_external_interrupt(3).expect("GuestVm - AllowExternalInterrupt failed");
+
+    // VSIP implementation is buggy in QEMU; use VSTOPI to check that we got the interrupt instead.
```
atishp04, line 350: How does it manifest ? Have you sent a fix ?

abrestic-rivos: Mailed a fix last week: https://lore.kernel.org/qemu-devel/20221215224541.1423431-1-abrestic@rivosinc.com/

Basically the bit-masking in VSIP (and to some extent VSIE) is kinda bogus and SSIP ends up being the only readable bit.

atishp04: Thanks!

PR #183, opened 2022-12-16
PMU refactoring and bug fixes...

atulkharerivos:
Per the offline discussion, the Supervisor Counter Delegation (Sscdeleg) patches will be maintained on a separate branch pending finalization of the specification. 

The repurposed version includes only the refactoring changes and bug fixes that were part of the original PR. The Sscdeleg patches will be submitted in a separate PR.

1. The first commit adds a missing type definition for an event type
2. The second commit doesn't contain any functional changes, and refactors existing code in anticipation of forthcoming changes
3. The final commit contains a fix for an hitherto unnoticed bug related to the handling of PMU event configuration flags from the VM.

### DIFF
--- a/src/vm_pmu.rs
+++ b/src/vm_pmu.rs
@@ -6,8 +6,8 @@ use drivers::pmu;
 use riscv_regs::{RiscvCsrInterface, CSR, CSR_CYCLE};
 use s_mode_utils::print::*;
 use sbi::{
-    api::pmu::*, Error as SbiError, PmuCounterConfigFlags, PmuCounterStartFlags,
-    PmuCounterStopFlags, PmuEventType, Result as SbiResult,
+    Error as SbiError, PmuCounterConfigFlags, PmuCounterStartFlags, PmuCounterStopFlags,
+    PmuEventType, Result as SbiResult,
 };
 
 #[derive(Default, Copy, Clone)]
@@ -103,9 +103,9 @@ impl VmPmuState {
             .read_and_clear_bits(1 << (csr - CSR_CYCLE as u64));
     }
 
-    /// Updates internal state for PMU counters.
-    /// This should be called following a successful SBI call to configure counters.
-    pub fn update_configured_counter(
+    // Updates internal state for PMU counters.
+    // This should be called following a successful SBI call to configure counters.
+    fn update_configured_counter(
         &mut self,
         counter_index: u64,
         config_flags: PmuCounterConfigFlags,
@@ -147,9 +147,9 @@ impl VmPmuState {
         }
     }
 
-    /// Updates internal state for PMU counters. Assumes a sanitized counter_index and counter_mask.
-    /// This should be called following a successful SBI call to start counters.
-    pub fn update_started_counters(&mut self, counter_index: u64, counter_mask: u64) {
+    // Updates internal state for PMU counters. Assumes a sanitized counter_index and counter_mask.
+    // This should be called following a successful SBI call to start counters.
+    fn update_started_counters(&mut self, counter_index: u64, counter_mask: u64) {
         use PmuCounterState::*;
         let bitmask_iter = CounterMaskIter::new(counter_index, counter_mask);
         for i in bitmask_iter {
@@ -160,9 +160,9 @@ impl VmPmuState {
         }
     }
 
-    /// Updates internal state for PMU counters. Assumes a sanitized counter_index and counter_mask.
-    /// This should be called following a successful SBI call to stop counters.
-    pub fn update_stopped_counters(
+    // Updates internal state for PMU counters. Assumes a sanitized counter_index and counter_mask.
+    // This should be called following a successful SBI call to stop counters.
+    fn update_stopped_counters(
         &mut self,
         counter_index: u64,
         counter_mask: u64,
@@ -192,12 +192,8 @@ impl VmPmuState {
         }
     }
 
-    /// Returns a filtered counter_mask if the PMU counter range can be started.
-    pub fn get_startable_counter_range(
-        &self,
-        counter_index: u64,
-        counter_mask: u64,
-    ) -> SbiResult<u64> {
+    // Returns a filtered counter_mask if the PMU counter range can be started.
+    fn get_startable_counter_range(&self, counter_index: u64, counter_mask: u64) -> SbiResult<u64> {
         use PmuCounterState::*;
         let pmu_info = pmu::PmuInfo::get()?;
         let counter_mask = pmu_info.filter_counter_mask(counter_index, counter_mask)?;
@@ -207,12 +203,8 @@ impl VmPmuState {
             .map_or_else(|| Err(SbiError::InvalidParam), |_| Ok(counter_mask))
     }
 
-    /// Returns a filtered counter mask if the PMU counter range can be stopped.
-    pub fn get_stoppable_counter_range(
-        &self,
-        counter_index: u64,
-        counter_mask: u64,
-    ) -> SbiResult<u64> {
+    // Returns a filtered counter mask if the PMU counter range can be stopped.
+    fn get_stoppable_counter_range(&self, counter_index: u64, counter_mask: u64) -> SbiResult<u64> {
         let pmu_info = pmu::PmuInfo::get()?;
         let counter_mask = pmu_info.filter_counter_mask(counter_index, counter_mask)?;
         // The current PMU driver attempts to stop counters before configuration since some platform
@@ -221,8 +213,8 @@ impl VmPmuState {
         Ok(counter_mask)
     }
 
-    /// Returns a filtered counter_mask if the PMU counter range can be configured.
-    pub fn get_configurable_counter_range(
+    // Returns a filtered counter_mask if the PMU counter range can be configured.
+    fn get_configurable_counter_range(
         &self,
         counter_index: u64,
         counter_mask: u64,
@@ -272,29 +264,30 @@ impl VmPmuState {
         }
     }
 
+    fn reset_all_counters(&mut self, counter_mask: u64) -> SbiResult<()> {
+        if counter_mask != 0 {
+            self.stop_counters(
+                0,
+                counter_mask,
+                PmuCounterStopFlags::default().set_reset_flag(),
+            )
+            .or_else(|e| {
+                // Treat already stopped error as success
+                if matches!(e, SbiError::AlreadyStopped) {
+                    Ok(())
+                } else {
+                    Err(e)
+                }
+            })
+        } else {
+            Ok(())
+        }
+    }
+
     /// Saves the internal state for PMU counters. Stops started counters, and resets all configured
     /// counters. This should be called in anticipation of an outbound context switch.
     pub fn save_counters(&mut self) {
         use PmuCounterState::*;
-        fn reset_all_counters(counter_mask: u64) -> SbiResult<()> {
-            if counter_mask != 0 {
-                stop_counters(
-                    0,
-                    counter_mask,
-                    PmuCounterStopFlags::default().set_reset_flag(),
-                )
-                .or_else(|e| {
-                    // Treat already stopped error as success
-                    if matches!(e, SbiError::AlreadyStopped) {
-                        Ok(())
-                    } else {
-                        Err(e)
-                    }
-                })
-            } else {
-                Ok(())
-            }
-        }
 
         if let Ok(pmu_info) = pmu::PmuInfo::get() {
             let num_counters = pmu_info.get_num_counters() as usize;
@@ -308,7 +301,8 @@ impl VmPmuState {
                     }
                 }
             }
-            let result = reset_all_counters(counter_mask);
+
+            let result = self.reset_all_counters(counter_mask);
             if result.is_err() {
                 println!(
                     "Warning: PMU failed to reset counters with mask {counter_mask:x}, {result:?}"
@@ -317,47 +311,49 @@ impl VmPmuState {
         }
     }
 
+    fn resume_counter(&mut self, counter_index: u64, c: &CounterState) -> SbiResult<()> {
+        let start_flags = PmuCounterStartFlags::default().set_init_value();
+        self.start_counters(counter_index, 0x1, start_flags, c.value)
+            .map(|_| VmPmuState::set_hcounteren_bit(counter_index))
+    }
+
+    fn configure_counter(&mut self, counter_index: u64, c: &CounterState) -> SbiResult<u64> {
+        let config_flags = c
+            .config_flags
+            .unset_auto_start()
+            .unset_skip_match()
+            .unset_clear_value();
+        self.configure_matching_counters(
+            counter_index,
+            0x1,
+            config_flags,
+            c.event_type,
+            c.event_data,
+        )
+    }
+
     /// Restores configured PMU counters, restarts started counters and enables CSR access as
     /// necessary. This should be called in anticipation of an inbound context switch.
     pub fn restore_counters(&mut self) {
         use PmuCounterState::*;
-        fn resume_counter(counter_index: u64, c: &CounterState) -> SbiResult<()> {
-            let start_flags = PmuCounterStartFlags::default().set_init_value();
-            start_counters(counter_index, 0x1, start_flags, c.value)
-                .map(|_| VmPmuState::set_hcounteren_bit(counter_index))
-        }
-
-        fn configure_counter(counter_index: u64, c: &CounterState) -> SbiResult<u64> {
-            let config_flags = c
-                .config_flags
-                .unset_auto_start()
-                .unset_skip_match()
-                .unset_clear_value();
-            configure_matching_counters(
-                counter_index,
-                0x1,
-                config_flags,
-                c.event_type,
-                c.event_data,
-            )
-        }
 
         if let Ok(pmu_info) = pmu::PmuInfo::get() {
             let num_counters = pmu_info.get_num_counters() as usize;
-            for (i, state) in self.counter_state.iter_mut().take(num_counters).enumerate() {
+            for i in 0..num_counters {
+                let state = &self.counter_state[i].clone();
                 let counter_index = i as u64;
                 let is_started_counter = matches!(state, Started(_));
                 match state {
                     Configured(c) | Started(c) => {
-                        let result = configure_counter(counter_index, c).and_then(|_| {
+                        let result = self.configure_counter(counter_index, c).and_then(|_| {
                             if is_started_counter {
-                                resume_counter(counter_index, c)
+                                self.resume_counter(counter_index, c)
                             } else {
                                 Ok(())
                             }
                         });
                         if result.is_err() {
-                            *state = Poisoned(*c);
+                            self.counter_state[i] = Poisoned(*c);
                             println!(
                                 "Warning: Failed to restore counter {counter_index}, {result:?}"
                             );
@@ -367,5 +363,70 @@ impl VmPmuState {
                 }
             }
         }
+    }
+
+    /// Calls the SBI configure_matching_counters() and performs internal bookkeeping on counter state.
+    pub fn configure_matching_counters(
+        &mut self,
+        counter_index: u64,
+        counter_mask: u64,
+        config_flags: PmuCounterConfigFlags,
+        event_type: PmuEventType,
+        event_data: u64,
+    ) -> SbiResult<u64> {
+        let config_flags = config_flags.set_sinh().set_minh();
+        let counter_mask =
+            self.get_configurable_counter_range(counter_index, counter_mask, config_flags)?;
+        let platform_counter_index = sbi::api::pmu::configure_matching_counters(
+            counter_index,
+            counter_mask,
+            config_flags,
+            event_type,
+            event_data,
+        )?;
+        self.update_configured_counter(
+            platform_counter_index,
+            config_flags,
+            event_type,
+            event_data,
+        )?;
+        Ok(platform_counter_index)
+    }
+
+    /// Calls the SBI start_counters() and performs internal bookkeeping on counter state.
+    pub fn start_counters(
+        &mut self,
+        counter_index: u64,
+        counter_mask: u64,
+        start_flags: PmuCounterStartFlags,
+        initial_value: u64,
+    ) -> SbiResult<()> {
+        let counter_mask = self.get_startable_counter_range(counter_index, counter_mask)?;
+        let result =
+            sbi::api::pmu::start_counters(counter_index, counter_mask, start_flags, initial_value);
+        // Special case "already started" to handle counters that are autostarted following configuration.
+        // Examples of such counters include the legacy timer and insret.
+        if result.is_ok() || matches!(result, Err(SbiError::AlreadyStarted)) {
+            self.update_started_counters(counter_index, counter_mask);
+        }
+        result
+    }
+
+    /// Calls the SBI stop_counters() and performs internal bookkeeping on counter state.
+    pub fn stop_counters(
+        &mut self,
+        counter_index: u64,
+        counter_mask: u64,
+        stop_flags: PmuCounterStopFlags,
+    ) -> SbiResult<()> {
+        let counter_mask = self.get_stoppable_counter_range(counter_index, counter_mask)?;
+        let result = sbi::api::pmu::stop_counters(counter_index, counter_mask, stop_flags);
+        // Special case "already stopped" to handle counters that can be reset following a stop
+        if result.is_ok()
+            || (matches!(result, Err(SbiError::AlreadyStopped)) && stop_flags.is_reset_flag())
+        {
+            self.update_stopped_counters(counter_index, counter_mask, stop_flags);
+        }
+        result
     }
 }
--- a/src/vm_pmu.rs
+++ b/src/vm_pmu.rs
@@ -374,13 +374,20 @@ impl VmPmuState {
         event_type: PmuEventType,
         event_data: u64,
     ) -> SbiResult<u64> {
-        let config_flags = config_flags.set_sinh().set_minh();
+        // Translate uinh/sinh from VM
+        let mut config_flags = config_flags;
+        if config_flags.is_uinh() {
+            config_flags = config_flags.set_vuinh();
+        }
+        if config_flags.is_sinh() {
+            config_flags = config_flags.set_vsinh();
+        }
         let counter_mask =
             self.get_configurable_counter_range(counter_index, counter_mask, config_flags)?;
         let platform_counter_index = sbi::api::pmu::configure_matching_counters(
             counter_index,
             counter_mask,
-            config_flags,
+            config_flags.set_sinh().set_minh(),
             event_type,
             event_data,
         )?;
--- a/sbi/src/pmu.rs
+++ b/sbi/src/pmu.rs
@@ -287,6 +287,8 @@ pub enum PmuEventType {
     Hardware(PmuHardware),
     /// Represents the hardware cache events (type #1) in the SBI documentation.
     Cache(PmuHwCacheParams),
+    /// Represents a raw event (type #2) in the SBI documentation
+    RawEvent,
     /// Represents the firmware events (type #15) in the SBI documentation.
     Firmware(PmuFirmware),
 }
@@ -302,11 +304,13 @@ impl PmuEventType {
     /// Returns the encoded representation of the type.
     pub fn raw(&self) -> u64 {
         const HARDWARE_CACHE_EVENT_TYPE: u64 = 1;
+        const HARDWARE_RAW_EVENT_TYPE: u64 = 2;
         const FIRMWARE_EVENT_TYPE: u64 = 0xF;
         use PmuEventType::*;
         match self {
             Hardware(p) => *p as u64,
             Cache(p) => p.raw() | (HARDWARE_CACHE_EVENT_TYPE << EVENT_TYPE_SHIFT),
+            RawEvent => HARDWARE_RAW_EVENT_TYPE << EVENT_TYPE_SHIFT,
             Firmware(p) => *p as u64 | (FIRMWARE_EVENT_TYPE << EVENT_TYPE_SHIFT),
         }
     }
@@ -322,6 +326,7 @@ impl PmuEventType {
             1 => Ok(Cache(PmuHwCacheParams::from_raw_value(
                 value & EVENT_TYPE_INVERSE_MASK,
             )?)),
+            2 if (value & EVENT_TYPE_INVERSE_MASK) == 0 => Ok(RawEvent),
             0xF => Ok(Firmware(PmuFirmware::from_raw_value(
                 value & EVENT_TYPE_INVERSE_MASK,
             )?)),
--- a/src/vm.rs
+++ b/src/vm.rs
@@ -755,22 +755,12 @@ impl<'a, T: GuestStagePagingMode> FinalizedVm<'a, T> {
             initial_value: u64,
             active_vcpu: &mut ActiveVmCpu<T>,
         ) -> EcallResult<u64> {
-            let counter_mask = active_vcpu
-                .pmu()
-                .get_startable_counter_range(counter_index, counter_mask)?;
-            let result = sbi::api::pmu::start_counters(
+            let result = active_vcpu.pmu().start_counters(
                 counter_index,
                 counter_mask,
                 start_flags,
                 initial_value,
             );
-            // Special case "already started" to handle counters that are autostarted following configuration.
-            // Examples of such counters include the legacy timer and insret.
-            if result.is_ok() || matches!(result, Err(SbiError::AlreadyStarted)) {
-                active_vcpu
-                    .pmu()
-                    .update_started_counters(counter_index, counter_mask);
-            }
             result.map(|_| 0).map_err(EcallError::from)
         }
 
@@ -780,18 +770,9 @@ impl<'a, T: GuestStagePagingMode> FinalizedVm<'a, T> {
             stop_flags: PmuCounterStopFlags,
             active_vcpu: &mut ActiveVmCpu<T>,
         ) -> EcallResult<u64> {
-            let counter_mask = active_vcpu
+            let result = active_vcpu
                 .pmu()
-                .get_stoppable_counter_range(counter_index, counter_mask)?;
-            let result = sbi::api::pmu::stop_counters(counter_index, counter_mask, stop_flags);
-            // Special case "already stopped" to handle counters that can be reset following a stop
-            if result.is_ok()
-                || (matches!(result, Err(SbiError::AlreadyStopped)) && stop_flags.is_reset_flag())
-            {
-                active_vcpu
-                    .pmu()
-                    .update_stopped_counters(counter_index, counter_mask, stop_flags);
-            }
+                .stop_counters(counter_index, counter_mask, stop_flags);
             result.map(|_| 0).map_err(EcallError::from)
         }
 
@@ -803,26 +784,14 @@ impl<'a, T: GuestStagePagingMode> FinalizedVm<'a, T> {
             event_data: u64,
             active_vcpu: &mut ActiveVmCpu<T>,
         ) -> EcallResult<u64> {
-            let config_flags = config_flags.set_sinh().set_minh();
-            let counter_mask = active_vcpu.pmu().get_configurable_counter_range(
-                counter_index,
-                counter_mask,
-                config_flags,
-            )?;
-            let platform_counter_index = sbi::api::pmu::configure_matching_counters(
+            let result = active_vcpu.pmu().configure_matching_counters(
                 counter_index,
                 counter_mask,
                 config_flags,
                 event_type,
                 event_data,
-            )?;
-            active_vcpu.pmu().update_configured_counter(
-                platform_counter_index,
-                config_flags,
-                event_type,
-                event_data,
-            )?;
-            Ok(platform_counter_index)
+            );
+            result.map_err(EcallError::from)
         }
 
         match pmu_func {


riscv-regs/src/inst.rs
```diff
@@ -2863,6 +2863,12 @@ pub const CSR_MHPMCOUNTER28H: u16 = 0xb9c;
 pub const CSR_MHPMCOUNTER29H: u16 = 0xb9d;
 pub const CSR_MHPMCOUNTER30H: u16 = 0xb9e;
 pub const CSR_MHPMCOUNTER31H: u16 = 0xb9f;
+pub const CSR_SMCOUNTEREN: u16 = 0xd00;
```
atishp04, line 2866: Just leave a comment here saying these are temporary. The actual CSR number gets finalized after ARCH review

src/vm_pmu.rs
```diff
@@ -373,6 +373,112 @@ impl VmPmuState {
...
+        }?;
+
+        if config_flags.is_clear_value() {
+            CSR.scounterselect.set(index);
+            CSR.shpmcounter.set(0);
+        }
+
+        if config_flags.is_auto_start() {
+            let value = CSR.scountinhibit.get() & !(1 << index);
+            CSR.scountinhibit.set(value);
```
atishp04, line 418: set ? to start it should be cleared.

src/vm_pmu.rs
```diff
@@ -373,6 +373,112 @@ impl VmPmuState {
...
+        let bitmask_iter = CounterMaskIter::new(counter_index, counter_mask & FIXED_COUNTERS_MASK);
+        for index in bitmask_iter.into_iter() {
+            if matches!(self.counter_state[index], Configured(_)) {
+                let index = index as u64;
+                if start_flags.is_init_value() {
+                    CSR.scounterselect.set(index);
+                    CSR.shpmcounter.set(initial_value);
+                }
+                let value = CSR.scountinhibit.get() & !(1 << index);
+                CSR.scountinhibit.set(value);
```
atishp04, line 447: clear as well

src/vm_pmu.rs
```diff
@@ -373,6 +373,112 @@ impl VmPmuState {
         }
     }
 
+    // Implementation of PMU counter configuration on platforms with Sscdeleg support
+    // TODO: Use code in the actual implementation
+    #[allow(dead_code)]
+    fn configure_matching_counters_internal(
```
atishp04, line 379: nit: The suffix can be Sscdeleg to be explicit. The SBI function can be sbi as a explicit prefix.

src/vm_pmu.rs
```diff
@@ -373,6 +373,112 @@ impl VmPmuState {
...
+        &mut self,
+        counter_index: u64,
+        counter_mask: u64,
+        start_flags: PmuCounterStartFlags,
+        initial_value: u64,
+    ) -> SbiResult<()> {
+        use PmuCounterState::*;
+        let pmu_info = pmu::PmuInfo::get()?;
+        let counter_mask = pmu_info.filter_counter_mask(counter_index, counter_mask)?;
+        const FIXED_COUNTERS_MASK: u64 = !0x7;
```
atishp04, line 437: Fixed countes are 0x7. Is it one of the rust magic I don't understand ?

src/vm_pmu.rs
```diff
@@ -373,6 +373,112 @@ impl VmPmuState {
...
+        let bitmask_iter = CounterMaskIter::new(counter_index, counter_mask & FIXED_COUNTERS_MASK);
```
atishp04, line 438: smcountern should be used to filter the programmable ones.
Sscountdeleg also applies to fixed ones. That should be covered.

src/vm_pmu.rs
```diff
@@ -373,6 +373,112 @@ impl VmPmuState {
...
+        counter_index: u64,
+        counter_mask: u64,
+        config_flags: PmuCounterConfigFlags,
+        _event_type: PmuEventType,
+        _event_data: u64,
+    ) -> SbiResult<u64> {
+        // TODO: Finalize the logic for using event_type and event_data
+        use PmuCounterState::*;
+        let pmu_info = pmu::PmuInfo::get()?;
+        let counter_mask = pmu_info.filter_counter_mask(counter_index, counter_mask)?;
```
abrestic-rivos, line 390: The information read from `PmuInfo` is provided by the PMU SBI extension, which a system with Sscdeleg may not even have. Is there a need to bother with it here? Shouldn't we be filtering against `smcounteren` instead?

We should also figure out what do with `PmuInfo` in the presence of Sscdeleg -- does it continue to use only SBI (and thus we'd ignore it in the Sscdeleg case), or do we have it use `smcounteren` when Sscdeleg is present?

src/vm_pmu.rs
```diff
@@ -373,6 +373,112 @@ impl VmPmuState {
...
+        const FIXED_COUNTERS_MASK: u64 = !0x7;
+        let index = if !config_flags.is_skip_match() {
+            let mut bitmask_iter =
+                CounterMaskIter::new(counter_index, counter_mask & FIXED_COUNTERS_MASK);
```
abrestic-rivos, line 394: `counter_mask` is relative to `counter_index`, so if the intention is to mask out the fixed counters this logic is incorrect if `counter_index` != 0.

But also, we do need to support configuring `cycle` and `instret`, right?

src/vm_pmu.rs
abrestic-rivos: It is getting cleared. Though a more concise (and readable) way to write this would be: `CSR.scountinhibit.read_and_clear_bits(1 << index)`

src/vm_pmu.rs
```diff
@@ -373,6 +373,112 @@ impl VmPmuState {
...
+            CSR.scounterselect.set(index);
+            CSR.shpmcounter.set(0);
+        }
+
+        if config_flags.is_auto_start() {
+            let value = CSR.scountinhibit.get() & !(1 << index);
+            CSR.scountinhibit.set(value);
+        }
+
+        Ok(index as u64)
```
abrestic-rivos, line 421: Don't we need to program `shpmevent` as well for event selection and privilege mode filtering?

src/vm_pmu.rs
abrestic-rivos: This mask is the programmable ones. I think you might be getting tripped up by `!` being bitwise-NOT in rust :)

src/vm_pmu.rs
atishp04: Ahh. Yes. Changing the name would be less confusing.

src/vm_pmu.rs
```diff
@@ -317,47 +319,49 @@ impl VmPmuState {
...
-                c.event_type,
-                c.event_data,
-            )
-        }
 
         if let Ok(pmu_info) = pmu::PmuInfo::get() {
             let num_counters = pmu_info.get_num_counters() as usize;
-            for (i, state) in self.counter_state.iter_mut().take(num_counters).enumerate() {
+            for i in 0..num_counters {
+                let state = &self.counter_state[i].clone();
```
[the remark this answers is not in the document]
atulkharerivos: The issue was that we need to pass a mutable reference to the new function, but the old code was using a mutable iterator, which precludes using another reference to self.

src/vm_pmu.rs
```diff
@@ -368,4 +372,49 @@ impl VmPmuState {
...
+        start_flags: PmuCounterStartFlags,
+        initial_value: u64,
+    ) -> SbiResult<()> {
+        // TODO: Call internal implementation if Sscdeleg is implemented
+        start_counters(counter_index, counter_mask, start_flags, initial_value)
+    }
+
+    /// Helper function to call either SBI, or the internal implementation
+    /// for stop_counters() depending on whether Sscdeleg is implemented
+    pub fn stop_counters_helper(
```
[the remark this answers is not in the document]
atulkharerivos: Yeah, it definitely can be structured a little better -- will refactor it.

src/vm_pmu.rs
atulkharerivos: Right, was unsure about how we wanted to support the fixed counters. Per the conversation with Atish, it's best to just use fixed counters for 0, 2 (instead of attempting the SBI approach of trying to mapping them to programmable counters).

src/vm_pmu.rs
atulkharerivos: Per the conversation with Atish, we'll simply use fixed counters for 0, 2 (1 isn't valid).

src/vm_pmu.rs
atulkharerivos: Added counter enumeration based on Sscdeleg.

drivers/src/pmu.rs
```diff
@@ -50,19 +53,62 @@ impl PmuInfo {
...
-            let _ = pmu::stop_counters(0, pmu_info.valid_hw_counters_mask, stop_flags);
+            if !cpu_id.has_sscdeleg() {
+                let _ = pmu::stop_counters(0, pmu_info.valid_hw_counters_mask, stop_flags);
+            } else {
+                let delegated_counters = pmu_info.valid_hw_counters_mask;
+                let base_csr = riscv_regs::CSR_CYCLE as u64;
+                riscv_regs::CSR.scountinhibit.set(delegated_counters);
+                for i in 0..MAX_HARDWARE_COUNTERS as u64 {
+                    if let Ok(info) = pmu_info.get_counter_info(i) {
+                        let csr_index = info.get_csr() - base_csr;
```
abrestic-rivos, line 106: Aren't these identity-mapped, i.e. counter `i` is at CSR `cycle + i`? In which case, can't you write `CSR.scounterselect.set(i)`?

src/vm_pmu.rs
```diff
@@ -368,4 +373,149 @@ impl VmPmuState {
...
+            let bitmask_iter = CounterMaskIter::new(counter_index, counter_mask);
+            for index in bitmask_iter {
+                if matches!(self.counter_state[index], Configured(_)) {
+                    let index = index as u64;
+                    if start_flags.is_init_value() {
+                        CSR.scounterselect.set(index);
+                        CSR.shpmcounter.set(initial_value);
+                    }
+                    CSR.scountinhibit
+                        .set(CSR.scountinhibit.read_and_clear_bits(1 << index));
```
abrestic-rivos, line 487: Just `CSR.scountinhibit.read_and_clear_bits(1 << index)`

src/vm_pmu.rs
```diff
@@ -368,4 +373,149 @@ impl VmPmuState {
...
+            CSR.shpmevent
+                .read_and_set_field(riscv_regs::defs::shpmevent::sinh);
+
+            if config_flags.is_clear_value() {
+                CSR.shpmcounter.set(0);
+            }
+
+            if config_flags.is_auto_start() {
+                CSR.scountinhibit
+                    .set(CSR.scountinhibit.read_and_clear_bits(1 << index));
```
abrestic-rivos, line 457: You just need the `read_and_clear_bits()` here. As written, this will actually undo the bit-setting.

src/vm_pmu.rs
```diff
@@ -368,4 +373,149 @@ impl VmPmuState {
...
+                let mut bitmask_iter = CounterMaskIter::new(counter_index, counter_mask);
+                bitmask_iter
+                    .find(|i| {
+                        matches!(self.counter_state[*i], NotConfigured) || is_fixed_counter(*i)
+                    })
+                    .map_or_else(|| Err(SbiError::InvalidParam), |i| Ok(i as u64))
+            } else {
+                if is_timer(counter_index) {
+                    return Err(SbiError::InvalidParam);
+                }
```
abrestic-rivos, line 425: Why only check this in the skip_match case? Shouldn't we be checking this always?

src/vm_pmu.rs
```diff
@@ -368,4 +373,149 @@ impl VmPmuState {
...
+            }?;
+
+            CSR.scounterselect.set(index);
+            if matches!(
+                event_type,
+                PmuEventType::Cache(_) | PmuEventType::Hardware(_)
+            ) {
+                CSR.shpmevent.read_and_set_bits(event_type.raw());
+            } else {
```
abrestic-rivos, line 445: Isn't there some translation needed here from the event type to the actual HW event number? I believe this comes from the DT.

src/vm_pmu.rs
```diff
@@ -368,4 +373,149 @@ impl VmPmuState {
...
+                CSR.shpmevent.read_and_set_bits(event_data);
+            }
+            CSR.shpmevent
+                .read_and_set_field(riscv_regs::defs::shpmevent::sinh);
```
abrestic-rivos, line 449: We should respect a CFG_FLAG_SET_SINH request from the VM, and set `shpmevent.VSINH` appropriately.

src/vm_pmu.rs
```diff
@@ -368,4 +373,149 @@ impl VmPmuState {
...
+            }?;
+
+            CSR.scounterselect.set(index);
+            if matches!(
+                event_type,
+                PmuEventType::Cache(_) | PmuEventType::Hardware(_)
+            ) {
+                CSR.shpmevent.read_and_set_bits(event_type.raw());
+            } else {
+                CSR.shpmevent.read_and_set_bits(event_data);
```
abrestic-rivos, line 446: Instead of manipulating the CSR directly multiple times, stage the value in a local variable and then actually write to the CSR when done. e.g.

```
let shpmevent = LocalRegisterCopy::new(0);
shpmevent.modify(shpmevent::event_selector.val(event));
shpmevent.modify(shpmevent::sinh.val(1));
CSR.shpmevent.set(shpmevent.get());
```

src/vm_pmu.rs
atulkharerivos: Since the mask gets filtered upstream, we should never have a case where the iterator returns an index of 1. Note that it's OK to start with a counter_index of 1 as long as the counter_mask doesn't select the bit (so say something like counter_index == 1 && counter_mask = 0x10 is fine). Since we already check the same for skip_match above, maybe we should just remove the check?

    ```
/// Returns a filtered counter_mask if the PMU counter range can be configured.
    pub fn get_configurable_counter_range(
        &self,
        counter_index: u64,
        counter_mask: u64,
        config_flags: PmuCounterConfigFlags,
    ) -> SbiResult<u64> {
        use PmuCounterState::*;
        let pmu_info = pmu::PmuInfo::get()?;
        let counter_mask = pmu_info.filter_counter_mask(counter_index, counter_mask)?;
        if !config_flags.is_skip_match() {
            let mut bitmask_iter = CounterMaskIter::new(counter_index, counter_mask);
            bitmask_iter
                .find(|i| matches!(self.counter_state[*i], NotConfigured))
                .map_or_else(|| Err(SbiError::InvalidParam), |_| Ok(counter_mask))
        } else {
            // If skip_match is set, the counter must already be configured
            let state = self
                .counter_state
                .get(counter_index as usize)
                .ok_or(SbiError::InvalidParam)?;
            if matches!(state, Started(_) | Configured(_)) {
                Ok(counter_mask)
            } else {
                Err(SbiError::InvalidParam)
            }
        }
    }
```

File: src/vm_pmu.rs (hunk 16 above)
Comment (atulkharerivos) [reply src/vm_pmu.rs:445]: Yes, saw that in the SBI code, and per the chat with Atish, he suggested forgoing it. @atishp04: Can you please confirm?

File: drivers/src/pmu.rs [hunk 19]
```diff
@@ -50,19 +53,62 @@ impl PmuInfo {
         Ok(ControlFlow::Continue(()))
     }
 
+    fn init_delegated_counters(&mut self) -> SbiResult<()> {
+        use riscv_regs::CSR;
+        // CYCLE and INSRET must be delegated
+        const MANDATORY_DELEGATED_COUNTERS_MASK: u64 = 0x5;
+        let delegated_counters = CSR.smcounteren.get();
+        if (delegated_counters & MANDATORY_DELEGATED_COUNTERS_MASK) == 0 {
```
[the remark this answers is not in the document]
atulkharerivos: Yes...good catch. I think it's a good idea to put in a sanity check for bit 1 == 0 as well.

src/vm_pmu.rs
atishp04: Yes. The PMU sig is yet to decide event standardization and mapping methods. There are few options:
1. Allow All-to-ALL mapping which means any programmable events can be monitored using any programmable counter
similar to ARM64. 
3. Use DT to provide the mapping (similar to OpenSBI DT PMU bindings) - The downside is to define the ACPI tables as well.
4. Rely on event data provided in json from perf framework. Similar to x86. Needs more support in Linux but needs even more parsing support for other projects such as Salus.

That's why, we are pushing for option 1 as that would be the ideal for software. That's why, I suggested to use #1 for now. Once PMU sig finalized one of the approaches. We can update if necessary.

src/vm_pmu.rs
abrestic-rivos: Yeah ok, if this is already accounted for then I'd suggest removing the check. It's confusing to have it one branch but not the other.

src/vm_pmu.rs
```diff
@@ -368,4 +373,149 @@ impl VmPmuState {
...
+            }
+
+            let pmu_info = pmu::PmuInfo::get()?;
+            let counter_mask = pmu_info.filter_counter_mask(counter_index, counter_mask)?;
+            let index = if !config_flags.is_skip_match() {
+                let mut bitmask_iter = CounterMaskIter::new(counter_index, counter_mask);
+                bitmask_iter
+                    .find(|i| {
+                        matches!(self.counter_state[*i], NotConfigured) || is_fixed_counter(*i)
+                    })
```
abrestic-rivos, line 420: Hm, is this match condition right? As written, we'd always select a fixed counter index if it were included in the counter mask, even if the requested event wasn't a fixed counter event. Shouldn't we only be picking a fixed counter if the user requested their respective events? e.g. always pick counter 0 if user requested the `cycles` event.

atulkharerivos: Since the iterator filters elements based on the counter_mask, find() should iterate only over those elements, meaning that we'll hit the condition only if a fixed counter was selected by both the counter_index and counter_mask, right?

```
// Convenience iterator to return counter_state indexes relative to counter_index
// if the counter_mask bit is set (assumes sanitized input).
impl Iterator for CounterMaskIter {
    type Item = usize;
    fn next(&mut self) -> Option<Self::Item> {
        loop {
            if self.counter_mask == 0 {
                break None;
            }
            let result = self.counter_index + self.mask_index;
            let mask_bit_set = self.counter_mask & (1 << self.mask_index);
            self.counter_mask &= !(1 << self.mask_index);
            self.mask_index += 1;
            if mask_bit_set != 0 {
                break Some(result as usize);
            }
        }
    }
}
```

src/vm_pmu.rs
atishp04: In addition to setting SINH, and VSINH
We also need to other *INH flags. If VM sets a UINH, VUINH also should be set.

src/vm_pmu.rs
abrestic-rivos: Yes, but I'm suggesting that we should only be selecting a fixed counter if the requested event type matches.

drivers/src/pmu.rs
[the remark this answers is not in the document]
atishp04: The spec doesn't say anything about mandatory delegation. Why does Salus needs to enforce that ?
We do it due to backward compatibility and ABI issues today but Salus should just print an warning message and continue. I don't think it should return an error message.

src/vm_pmu.rs
atulkharerivos: Ah, definitely makes sense.

drivers/src/pmu.rs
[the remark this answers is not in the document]
atulkharerivos: Ack.

drivers/src/pmu.rs
```diff
@@ -50,19 +53,62 @@ impl PmuInfo {
...
     /// Initializes the global PmuInfo structure with the information obtained from the platform SBI.
     pub fn init() -> SbiResult<()> {
         let mut pmu_info = PmuInfo::new();
-        let num_counters = pmu::get_num_counters()?;
-        for i in 0..num_counters.min(MAX_HARDWARE_COUNTERS as u64) {
-            if matches!(pmu_info.init_counter(i)?, ControlFlow::Break(())) {
-                break;
+        let cpu_id = CpuInfo::get();
+        if !cpu_id.has_sscdeleg() {
+            let num_counters = pmu::get_num_counters()?;
```
atishp04, line 85: IMO, it would be nice to split into different functions fro code quality perspective. It makes it easier to read through.
Something like this:

get_num_counters_sbi() {
}

get_delegated_counters() {

}

e.g.
pmu::get_num_counters() {
     if (cpu_id.has_sscdeleg())
           get_delegated_counters()
      else
           get_num_counters_sbi()
}

Same goes for stopping counters below and all other counter related ops.

drivers/src/pmu.rs
atishp04: Yup. For delegated it is identity mapped.

src/vm_pmu.rs
atishp04: Do the scounterselect right before it, so that it is easier to follow all the scounterselect modifications.

src/vm_pmu.rs
```diff
@@ -368,4 +373,149 @@ impl VmPmuState {
...
+            let pmu_info = pmu::PmuInfo::get()?;
+            let counter_mask = pmu_info.filter_counter_mask(counter_index, counter_mask)?;
+            let bitmask_iter = CounterMaskIter::new(counter_index, counter_mask);
+            for index in bitmask_iter {
+                if matches!(self.counter_state[index], Started(_)) {
+                    let index = index as u64;
+                    CSR.scountinhibit.read_and_set_bits(1 << index);
+                    if stop_flags.is_reset_flag() {
+                        CSR.scounterselect.set(index);
+                        CSR.shpmcounter.set(0);
```
atishp04, line 514: SBI_PMU_STOP_FLAG_RESET only resets the mapping doesn't reset the counter value.
shpmevent should be only reset.

atulkharerivos: Ack...I had assumed that it reset the counter as well...will change.

src/vm_pmu.rs
atulkharerivos: Changed...given the conversation about the preferred approach, there's no need for us to treat fixed counters differently. If the fixed counter is selected, and is available, it will be automatically picked since it'll be a lower index. if not, we'll simply choose one of the available configurable counters, and the assumption is that the platform can use it for fixed counters as well.

atulkharerivos: Changed to ensure static mapping for fixed counters per offline chat with @atishp04

src/vm_pmu.rs
```diff
@@ -368,4 +375,167 @@ impl VmPmuState {
...
+            let index = if !config_flags.is_skip_match() {
+                // Fixed counters (0, 2) must use a static mapping pending finalization of event mapping
+                if let Some(index) = CounterMaskIter::new(counter_index, counter_mask)
+                    .into_iter()
+                    .next()
+                {
+                    if index <= 2 && !matches!(self.counter_state[index], NotConfigured) {
+                        return Err(SbiError::InvalidParam);
+                    }
+                }
```
abrestic-rivos, line 417: I'm not quite following this logic. Looks like we're just enforcing that if a fixed counter is in `counter_mask` that it be unconfigured, and thus will guaranteed to be picked below on line 419. I thought the plan was to always pick counter 0 if the event was `cycles` and counter 2 if the event was `instret`?

atulkharerivos: As I see it, there's no canonical way of picking a counter because it's a combination of the starting counter index and the counter_mask. 

For example:

A counter_index of 0 and a mask of 0100b => find() returns counter 2
A counter index of 1 and a mask of 0010b => find() returns counter 2
A counter index of 2 and a mask of 0001b => find() returns counter 2
A counter index of 0 and a mask of xxx1b => find() returns counter 0

In each of the above cases, next() returns the same element as find(), and we enforce that it's not configured. @atishp04: Any comments on the above?

atulkharerivos: Per offline conversation with @atishp04: First filter by event_type, and if CYCLES / INSRET, confirm that they are selected by the index / mask and are not configured. If OK, return a fixed counter, else error out early

src/vm_pmu.rs
```diff
@@ -368,4 +375,167 @@ impl VmPmuState {
...
+        start_flags: PmuCounterStartFlags,
+        initial_value: u64,
+    ) -> SbiResult<()> {
+        if !CpuInfo::get().has_sscdeleg() {
+            sbi::api::pmu::start_counters(counter_index, counter_mask, start_flags, initial_value)
+        } else {
+            let pmu_info = pmu::PmuInfo::get()?;
+            let counter_mask = pmu_info.filter_counter_mask(counter_index, counter_mask)?;
+            let bitmask_iter = CounterMaskIter::new(counter_index, counter_mask);
+            let mut clear_inhibit_bits = 0u64;
```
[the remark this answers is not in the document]
atulkharerivos: Changed.

src/vm_pmu.rs
```diff
@@ -368,4 +375,167 @@ impl VmPmuState {
...
+                shpmevent.modify(shpmevent::vsinh.val(1));
+            }
+
+            if config_flags.is_uinh() {
+                shpmevent.modify(shpmevent::uinh.val(1));
+            }
+
+            if config_flags.is_sinh() {
+                shpmevent.modify(shpmevent::sinh.val(1));
+            }
```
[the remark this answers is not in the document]
atulkharerivos: Since sinh is set by Salus, we can process it as is, but we can translate uinh => vuinh, and ignore the reset.

src/vm_pmu.rs
atulkharerivos: Also, skip_match takes precedence, so changed accordingly.

src/vm_pmu.rs
```diff
@@ -368,4 +375,195 @@ impl VmPmuState {
...
+                event_data: u64,
+                counter_index: u64,
+            ) {
+                let mut shpmevent = LocalRegisterCopy::<u64, shpmevent::Register>::new(0);
+                // Ignore vuinh / vsinh from the VM and translate uinh
+                if config_flags.is_uinh() {
+                    shpmevent.modify(shpmevent::vuinh.val(1));
+                }
+                if config_flags.is_sinh() {
+                    shpmevent.modify(shpmevent::sinh.val(1));
```
abrestic-rivos, line 410: This should be `vsinh`. And we should always be setting `sinh` to inhibit counting in Salus.

src/vm_pmu.rs
```diff
@@ -368,4 +375,195 @@ impl VmPmuState {
...
+                }
+                match event_type {
+                    PmuEventType::Hardware(_) | PmuEventType::Cache(_) => {
+                        shpmevent.modify(shpmevent::event_selector.val(event_type.raw()))
+                    }
```
abrestic-rivos, line 415: This should have a TODO about event type to raw event selector mapping.

src/vm_pmu.rs
```diff
@@ -368,4 +375,195 @@ impl VmPmuState {
...
+                            .counter_state
+                            .get(index)
+                            .map_or(false, |counter_state| {
+                                matches!(counter_state, PmuCounterState::NotConfigured)
+                            });
+                        // Fixed counters must use a static mapping pending finalization of event mapping
+                        match hardware_event_type {
+                            PmuHardware::CpuCycles => {
+                                if is_unconfigured && index == PmuHardware::CpuCycles as usize {
+                                    let counter_index = PmuHardware::CpuCycles as u64;
```
abrestic-rivos, line 450: The event ID is not the same as the counter index. The fixed `cycles` counter is counter 0, while the `cycles` event ID is 1. (They do happen to be the same for `instret`, though.)

src/vm_pmu.rs
```diff
@@ -368,4 +375,195 @@ impl VmPmuState {
...
+                }
+                if config_flags.is_auto_start() {
+                    CSR.scountinhibit.read_and_clear_bits(1 << counter_index);
+                }
+            }
+
+            let pmu_info = pmu::PmuInfo::get()?;
+            let counter_mask = pmu_info.filter_counter_mask(counter_index, counter_mask)?;
+            let index = if !config_flags.is_skip_match() {
+                match event_type {
```
abrestic-rivos, line 434: I think it might be more readable to write this logic like this:

```
let index = if !config_flags.is_skip_match() {
    let iter = CounterMaskIter::new(...);
    match event_type {
        PmuHardware::CpuCycles => {
            if iter.find(|c| c == CYCLES).is_some() && matches!(self.counter_state[CYCLES], NotConfigured) {
                CYCLES
            } else {
                return Err(...);
            }
        }
        PmuHardware::Instructions => { ... /* Same as above but s/CYCLES/INSTRET/ */ }
        _ => iter.find(...).ok_or(...),
    }
} else {
    ...
}?;
...
```

src/vm_pmu.rs
```diff
@@ -368,4 +375,195 @@ impl VmPmuState {
...
+                    .get(counter_index as usize)
+                    .ok_or(SbiError::InvalidParam)?;
+                if matches!(state, Started(_) | Configured(_)) {
+                    Ok(counter_index)
+                } else {
+                    Err(SbiError::InvalidParam)?
+                }
+            }?;
+
+            configure_counter_csrs(config_flags, event_type, event_data, index);
```
abrestic-rivos, line 500: We shouldn't be re-configuring the event type in the `skip_match` case.

src/vm_pmu.rs
```diff
@@ -365,7 +368,110 @@ impl VmPmuState {
...
+    ) -> SbiResult<u64> {
+        use sbi::PmuHardware::*;
+        fn configure_counter_csrs(
+            config_flags: PmuCounterConfigFlags,
+            event_type: PmuEventType,
+            event_data: u64,
+            counter_index: u64,
+        ) {
+            let mut shpmevent = LocalRegisterCopy::<u64, shpmevent::Register>::new(0);
+            // Ignore vuinh from the VM and translate uinh/sinh
```
atishp04, line 387: We shouldn't ignore vuinh/vsinh as we are already running a TVM  :).

src/vm_pmu.rs
atulkharerivos: Changed. As discussed offline, Salus was unconditionally setting sinh upstream, but it's been modified now.

src/vm_pmu.rs
atulkharerivos: Changed....PTAL.

src/vm_pmu.rs
atulkharerivos: Seems like OpenSBI permits re-configuration in the skip match case per my reading of the code (see lib/sbi/sbi_pmu.c:695):

`skip_match:
	if (event_type == SBI_PMU_EVENT_TYPE_HW) {
		if (flags & SBI_PMU_CFG_FLAG_CLEAR_VALUE)
			pmu_ctr_write_hw(ctr_idx, 0);`

src/vm_pmu.rs
atulkharerivos: Aggh...good catch....got them mixed up somehow....fixed.

src/vm_pmu.rs
atishp04: configure_counter_csrs here updates the shpmevent as well which is reconfiguring the event type. OpenSBI only clears the counter value if the SBI_PMU_CFG_FLAG_CLEAR_VALUE is set.

SBI_PMU_CFG_FLAG_SKIP_MATCH - skips the counter matching part only. But implementation should honor the counter value update flags.

drivers/src/pmu.rs
```diff
@@ -50,19 +52,61 @@ impl PmuInfo {
...
-    pub fn init() -> SbiResult<()> {
-        let mut pmu_info = PmuInfo::new();
+    fn check_delegated_counters() -> SbiResult<u64> {
+        const TIMER_COUNTER_MASK: u64 = 0x2;
+        let delegated_counters = CSR.smcounteren.get();
+
+        // TIMER can never be valid
+        if (delegated_counters & TIMER_COUNTER_MASK) != 0 {
+            return Err(SbiError::NotSupported);
+        }
```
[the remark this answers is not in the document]
atishp04: As per the spec this implementation is wrong also.

When HPMDE=1, smcounteren is shadow of the mcounteren register, otherwise reads return 0. TM bit will always be set.
If a vm request a write to TIME counter (shpmcounter[1] = 1) via any SBI call, it can return an SBI error though.

Any writes to TIME shpmcounter[1] should be ignored in hardware as per the spec if it is delegated.

src/vm_pmu.rs
```diff
@@ -365,7 +368,112 @@ impl VmPmuState {
...
+            counter_index: u64,
+        ) {
+            let mut shpmevent = LocalRegisterCopy::<u64, shpmevent::Register>::new(0);
+            // Translate uinh/sinh
+            if config_flags.is_uinh() || config_flags.is_vuinh() {
+                shpmevent.modify(shpmevent::vuinh.val(1));
+            }
+            if config_flags.is_sinh() || config_flags.is_vsinh() {
+                shpmevent.modify(shpmevent::vsinh.val(1));
+            }
```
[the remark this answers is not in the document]
atishp04: yeah you are correct. We should ignore VSINH and VUINH.

src/vm_pmu.rs
```diff
@@ -365,7 +368,112 @@ impl VmPmuState {
...
+                        const CYCLE: u64 = 0;
+                        CYCLE
+                    } else {
+                        const INSRET: u64 = 2;
+                        INSRET
+                    };
+                    if iter.any(|index| {
+                        index == platform_index as usize
+                            && matches!(self.counter_state[index], PmuCounterState::NotConfigured)
+                            && pmu_info.get_counter_info(platform_index).is_ok()
```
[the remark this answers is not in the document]
atishp04: May be I was not clear in early attempts. Without delegation, any counter can not be accessed/updated. So that's always my implicit assumption. As said in the above comment, 
Statement from the counter delegation spec
"When HPMDE=1, smcounteren is shadow of the mcounteren register, otherwise reads return 0."

and priv spec
"When the CY, TM, IR, or HPMn bit in the mcounteren register is clear, attempts to read the
cycle, time, instret, or hpmcountern register while executing in S-mode or U-mode will cause
an illegal instruction exception. When one of these bits is set, access to the corresponding register
is permitted in the next implemented privilege mode (S-mode if implemented, otherwise U-mode)."

Thus, we shouldn't skip the delegation check at any point. 

TLDR; we should pick option "a" pointed by @abrestic-rivos

src/vm_pmu.rs
```diff
@@ -374,16 +482,26 @@ impl VmPmuState {
...
-            counter_mask,
-            config_flags,
-            event_type,
-            event_data,
-        )?;
+        let platform_counter_index = if !CpuInfo::get().has_sscdeleg() {
+            sbi::api::pmu::configure_matching_counters(
+                counter_index,
+                counter_mask,
+                config_flags.set_sinh().set_minh(),
```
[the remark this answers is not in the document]
atishp04: Oops. We missed that during earlier reviews.

src/vm_pmu.rs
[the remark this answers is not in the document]
atulkharerivos: FWIW, we were enforcing the CYCLE/INSRET delegation check before if was removed because it wasn't outlined in the specification. We should update it in the spec as well.

src/vm_pmu.rs
```diff
@@ -365,7 +368,111 @@ impl VmPmuState {
...
+                CSR.shpmcounter.set(0);
+            }
+            if config_flags.is_auto_start() {
+                CSR.scountinhibit.read_and_clear_bits(1 << counter_index);
+            }
+        }
+
+        let pmu_info = pmu::PmuInfo::get()?;
+        let counter_mask = pmu_info.filter_counter_mask(counter_index, counter_mask)?;
+        let index = if !config_flags.is_skip_match() {
```
abrestic-rivos, line 420: The logic in this block can be a bit hard to follow. I think it might be more readable if written like this:

```
let index = if config_flags.is_skip_match() {
    self.counter_state
        .get(counter_index as usize)
        .filter(|s| matches!(s, Started(_) | Configured(_)))
        .ok_or(SbiError::InvalidParam)
} else {
    let fixed_index = match event_type {
        PmuEventType::Hardware(CpuCycles) => Some(0),
        PmuEventType::Hardware(Instructions) => Some(2),
        PmuEventType::Firmware(_) => {
            return Err(SbiError::NotSupported);
        }
        _ => None,
    };
    let mut iter = CounterMaskIter::new(counter_index, counter_mask);
    iter.filter(|i| fixed_index.is_none() || fixed_index == Some(i))
        .find(|i| matches!(self.counter_state[i], NotConfigured))
        .ok_or(SbiError::NotSupported)
}?;
configure_counter_csrs(...);
```

abrestic-rivos (commented):
Thanks, I think this looks correct. Mostly style comments left. BTW, I assume you need QEMU & OpenSBI patches for Sscdeleg to be usable? Can you include pointers to those branches in the PR description or one of the commit messages?

drivers/src/pmu.rs
```diff
@@ -50,19 +52,50 @@ impl PmuInfo {
...
-            if matches!(pmu_info.init_counter(i)?, ControlFlow::Break(())) {
+            if matches!(self.init_sbi_counter(i)?, ControlFlow::Break(())) {
                 break;
             }
         }
+        Ok(())
+    }
+
+    /// Initializes the global PmuInfo structure with the information obtained from the platform SBI.
+    pub fn init() -> SbiResult<()> {
```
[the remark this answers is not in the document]
atulkharerivos: Ack...will submit a separate follow-up PR instead of cluttering this one with it.

drivers/src/pmu.rs
```diff
@@ -50,19 +52,50 @@ impl PmuInfo {
...
         for i in 0..num_counters.min(MAX_HARDWARE_COUNTERS as u64) {
-            if matches!(pmu_info.init_counter(i)?, ControlFlow::Break(())) {
+            if matches!(self.init_sbi_counter(i)?, ControlFlow::Break(())) {
                 break;
             }
         }
+        Ok(())
+    }
+
+    /// Initializes the global PmuInfo structure with the information obtained from the platform SBI.
```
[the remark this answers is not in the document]
atulkharerivos: Comment has been updated.

drivers/src/pmu.rs
```diff
@@ -50,19 +52,50 @@ impl PmuInfo {
         Ok(ControlFlow::Continue(()))
     }
 
-    /// Initializes the global PmuInfo structure with the information obtained from the platform SBI.
-    pub fn init() -> SbiResult<()> {
-        let mut pmu_info = PmuInfo::new();
+    fn get_delegated_counters(&mut self) -> SbiResult<()> {
+        let delegated_counters = CSR.smcounteren.get();
+        // Always ignore index 1 (TIME)
+        self.valid_hw_counters_mask = delegated_counters & !0x2;
```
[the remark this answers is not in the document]
atulkharerivos: Changed, but still need to invert the mask: `delegated_counters & !(1 << (CSR_TIME - CSR_CYCLE)`. Sorry, @atishp04, they decided to use "!" instead of "~" in Rust for bit-wise negation :).

src/vm_pmu.rs
atulkharerivos: Nice use of of the Some / None trick. Also, good use of the non-iterator `.filter()` -- I likely read about it in the passing, but this is a great use case.

abrestic-rivos (approved):

PR #190, opened 2022-12-23
Forward debug console to host

### LINKED ISSUE #146: Forward guest VM SBI console output to host

abrestic-rivos:
Forward debug console ECALLs (DBCN and legacy `put_char`) to a VM's host instead of handling console output directly within Salus. For guest TVMs, this means that that buffer to be printed must reside in memory shared with the host VM.

- Patches 1&2 prepare for console forwarding by putting the DBCN buffer in shared memory.
- Patches 3&4 forward the DBCN and `put_char` calls to the host, respectively.
- Patch 5 forwards the return value from the host back to the guest for these ECALLs.

Looking for opinions on the forwarding of ECALL return values. Currently I'm just using A0/A1 from the shared memory area, which means the host will have to know which ECALLs for which it's expected to provide a return value. Other options I considered are: setting a flag (in shared memory?) that indicates to the host that it needs to provide a return value, or having a separate `sbi_return` struct in the shared memory area where the host puts the return value (with sentinel values to indicate whether or not the TSM is expecting to find something there).

Fixes #146

### DIFF
--- a/s-mode-utils/src/sbi_console.rs
+++ b/s-mode-utils/src/sbi_console.rs
@@ -4,27 +4,40 @@
 
 use sbi::api::debug_console::console_puts;
 use sbi::SbiMessage;
+use spin::{Mutex, Once};
 
 use crate::ecall::ecall_send;
 use crate::print::{Console, ConsoleWriter};
 
 /// Driver for an SBI based console.
-pub struct SbiConsole {}
+pub struct SbiConsole {
+    buffer: Mutex<&'static mut [u8]>,
+}
 
-static SBI_CONSOLE: SbiConsole = SbiConsole {};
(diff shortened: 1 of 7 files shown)


atulkharerivos (approved):
LGTM.

dgreid (approved):
Would be nice to drop legacy console soon so we can stop dealing with it.

For return values, I feel a0/a1 is fine, it's simple and VMMs will already have to support that pattern for non-TEE use cases.

LGTM, thanks.

PR #191, opened 2022-12-25
VmCpu: Don't allow setting Zero GPR.

rajnesh-kanwal:
It's possible that Zero GPR be used as destination register for some instructions. For example the assembler pseudoinstruction to write a CSR, CSRW csr, rs1, is encoded as CSRRW x0, csr, rs1. While emulating a csr we can end up updating the x0 gpr. This gets problematic when emulating the CSRR instruction as it is encoded as CSRRS rd, csr, x0. Here we will end up setting bits in csr from x0 written from last CSRW emulation giving us undefined behavior.

### DIFF
--- a/riscv-regs/src/regs.rs
+++ b/riscv-regs/src/regs.rs
@@ -101,6 +101,10 @@ impl GeneralPurposeRegisters {
 
     /// Sets the value of the given register.
     pub fn set_reg(&mut self, reg_index: GprIndex, val: u64) {
+        if reg_index == GprIndex::Zero {
+            return;
+        }
+
         self.0[reg_index as usize] = val;
     }
 


src/vm_cpu.rs
```diff
@@ -869,6 +869,10 @@ impl<'vcpu, 'pages, 'host, T: GuestStagePagingMode> ActiveVmCpu<'vcpu, 'pages, '
 
     /// Sets one of the vCPU's general-purpose registers.
     pub fn set_gpr(&mut self, gpr: GprIndex, value: u64) {
+        if gpr == GprIndex::Zero {
```
glg-rv, line 872: Personally, I would put this on the `GeneralPurposeRegisters` methods (`set_reg`, or even returning zero on `reg` for zero) itself. Afterall AFAIU the zero register should be zero always. So preventing it to be set at the type structure would just avoid looking like a corner special case.

rajnesh-kanwal: Makes sense. :+1: Thanks. Fixed.

riscv-regs/src/regs.rs
```diff
@@ -101,6 +101,10 @@ impl GeneralPurposeRegisters {
 
     /// Sets the value of the given register.
     pub fn set_reg(&mut self, reg_index: GprIndex, val: u64) {
+        if reg_index == GprIndex::Zero {
+            return
```
glg-rv, line 105: This is still failing `cargo fmt` checks. `make format` is a useful check for this.

Rust police says you have to add a `;` after return.

rajnesh-kanwal: Nice catch. Thanks. My rust got rusty in just 2 weeks of C work.
Fixed.

abrestic-rivos (approved):
Good catch!

PR #194, opened 2022-12-30
Add support to remove an emulated MMIO region.

rajnesh-kanwal:


sbi/src/tee_guest.rs
```diff
@@ -22,6 +22,18 @@ pub enum TeeGuestFunction {
         /// a1 = length of the region
         len: u64,
     },
+    /// Removes the specified range of guest physical address space from the emulated MMIO regions. Upon
```
rajnesh-kanwal, line 25: This breaks the backwards compatibility but given we are still in the development phase I thought this should be fine. This is to mainly keep `AddMmioRegion` and `RemoveMmioRegion` together.

src/vm_pages.rs
```diff
@@ -315,6 +315,58 @@ impl VmRegionList {
...
+        let mut index = self
+            .regions
+            .iter()
+            .enumerate()
+            .find(|(_, r)| r.start <= start && r.end >= end && r.region_type == region_type)
+            .map(|(i, _)| i)
+            .ok_or(Error::InvalidMapRegion)?;
+
+        // Unwrap ok here since we know 'index' is valid.
+        let region = self.regions.get(index).unwrap();
```
abrestic-rivos, line 334: Instead of doing a `self.regions.get(index)` here, can we get a reference to the region in the `find()` call above?

e.g. ```let (index, region) = self.regions.iter().enumerate().find(...).ok_or(...)?;```

src/vm_pages.rs
```diff
@@ -315,6 +315,58 @@ impl VmRegionList {
...
+        // Add the remaining range at the start as a new region.
+        if r_start < start {
+            let new_region = VmRegion {
+                start: r_start,
+                end: start,
+                region_type,
+            };
+            self.regions
+                .try_insert(index, new_region)
+                .map_err(|_| Error::InsufficientVmRegionSpace)?;
```
abrestic-rivos, line 350: Need to check that we have space for the new region before we go to remove the old one, otherwise we could leave the list in an inconsistent state.

abrestic-rivos (commented):
Just a couple minor comments, otherwise LGTM -- thanks!

src/vm_pages.rs
rajnesh-kanwal: Nice catch. Thanks.
Fixed.

src/vm_pages.rs
rajnesh-kanwal: Fixed.

PR #195, opened 2023-01-04
Add bug fix for PMU counter logic

atulkharerivos:
A recent refactoring of the PMU counter logic had introduced a latent bug that was exposed by a code change that introduced an additional context switch between counter configuration and start. The gist is that on the outbound path, the code change was causing the internal bookkeeping to revert it back to the non-configured state, meaning that the subsequent call to start the counter would fail.

This fixes the issue by introducing helper functions that help distinguish between the internal code paths required required for context switch operations, and SBI calls initiated by the VM.

### DIFF
--- a/src/vm_pmu.rs
+++ b/src/vm_pmu.rs
@@ -266,7 +266,7 @@ impl VmPmuState {
 
     fn reset_all_counters(&mut self, counter_mask: u64) -> SbiResult<()> {
         if counter_mask != 0 {
-            self.stop_counters(
+            self.stop_counters_internal(
                 0,
                 counter_mask,
                 PmuCounterStopFlags::default().set_reset_flag(),
@@ -313,7 +313,7 @@ impl VmPmuState {
 
     fn resume_counter(&mut self, counter_index: u64, c: &CounterState) -> SbiResult<()> {
         let start_flags = PmuCounterStartFlags::default().set_init_value();
-        self.start_counters(counter_index, 0x1, start_flags, c.value)
+        self.start_counters_internal(counter_index, 0x1, start_flags, c.value)
             .map(|_| VmPmuState::set_hcounteren_bit(counter_index))
     }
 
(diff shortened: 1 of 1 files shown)


src/vm_pmu.rs
```diff
@@ -365,14 +362,14 @@ impl VmPmuState {
...
-    /// Calls the SBI configure_matching_counters() and performs internal bookkeeping on counter state.
-    pub fn configure_matching_counters(
+    fn configure_matching_counters_internal(
         &mut self,
         counter_index: u64,
         counter_mask: u64,
         config_flags: PmuCounterConfigFlags,
         event_type: PmuEventType,
         event_data: u64,
+        is_context_switch: bool,
```
abrestic-rivos, line 372: All this argument is used for is to set another flag in `config_flags`. How about doing that at the callsite instead?

src/vm_pmu.rs
```diff
@@ -419,21 +440,37 @@ impl VmPmuState {
...
     }
 
-    /// Calls the SBI stop_counters() and performs internal bookkeeping on counter state.
-    pub fn stop_counters(
+    fn stop_counters_internal(
         &mut self,
         counter_index: u64,
         counter_mask: u64,
         stop_flags: PmuCounterStopFlags,
+        is_context_switch: bool,
```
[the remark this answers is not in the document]
atulkharerivos: We need to tell apart the stop_flags passed in by SBI apart from the extra flags set by the context switch code. I previously had an Option<Flags> with the additional flags, so perhaps that's better?

src/vm_pmu.rs
atulkharerivos: The issue happened (after refactoring) because the call-site set `PmuCounterStopFlags::default().set_reset_flag()`, and there was no way to tell it apart from the flags passed in by the VM. Note that it wasn't an issue before because the bookkeeping code wasn't part of the same function, so the extra flags were visible only to the SBI layer. However, it required the caller to make follow-up calls for bookkeeping and so forth, and having the callee handle everything is a better approach.

atulkharerivos: Separated the bookkeeping code from the context switch path per offline discussion.

abrestic-rivos (approved):

PR #197, opened 2023-01-06
virtual memory: Extend use and support for huge pages

stillson:
salus: virtual memory
Condense stretched of memory into huge pages

Add a function to break up regions being mapped into the minimal number of pages using huge pages. Apply it to selected memory regions (currently hypervisor heap and hypervisor per cpu)

salus:heap memory
guarantee that the heap will fit into 2m pages in virtual memory.

When getting the memory region for heap memory, guarantee that it starts on a 2m boundary and is a multiple of 2m long to guarantee that the heap can be broken up into huge pages.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-t<!--
+<!--
 SPDX-FileCopyrightText: 2023 Rivos Inc.
 
 SPDX-License-Identifier: Apache-2.0


src/hyp_map.rs
```diff
@@ -70,18 +72,122 @@ impl SharedRegion {
...
+                variable_mapped,
+                page_size: None,
             })
         } else {
             None
         }
     }
 
-    // Map this region into a page table.
+    /// partition a HypMapRegion to use minimal pages
```
dgreid, line 93: I'm slow and having trouble parsing this function. Could you provide an overview of what's going on? Is it finding the biggest page size that fits, using that until exhausted, then using smaller ones?

riscv-pages/src/page.rs
```diff
@@ -88,6 +88,54 @@ impl PageSize {
...
+        match self {
+            PageSize::Size512G => Some(PageSize::Size1G),
+            PageSize::Size1G => Some(PageSize::Size2M),
+            PageSize::Size2M => Some(PageSize::Size4k),
+            PageSize::Size4k => None,
+        }
+    }
+
+    /// Iterate up one page size
+    pub fn up_size(&self) -> PageSize {
```
abrestic-rivos, line 114: This should probably return an `Option<>` as well, with `up_size(Size512G)` returning `None`.

src/hyp_map.rs
```diff
@@ -70,18 +72,122 @@ impl SharedRegion {
             let vaddr = PageAddr::new(RawAddr::supervisor_virt(r.base().bits())).unwrap();
             let page_count = PageSize::num_4k_pages(r.size()) as usize;
             let pte_fields = PteFieldBits::leaf_with_perms(pte_perms);
+            let variable_mapped = matches!(
+                r.region_type(),
+                HwMemRegionType::Reserved(HwReservedMemType::HypervisorHeap)
+                    | HwMemRegionType::Reserved(HwReservedMemType::HypervisorPerCpu)
+            );
```
abrestic-rivos, line 79: I think `variable_mapped` is meant to imply that the region can be mapped with huge-pages (see also Dylan's comment above). If that's the case, why would we restrict using hugepages to just these regions of memory? They're relatively quite small; the main benefit we get from using hugepages is mapping the rest of available DRAM or PCI MMIO space with hugepages.

src/hyp_map.rs
```diff
@@ -260,11 +374,33 @@ impl HypMap {
...
+                        let i_start = i.paddr.bits();
+                        let i_end = i_start + i.page_size() as u64 * i.page_count as u64;
+                        assert!(i_start >= r_start && i_start < r_end);
+                        assert!(i_end > r_start && i_end <= r_end);
+                        i.map(&sv48, &mut || {
+                            hyp_mem
+                                .take_pages_for_hyp_state_with_size(1, i.page_size())
+                                .into_iter()
+                                .next()
+                        })
```
abrestic-rivos, line 396: The closure passed to `map()` is meant to retrieve a free page-table page, which are always 4kB.

src/hyp_map.rs
```diff
@@ -70,18 +72,122 @@ impl SharedRegion {
...
+                page_size: None,
             })
         } else {
             None
         }
     }
 
-    // Map this region into a page table.
+    /// partition a HypMapRegion to use minimal pages
+    pub fn partition(&self) -> Option<SharedRegionsVec> {
```
abrestic-rivos, line 94: This logic is pretty confusing to follow and there are a couple edge cases that I think aren't quite handled correctly (but need to read again more carefully).

First, I'd recommend doing this iteratively in `map()`; there's really no need to recursively create an intermediate vector of regions.

Next, I think a more concise (and easier to read) way to structure this would be something like:

```
fn select_page_size(va: u64, pa: u64, len: u64) -> PageSize {
    for size in [Size512G, Size1G, Size2m] {
        if size.is_aligned(va) && size.is_aligned(pa) && len >= (size as u64) {
            return size;
        }
    }

    return Size4k;
}

fn map(...) {
    let mut offset = 0;
    while offset != size {
        let remaining = size - offset;
        let va = self.vaddr() + offset;
        let pa = self.paddr() + offset;
        let page_size = select_page_size(va, pa,, remaining);
        let num_pages = remaining / (page_size as u64);
        let mapper = map_range(va, page_size, num_pages, get_pte_page);
        for (v, p) in va.iter_from_with_size(page_size).zip(...) {
            mapper.map_addr(...);
        }
        offset += page_size as u64 * num_pages;
    }
}
```

page-tracking/src/page_tracker.rs
```diff
@@ -649,7 +649,16 @@ impl HypPageAlloc {
```
abrestic-rivos, line 649: Same comment as the previous commit re: commit message.

src/hyp_map.rs
```diff
@@ -36,6 +36,8 @@ struct SharedRegion {
     paddr: PageAddr<SupervisorPhys>,
     page_count: usize,
     pte_fields: PteFieldBits,
+    variable_mapped: bool,
```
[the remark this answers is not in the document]
stillson: fixed

PR #198, opened 2023-01-06
Move SBI to its own project

dgreid:
This series switches salus to use the newly created sbi-rs repository instead of the local sbi/

The first patch fixes a logic error in the cfg and is the sha that was copied from to create the new repo

The following commits add the new remote, handle updating it from the Makefile, and finally switching to use it and removing the local copy.

The Makefile update might be overly ambitious. It's maximally convenient for building the code, but adds a foot gun when updating sbi-rs. The Makefile checks out the committed version when building, meaning any changes will need to be committed before being tested. What are people's thoughts on that?


abrestic-rivos (commented):
My $0.02: I'd prefer not to auto-update the submodule via the Makefile as it'll almost certainly trip me up, and I'd rather force myself to learn how to manage submodules the right way :)

Otherwise LGTM

dgreid:
> My $0.02: I'd prefer not to auto-update the submodule via the Makefile as it'll almost certainly trip me up, and I'd rather force myself to learn how to manage submodules the right way :)
> 

I can remove the auto update but keep the auto-fetch-if-empty, if that seems better. But removing and letting the developer run submodule update --init is cool with me too, that's pretty normal these days.

abrestic-rivos:
> I can remove the auto update but keep the auto-fetch-if-empty, if that seems better.

That works for me.

abrestic-rivos (approved):

PR #203, opened 2023-01-14
CI: Add qemu run to verification of new pull request

stillson:
Created an action to run tellus under salus under qemu and added to verification.


dgreid (commented):
Thanks! lgtm, just needs to be rebased on top of main again before merging as it has an already merged commit.

atulkharerivos (approved):

PR #204, opened 2023-01-17
s-mode-utils: Remove ecall_send implementation.

glg-rv:
There's already the same implementation in sbi-rs. Remove code from s-mode-utils and use the sbi-rs one.


glg-rv:
Test fails and this needs a sbi-rs change. Postponing to upcoming sbi-rs changes.

glg-rv:
Reopening as SBI-rs has been updated in the meanwhile

PR #205, opened 2023-01-18
attestation: check that guest-supplied indices do not overflow.

glg-rv:
In the ReadMeasurements and ExtendMeasurements guest calls salus is passed a PCR index. This index was then hard casted to u8 (from usize) to find the actual PCR.

While this doesn't cause any crash and it's not a serious bug, it can creates some confusion as it might extend the wrong PCR when a number > 255 is passed.


src/vm.rs
```diff
@@ -1624,9 +1624,14 @@ impl<'a, T: GuestStagePagingMode> FinalizedVm<'a, T> {
         &self,
         msmt_addr: u64,
         msmt_size: usize,
-        index: usize,
+        index_usize: usize,
```
sameo, line 1627: Nit picking, but can we keep the argument variable name (`index`) and call the internal one `index_u8`.

sameo (commented):
Good catch @glg-rv. Just some nitpicking otherwise LGTM.

glg-rv:
Commit updated attempting the rust path of implementing a TryInto<usize> and using that to get the TcgPcrIndex directly. Shorter and cleaner.

sameo (approved):

dgreid (approved):

PR #207, opened 2023-01-23
Update SBI main.

glg-rv:
SBI-rs has been updated with two PRs.

1. Vendor PR
2. nacl shared mem fix.

Originally I put the Vendor PR update in the umode mem mapping patch, but this would effectively block updating to new SBI-rs patches until that PR is merged.

Unblock this by updating it now, returning Unhandled on Vendor() in salus.

PR #208, opened 2023-01-23
Exercise multi-CPU TLB fencing behavior in Salus

mnissler-rivos:
Here are my first steps into multi-CPU testing for tellus.

@glg-rv @dgreid can you please take a look?

### DIFF
--- a/device-tree/src/fdt.rs
+++ b/device-tree/src/fdt.rs
@@ -8,7 +8,7 @@ use crate::DeviceTreeResult;
 use core::str;
 use fdt_rs::base::iters::{DevTreeNodeIter, DevTreeReserveEntryIter};
 use fdt_rs::base::parse::ParsedTok;
-use fdt_rs::base::{DevTree, DevTreeProp};
+use fdt_rs::base::{DevTree, DevTreeNode, DevTreeProp};
 use fdt_rs::prelude::*;
 
 /// Represents a flattened device-tree (FDT) as passed to the hypervisor by firmware. Currently
@@ -59,6 +59,18 @@ impl<'a> Fdt<'a> {
     /// Returns the range of memory where the host VM's initramfs is loaded, if present.
     pub fn host_initramfs_region(&self) -> Option<FdtMemoryRegion> {
         self.get_module_node_region("multiboot,ramdisk")
+    }
+
+    /// Returns an iterator over CPU nodes.
+    pub fn cpus<'b>(&'b self) -> impl Iterator<Item = Cpu<'b, 'a>> {
+        self.inner
+            .nodes()
+            .filter(|n| {
+                n.props()
+                    .any(|p| Ok(p.name()? == "device_type" && p.str()? == "cpu"))
+            })
+            .iterator()
+            .filter_map(|n| Some(Cpu { inner: n.ok()? }))
     }
 
     /// This returns the property buf for the first property with the give name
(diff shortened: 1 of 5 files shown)


test-workloads/src/bin/tellus.rs
```diff
@@ -300,6 +411,34 @@ extern "C" fn kernel_init(hart_id: u64, fdt_addr: u64) {
         mem_range.base(),
         mem_range.size()
     );
+    let mut next_page = (mem_range.base() + mem_range.size() / 2) & !0x3fff;
+
+    // Initialize CPUs.
+    PER_CPU.call_once(|| {
```
glg-rv, line 417: Hm? You set PER_CPU only when this block is executed and returns CPUs. But then you start the CPUs here, creating a race condition as the first thing the CPU does is calling PER_CPU.get().

I think this is shown in the test failures, perhaps (haven't checked) because QEMU is been running with the multi threaded accel, or we're just lucky on this instance.

test-workloads/src/bin/tellus.rs
```diff
@@ -57,6 +64,102 @@ pub fn alloc_error(_layout: Layout) -> ! {
...
+    }
+
+    fn init_tp(hart_id: u64) {
+        let my_tp = PER_CPU
+            .get()
+            .expect("PER_CPU not initialized")
+            .iter()
+            .position(|c| c.hart_id == hart_id)
+            .expect(
+                "missing
```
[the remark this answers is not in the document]
mnissler-rivos: Oops - fixing.

test-workloads/src/bin/tellus.rs
mnissler-rivos: Ah, good catch. No idea why I never ended up triggering this locally... will fix.

mnissler-rivos:
(Rebased on top of #207, so @glg-rv 's commit is now at the bottom of the stack to make checks work, please ignore that commit, I'll wait for #207 to land)

device-tree/src/fdt.rs
```diff
@@ -61,6 +61,11 @@ impl<'a> Fdt<'a> {
         self.get_module_node_region("multiboot,ramdisk")
     }
 
+    /// Returns an iterator over CPU nodes.
+    pub fn cpus(&self) -> CpuIter<'_, 'a> {
```
dgreid, line 65: Does this work?
```
    /// Returns an iterator over CPU nodes.
    pub fn cpus(&self) -> impl Iterator<Item = Cpu<'_, 'a>> {
        self.inner
            .nodes()
            .filter(|n| {
                n.props()
                    .any(|p| Ok(p.name()? == "device_type" && p.str()? == "cpu"))
            })
            .iterator()
            .map_while(|r| r.map(|n| Cpu { inner: n }).ok())
    }
```

Then the CpuIter below isn't needed. I realize the fallible -> normal iter conversion is messy and the other iters here don't set that example(but can, I think), so I'm OK either way.

dgreid (approved):
Thanks, the per-cpu runner is really nice to have. Just one suggestion.

glg-rv:
Haven't commented since my first review, but this LGTM. Keeping out of any rust-specific style discussion.

mnissler-rivos:
> Also `NCPU=2` from cmdline to overwrite Makefile default.

That reminds me that I should probably change the Makefile default as well.

device-tree/src/fdt.rs
mnissler-rivos: Ah, thanks for calling out the `iterator()` function! Replaced now with something along the lines of your suggestion.

PR #210, opened 2023-01-25
Vm: Deny TEEG ABI calls from host.

rajnesh-kanwal:
Given TEEG ABI calls are not supported for host, return error for these calls and also report EXT_TEE_GUEST to be not supported.

As of now if host does a TEEG call it leads to unhandled ecall and then system shutdown. This change also avoid this by returning unsupported ecall error to host.


atishp04 (approved):
LGTM

dgreid (approved):

PR #212, opened 2023-01-26
Topic/page addr unwraps

dgreid:
remove some unwraps that aren't needed. It's a start, and there is more to be done, but it's a move in the right direction.


glg-rv:
LGTM.

glg-rv (approved):
Looks like a step in the right direction for sure.

PR #217, opened 2023-01-27
Point SBI-rs to main again

glg-rv:
Joy of submodules.

Yesterday tellus PR inadvertedly changed SBI-rs off main, mostly removing FID fixes patch.

Point it back again to main.


mnissler-rivos (approved):
I find the GitHub UI confusing, but I think I managed to convince myself that it's pointing to the correct commit now.

glg-rv:
Yeah, the only way is to look at it through the raw patch. I think from now on we need an explicit explanation in PR on why we're updating SBI-rs. And check for it. It's very easy to miss.

dgreid:
I think this should move to 0d1037c68 instead. then the vm.rs patch isn't needed.

dgreid:
I'll post one.

PR #222, opened 2023-01-30
Introduce U-mode buffer.

glg-rv:
U-mode buffer is the mechanism used to pass hypervisor data to
U-mode. While guest pages are mapped directly into U-mode, hypervisor
pages should not be mapped.

A U-mode buffer is a small set of contiguous physical pages where
hypervisor data shared with U-mode are copied. It is mapped read-only
in U-mode.

There is single U-mode buffer per CPU. A U-mode buffer cannot be
mapped in multiple CPUs at the same time.

This PR says goodbye to cowsay, and rather test the buffer with a message sent from hypervisor to U-mode.

### DIFF
--- a/src/hyp_map.rs
+++ b/src/hyp_map.rs
@@ -308,20 +308,6 @@ impl HypPageTable {
         }
     }
 
-    /// Returns the virtual address of U-mode mapping slot `slot`.
-    pub fn umode_slot_va(&self, slot: UmodeSlotId) -> PageAddr<SupervisorVirt> {
-        let slot_num = match slot {
-            UmodeSlotId::A => 0,
-            UmodeSlotId::B => 1,
-        };
-        // UMODE_MAPPING_START and UMODE_MAPPING_SLOT_SIZE are required by static assertion to be
-        // page aligned, `round_down` will be a nop.
-        PageAddr::with_round_down(
-            RawAddr::supervisor_virt(UMODE_MAPPINGS_START + slot_num * UMODE_MAPPING_SLOT_SIZE),
-            PageSize::Size4k,
-        )
-    }
-
     /// Returns a mapper for U-mode slot `slot` for `num_pages` pages. If `writable` is true, the
     /// mapper will map pages User-writable, otherwhise will be mapped User-readable.
     pub fn umode_slot_mapper(
@@ -333,7 +319,7 @@ impl HypPageTable {
         if num_pages > PageSize::num_4k_pages(UMODE_MAPPING_SLOT_SIZE) {
             return Err(Error::OutOfMap);
         }
-        let vaddr = self.umode_slot_va(slot);
+        let vaddr = HypMap::umode_slot_va(slot);
         let mapper = self
             .sv48
             .map_range(vaddr, PageSize::Size4k, num_pages, &mut || {
@@ -359,7 +345,7 @@ impl HypPageTable {
         slot: UmodeSlotId,
         num_pages: u64,
     ) -> Result<impl Iterator<Item = SupervisorPageAddr> + '_, Error> {
-        let vaddr = self.umode_slot_va(slot);
+        let vaddr = HypMap::umode_slot_va(slot);
         if num_pages > PageSize::num_4k_pages(UMODE_MAPPING_SLOT_SIZE) {
             return Err(Error::OutOfMap);
         }
@@ -408,6 +394,20 @@ impl HypMap {
     // Returns an iterator for this Hypervisor private regions.
     fn private_regions(&self) -> impl Iterator<Item = &PrivateRegion> {
         self.private_regions.iter()
+    }
+
+    /// Returns the virtual address of U-mode mapping slot `slot`.
+    pub fn umode_slot_va(slot: UmodeSlotId) -> PageAddr<SupervisorVirt> {
+        let slot_num = match slot {
+            UmodeSlotId::A => 0,
+            UmodeSlotId::B => 1,
+        };
+        // UMODE_MAPPING_START and UMODE_MAPPING_SLOT_SIZE are required by static assertion to be
+        // page aligned, `round_down` will be a nop.
+        PageAddr::with_round_down(
+            RawAddr::supervisor_virt(UMODE_MAPPINGS_START + slot_num * UMODE_MAPPING_SLOT_SIZE),
+            PageSize::Size4k,
+        )
     }
 
     /// Creates a new page table based on this memory map.
@@ -462,7 +462,8 @@ impl UmodeSlotMapper<'_> {
     ///
     /// # Safety
     ///
-    /// Caller must guarantee that the page at address `paddr` is owned by a guest and has been shared with the hypervisor.
+    /// Caller must guarantee that the page at address `paddr` is owned by a guest and has been
+    /// shared with the hypervisor.
     pub unsafe fn map_addr(
         &self,
         vaddr: PageAddr<SupervisorVirt>,
--- a/src/hyp_map.rs
+++ b/src/hyp_map.rs
@@ -3,7 +3,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 use arrayvec::ArrayVec;
-use core::cell::RefCell;
+use core::cell::{RefCell, RefMut};
 use page_tracking::{HwMemMap, HwMemRegion, HwMemRegionType, HwReservedMemType, HypPageAlloc};
 use riscv_elf::{ElfMap, ElfSegment, ElfSegmentPerms};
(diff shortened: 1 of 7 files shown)


src/hyp_map.rs
```diff
@@ -410,6 +396,20 @@ impl HypMap {
         self.private_regions.iter()
     }
 
+    /// Returns the virtual address of U-mode mapping slot `slot`.
+    pub fn umode_slot_va(slot: UmodeSlotId) -> PageAddr<SupervisorVirt> {
+        let slot_num = match slot {
```
mnissler-rivos, line 417: Alternatively:

let slot_num = UmodeSlotId as usize;

(and if you want to be extra sure, give the enum cases explicit values)

src/hyp_map.rs
```diff
@@ -476,3 +504,103 @@ impl UmodeSlotMapper<'_> {
...
+        for (virt, phys) in vaddr
+            .iter_from()
+            .zip(pages.base().iter_from())
+            .take(num_pages as usize)
+        {
+            // Safety: These pages will be written by the hypervisor using the physical hypervisor
+            // mappings and read by U-mode through this mapping in read only. Safe because these are
+            // per-CPU mappings, and when the hypervior will be writing to these pages via the
+            // physical mappings no CPU will be able to access these pages through the U-mode
+            // mappings.
```
mnissler-rivos, line 558: nit: Also, you're not using memory outside of the ranges. This is obvious for pages, since we have allocated num_pages above. It's not immediately clear for vaddr though - probably guaranteed by memory layout though?

src/hyp_map.rs
```diff
@@ -476,3 +504,103 @@ impl UmodeSlotMapper<'_> {
...
+/// Umode Buffer errors.
+#[derive(Debug)]
+pub enum UmodeBufferError {
+    /// Address overflow on increment.
+    AddressOverflow,
+    /// Buffer is full.
+    NoSpace,
+}
+
+/// The U-mode buffer is a private area of the page table used by the hypervisor to pass data to the
```
mnissler-rivos, line 517: For my understanding: Can you explain what data items we're expecting to pass via this mechanism? Do we have additional static knowledge, e.g. that the data must be of one of a finite number of types?

(background: having seen only this patch, I'm unconvinced that we need a fully general buffer mechanism)

src/hyp_map.rs
glg-rv: Aware. I'd rather keep it like this to be explicit on their mapping. Giving explicit values in a enum as I've done for other types would just make it harder to understand where the address go.

In general in case SlotId grows as a type this should become a function associated to the type (unsure of the rust slang for this), but for now I like this little functions because on reading tells you exactly where the VA goes.

src/hyp_map.rs
glg-rv: So, lesson learned after _n_ reviews in salus is: This is safety, not correctness.

The safety to guarantee here is that we're not creating aliases in the hypervisor mappings. I.e., we're not breaking the rust memory model for salus by allowing to have the same memory mapped in multiple places -- I am doing that so I needed to be more verbose.

The question whether I am mapping the right page is a question of correctness of the patch. The question I am trying to answer with the safety comment: it is fine to create a safe function that uses this unsafe function?

src/hyp_map.rs
glg-rv: The data type will be call specific. I've been going back and forth between copying a repr(C) structure or having a byte buffer, and I went for the second for two reasons:

1. In the immediate need, the easiest way to get the data needed for attestation (PCRs and CDI_ID) is that of the byte buffer.
2. Nothing will prevent to expand the interface and just be able to append a particular type.

This is not a generic buffer, at least on this side. It's the simplest mechanism that will allow -- with fewer moving parts -- to implement a passing of data from salus to u-mode that I could think of.

glg-rv: (Edited and then deleted the wrong comment by mistake. Rewriting it)
This will be call specific. I have been going back and forth whether to have a repr(c) structure in a shared memory or going with byte buffer, I went for the second for two main reasons:

1. Given the immediate need of this, the best way to get data for attestation, PCRs and CDI ID, is through byte buffers.
2. In case things change, it will be easy to just expand this to be able to append (copy) a specific type.

This is not a generic buffer, at least not on the hypervisor side. This was the simplest way (fewer moving parts in Salus) to get data to U-mode.

Even the append() function, which might appear odds, serves the purpose of being able to collect all data and write them in order to the buffer.

src/hyp_map.rs
glg-rv: Added to the safety explanation the concept that we're mapping in the VA address reserved for it. Was giving it for granted.

src/hyp_map.rs
```diff
@@ -476,3 +504,103 @@ impl UmodeSlotMapper<'_> {
...
+/// current U-mode operation. This avoids mapping hypervisor data directly in U-mode. This area is
+/// written by the hypervisor and mapped read-only in U-mode.
+pub struct UmodeBuffer {
+    /// The start address of the U-mode buffer memory.
+    start: SupervisorPageAddr,
+    /// Pointer to the unused part of the buffer.
+    addr: SupervisorPhysAddr,
```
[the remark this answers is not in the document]
glg-rv: We wouldn't need to volatile part of it, nor the reading part of it. Feels a bit excessive for this use.

The dynamics here are a bit strange. This area is definitely fully owned by salus, and there will be really no concurrent reading. I like the use of DataInit.

This is essentially more of a rust interface question rather than architectural. I don't have particularly strong opinions about it.

src/umode.rs
```diff
@@ -262,7 +264,12 @@ impl UmodeTask {
         Ok(())
     }
 
-    pub fn reset(&mut self) -> Result<(), Error> {
+    /// Return the U-mode buffer for this task.
+    pub fn umode_buffer() -> RefMut<'static, UmodeBuffer> {
```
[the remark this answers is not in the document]
glg-rv: makes sense. Didn't like it but yes, i'll fix.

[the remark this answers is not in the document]
glg-rv: Done.

src/hyp_map.rs
mnissler-rivos: I'm probably making incorrect assumptions on how this is to be used and as a result I'm asking weird questions... Let me clarify where I'm coming from:

At a very high level, I was assuming that all you need is to encode an operation type and the data associated with it, dump that somewhere in memory. The umode can decode and process it. The (IMHO) natural way of doing this in rust is defining an enum type with each case holding relevant data. Then you can use a suitable serialization mechanism (internal interface with rust on either end, so something like bincode would probably work well) to encode the request. That's my mental model at least, probably overly simplistic.

What you have in the test code gave me the impression that the request handlers will have to do their own interpretation of what they find in the buffer (that's what I meant by "general buffer mechanism"), but I really don't know how you imagine actual operations to look like, so perhaps you can just sketch how an attestation-related operation would work?

src/hyp_map.rs
mnissler-rivos: I should have expressed myself more clearly: Memory safety also depends on avoiding buffer overflows, and so it's critical to make sure that we're mapping only the number of pages that we actually reserved (both in terms of address space and physical pages). The code above ensures this via num_pages, and there's UMODE_BUFFER_SIZE (rounded up to 4k alignment) free space in the memory map.

mnissler-rivos: Ah, thanks for elaborating.  I understand now that address space layout and region sizes are actually checked elsewhere and aren't depending on the parameters in this unsafe block to be accurate.

src/hyp_map.rs
mnissler-rivos: Fair enough, that approach makes sense. Please don't read these comments as requests to change something, I'm really just asking questions on stuff that seems unintuitive to me :)

src/hyp_map.rs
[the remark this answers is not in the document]
dgreid: I wasn't so much worried about the concurrent reading part, but liked that it wraps the unsafe access, I sent you a patch, let me know what you think.

glg-rv:
Bit of an update. Discussion about whether to use slices or  volatile slices continued offline. I think voltatile slice is more future-ready and, as @dgreid noted, has a quite a nice interface to handle data.

I implemented this as a buffer because I was worried aobut the amount of data and how to retrieve layer data from current attestation code. Upon testing with certificates in u-mode, I am now less worried about keeping this interface generic.

Upcoming a new series of patches (mostly a rewrite) that will implement something using the same mechanism, but in a simpler way.

PR #226, opened 2023-02-06
U-mode: share data between hypervisor and u-mode.

glg-rv:
This PR is the logical continuation (and rework) of #222. What has changed is that I have worked until the certificate generation was fully working in U-mode, and made things as simple as possible in order to create a simple, but sane and future proof interface.

First five patches are a clean up:

Patch 1 & 2 remove the use of unwrap on PageAddr creation similar to what @dgreid has done for other parts of the code.
Patch 3 & 4 are cleanup over the recent U-mode patches (Goodbye cowsay).
Patch 5 is a cleanup of the API. Recently the `input` and `output` functions were removed from the API, as they were considered unsafe for salus, and this meant that the registers as passed in U-mode API calls were strictly call-dependent. The logical implication for that was that the API could be greatly simplified and made more clear. This patch implements that.

The rest of the patches introduce the Umode Shared Memory region. Equivalent to #222 U-mode buffer in mechanism, is an area of hypervisor memory that is mapped read-only on U-mode.

### DIFF
--- a/src/hyp_map.rs
+++ b/src/hyp_map.rs
@@ -4,6 +4,7 @@
 
 use arrayvec::ArrayVec;
 use core::cell::RefCell;
+use data_model::{DataInit, VolatileMemory, VolatileMemoryError, VolatileSlice};
 use page_tracking::{HwMemMap, HwMemRegion, HwMemRegionType, HwReservedMemType, HypPageAlloc};
 use riscv_elf::{ElfMap, ElfSegment, ElfSegmentPerms};
 use riscv_page_tables::{
@@ -34,7 +35,7 @@ pub enum Error {
     ElfUnalignedSegment,
     /// U-mode ELF segment is not in U-mode VA area.
     ElfInvalidAddress,
-    /// Not enough space on the U-mode map area.
+    /// Not enough space in the U-mode map area.
     OutOfMap,
     /// Could not create a mapper for the U-mode area.
     MapperCreationFailed,
@@ -42,6 +43,8 @@ pub enum Error {
     MapFailed,
     /// Could not unmap the U-mode area.
     UnmapFailed,
+    /// U-mode Shared Memory Error.
+    UmodeShared(VolatileMemoryError),
 }
 
 // Represents a virtual address region of the hypervisor that will be the same in all pagetables.
@@ -124,11 +127,11 @@ impl SharedRegion {
 }
 
 // U-mode binary mappings start here.
-const UMODE_VA_START: u64 = 0xffffffff00000000;
+const UMODE_BINARY_START: u64 = 0xffffffff00000000;
 // Size in bytes of the U-mode binary VA area.
-const UMODE_VA_SIZE: u64 = 128 * 1024 * 1024;
+const UMODE_BINARY_SIZE: u64 = 128 * 1024 * 1024;
 // U-mode binary mappings end here.
-const UMODE_VA_END: u64 = UMODE_VA_START + UMODE_VA_SIZE;
+const UMODE_BINARY_END: u64 = UMODE_BINARY_START + UMODE_BINARY_SIZE;
 
 // The addresses between `UMODE_MAPPINGS_START` and `UMODE_MAPPINGS_START` + `UMODE_MAPPINGS_SIZE`
 // is an area of the private page table where the hypervisor can map pages shared from guest
@@ -138,12 +141,21 @@ const UMODE_MAPPING_SLOT_SIZE: u64 = 4 * 1024 * 1024;
 const_assert!(PageSize::Size4k.is_aligned(UMODE_MAPPING_SLOT_SIZE));
 
 // Start of the private U-mode mappings area.  Must be 4k-aligned.
-const UMODE_MAPPINGS_START: u64 = UMODE_VA_END + 4 * 1024 * 1024;
+const UMODE_MAPPINGS_START: u64 = UMODE_BINARY_END + 4 * 1024 * 1024;
 const_assert!(PageSize::Size4k.is_aligned(UMODE_MAPPINGS_START));
 //The number of slots available for mapping.
 const UMODE_MAPPING_SLOTS: u64 = 2;
 // Maximum size of the private mappings area. Must be 4k-aligned.
 const UMODE_MAPPINGS_SIZE: u64 = UMODE_MAPPING_SLOTS * UMODE_MAPPING_SLOT_SIZE;
+const UMODE_MAPPINGS_END: u64 = UMODE_MAPPINGS_START + UMODE_MAPPINGS_SIZE;
+const_assert!(PageSize::Size4k.is_aligned(UMODE_MAPPINGS_END));
+
+/// Start of the U-mode Shared area.
+pub const UMODE_SHARED_START: u64 = UMODE_MAPPINGS_END + 4 * 1024 * 1024;
+const_assert!(PageSize::Size4k.is_aligned(UMODE_SHARED_START));
+/// Size of the U-mode Shared area.
+pub const UMODE_SHARED_SIZE: u64 = 4 * 1024;
+const_assert!(PageSize::Size4k.is_aligned(UMODE_SHARED_SIZE));
 
 /// Generic Id names for each of the U-mode mapping slots.
 /// There is no mandated use for each of the slots, and caller can decide to map each of them
@@ -156,7 +168,7 @@ pub enum UmodeSlotId {
 
 // Returns true if `addr` is contained in the U-mode VA area.
 fn is_umode_addr(addr: u64) -> bool {
-    (UMODE_VA_START..UMODE_VA_END).contains(&addr)
+    (UMODE_BINARY_START..UMODE_BINARY_END).contains(&addr)
 }
 
 // Returns true if (`addr`, `addr` + `len`) is a valid non-empty range in the VA area.
@@ -286,6 +298,8 @@ impl PrivateRegion {
 pub struct HypPageTable {
     /// The pagetable containing hypervisor mappings.
     sv48: FirstStagePageTable<Sv48>,
+    /// U-mode shared region for this page-table.
+    umode_shared: RefCell<UmodeSharedRegion>,
     /// A pte page pool for U-mode mappings.
     pte_pages: RefCell<SeqPageIter<InternalClean>>,
 }
@@ -298,7 +312,7 @@ impl HypPageTable {
         satp.get()
     }
 
-    /// Restores U-mode mappings to initial state.
+    /// Restores U-mode ELF mappings to initial state.
     pub fn restore_umode(&self) {
         for r in HypMap::get()
             .private_regions()
@@ -352,6 +366,17 @@ impl HypPageTable {
         self.sv48
             .unmap_range(vaddr, PageSize::Size4k, num_pages)
             .map_err(|_| Error::UnmapFailed)
+    }
+
+    /// Share a structure with U-mode by copying it into the `UmodeSharedRegion`. The structure will be
+    /// mapped in U-mode at UMODE_SHARED_START and will be accessible by U-mode read-only.
+    pub fn share_with_umode<T: DataInit>(&self, data: T) -> Result<(), Error> {
+        self.umode_shared.borrow_mut().store(data)
+    }
+
+    /// Clear all data previously shared with U-mode with `share_with_umode`.
+    pub fn clear_umode_shared(&self) {
+        self.umode_shared.borrow_mut().clear()
     }
 }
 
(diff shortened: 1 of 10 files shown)


u-mode-api/src/lib.rs
```diff
@@ -144,73 +144,61 @@ impl TryFrom<u64> for UmodeOp {
...
+                args: [out_addr, in_addr, len, 0, 0, 0, 0],
             })
         }
     }
 }
 
 impl TryIntoRegisters for UmodeRequest {
     fn try_from_registers(regs: &[u64]) -> Result<UmodeRequest, Error> {
+        let mut args = [0; 7];
+        args.as_mut_slice().copy_from_slice(&regs[1..8]);
```
dgreid, line 211: nit: `1..8` could use a comment.

dgreid (approved):
Thanks, LGTM

u-mode-api/src/lib.rs
glg-rv: Fair. Updated.

u-mode-api/src/lib.rs
```diff
@@ -144,73 +144,65 @@ impl TryFrom<u64> for UmodeOp {
...
-    /// Optional start of mapped area accessible as read-only.
-    pub in_addr: Option<u64>,
-    /// If in_addr is valid, length of the area acessible as read-only.
-    pub in_len: usize,
-    /// Optional start of mapped area accessible as read-write.
-    pub out_addr: Option<u64>,
-    /// If in_addr is valid, length of the area acessible as read-write.
-    pub out_len: usize,
+    /// Arguments of the operation.
+    pub args: [u64; 7],
```
mnissler-rivos, line 151: Side note suggestion for consideration, not a request for code changes: With my rust data modeling nerd hat on, I would rather make these data fields in the UmodeOp enum constants like so:

```
pub enum UmodeOp {
    // ...
    MemCopy(u64 dest_addr, u64 src_addr, u64 len),
}
```

And then you can have the appropriate TryIntoRegisters implementations to encode/decode them.

src/umode.rs
```diff
@@ -287,35 +297,59 @@ impl UmodeTask {
...
                     // 2. setup umode again.
                     // Panic if we can't restore umode: the hypervisor is in an unrecoverable state.
                     task.reset().expect("Failed to recover U-mode.");
                 }
-                Error::Umode(_) => {}
+                ExecError::Umode(_) => {}
             }
         }
-        ret
+        if shared_data.is_some() {
```
mnissler-rivos, line 326: Another rust language nerd comment just for consideration, no need to implement: The share_with_umode / clear_umode_shared bracket here would seem to be a good fit for a RAII object - if share_with_umode() were to return an object that clears the umode buffer when it gets dropped, you would have more confidence that the clearing can't accidentally get bypassed.

mnissler-rivos (commented):
This looks pretty nice now! I put some comments on opportunities to sacrifice to the rust language gods, but these are all optional.

u-mode-api/src/lib.rs
glg-rv: I thought about this. What you suggest is indeed the rust-est way to do it. I just felt that this was more concise and explained easier to people reading it how the API maps to it. No strong opinions on it.

src/umode.rs
glg-rv: Yes, thoguht about this too. But then I thought that I could get away with defining yet another type that had to be looked up when reading the code by bolting the behavior in send_req, which is I think the right interface.

u-mode-api/src/lib.rs
glg-rv: Right, so I implemented this -- I already did something similar for hypcalls right below this -- and yes, it is much better.

Updated the PR. Thank you.

u-mode-api/src/lib.rs
```diff
@@ -117,100 +117,96 @@ impl IntoRegisters for Result<(), Error> {
...
-            in_addr: if regs[1] == 0 { None } else { Some(regs[1]) },
-            in_len: regs[2] as usize,
-            out_addr: if regs[3] == 0 { None } else { Some(regs[3]) },
-            out_len: regs[4] as usize,
-        };
-        Ok(req)
+        match regs[0] {
+            UMOP_NOP => Ok(UmodeRequest::Nop),
+            UMOP_MEMCOPY => Ok(UmodeRequest::MemCopy {
+                //   A1: `out_addr`
```
mnissler-rivos, line 200: nit: these comments are pretty redundant with the code and don't really add anything - get rid of them?

glg-rv:
Updated with @mnissler-rivos suggestion of switching to enum with associated data, much like I was already doing with hypcalls.

Please note, the functions defined as `impl UmodeRequest` are now looking silly, and will be removed in the upcoming commits.

u-mode-api/src/lib.rs
mnissler-rivos: Nice, I like it!

src/hyp_map.rs
```diff
@@ -132,13 +133,16 @@ const UMODE_VA_END: u64 = UMODE_VA_START + UMODE_VA_SIZE;
 // The addresses between `UMODE_MAPPINGS_START` and `UMODE_MAPPINGS_START` + `UMODE_MAPPINGS_SIZE`
 // is an area of the private page table where the hypervisor can map pages shared from guest
 // VMs. The area is divided in slots, of equal size `UMODE_MAPPING_SLOT_SIZE`.
+// Must be multiple of 4k.
 const UMODE_MAPPING_SLOT_SIZE: u64 = 4 * 1024 * 1024;
+const_assert!(PageSize::Size4k.is_aligned(UMODE_MAPPING_SLOT_SIZE));
 
-// Start of the private U-mode mappings area.
+// Start of the private U-mode mappings area.  Must be 4k-aligned.
 const UMODE_MAPPINGS_START: u64 = UMODE_VA_END + 4 * 1024 * 1024;
```
abrestic-rivos, line 144: pre-existing, but since all these constants are technically U-mode virtual addresses, should `UMODE_VA_*` maybe be renamed to `UMODE_BINARY_*`?

src/hyp_map.rs
```diff
@@ -144,6 +147,15 @@ const_assert!(PageSize::Size4k.is_aligned(UMODE_MAPPINGS_START));
 const UMODE_MAPPING_SLOTS: u64 = 2;
 // Maximum size of the private mappings area. Must be 4k-aligned.
 const UMODE_MAPPINGS_SIZE: u64 = UMODE_MAPPING_SLOTS * UMODE_MAPPING_SLOT_SIZE;
+const UMODE_MAPPINGS_END: u64 = UMODE_MAPPINGS_START + UMODE_MAPPINGS_SIZE;
+const_assert!(PageSize::Size4k.is_aligned(UMODE_MAPPINGS_END));
+
+/// Start of the U-mode Shared area.
+pub const UMODE_SHARED_START: u64 = UMODE_MAPPINGS_END + 4 * 1024 * 1024;
```
abrestic-rivos, line 154: bikeshed: Might `UMODE_INPUT` or `UMODE_INPUT_START` be a more precise name for this buffer? Sure it's "shared", but it's used solely for passing an input to the U-mode program.

u-mode/src/main.rs
```diff
@@ -26,6 +26,29 @@ struct UmodeTask {
...
+    fn op_print_string(&self, len: usize) -> Result<(), UmodeApiError> {
+        // Print maximum 80 chars.
+        const MAX_LENGTH: usize = 80;
+        let vs_input = self
+            .vslice
+            .get_slice(0, len)
+            .map_err(|_| UmodeApiError::InvalidArgument)?;
+        // Copy input from volatile slice.
+        let mut input = [0u8; MAX_LENGTH];
+        vs_input.copy_to(&mut input[..]);
```
abrestic-rivos, line 42: Because we're using `VolatileSlice` for the shared buffer, that means we're effectively forced to copy the bytes out in order to operate on the data (unless we go `unsafe`), right? I get wanting to copy the data when passing data from the hypervisor to U-mode, but once we're in U-mode this extra `memcpy()` seems pretty wasteful :/

abrestic-rivos (approved):
Some minor comments, but overall LGTM. I have some catching up to do on U-mode stuff :)

src/hyp_map.rs
glg-rv: Yes, I had that patch somewhere in my tree. It is on my list to change.

src/hyp_map.rs
glg-rv: Yes. I am still keeping it open to the possibility of making this buffer read/writable. Name reflects that.

u-mode/src/main.rs
glg-rv: Yes, it is. It's  404 bytes though. So I thought this wouldn't be that much worth of an unsafe, given also the amount of memory the attestation code will move around in U-mode.

src/hyp_map.rs
```diff
@@ -310,11 +314,12 @@ impl HypPageTable {
...
-        PageAddr::new(RawAddr::supervisor_virt(
-            UMODE_MAPPINGS_START + slot_num * UMODE_MAPPING_SLOT_SIZE,
-        ))
-        .unwrap()
+        // UMODE_MAPPING_START and UMODE_MAPPING_SLOT_SIZE are required by static assertion to be
+        // page aligned, `round_down` will be a nop.
+        PageAddr::with_round_down(
+            RawAddr::supervisor_virt(UMODE_MAPPINGS_START + slot_num * UMODE_MAPPING_SLOT_SIZE),
+            PageSize::Size4k,
+        )
```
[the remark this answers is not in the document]
glg-rv: Yes, This was my attempt at reproducing changes as started here: https://github.com/rivosinc/salus/pull/212/files

And wanted to go with something that follows along the same lines so they can all be changed later, when we find a better interface.

src/hyp_map.rs
glg-rv: Done. It was a change I had in the upcoming (next after next) signing PR.

src/hyp_map.rs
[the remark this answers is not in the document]
dgreid: I hate unwraps because they almost always signal we aren't using the defined types correctly. And in this case a const PageAddr is the 'correct' fix. But that's a pretty tough exercise as you can't declare the `AS` type parameter as const. I'm OK landing this either way and we can fight the const constructor battle separately.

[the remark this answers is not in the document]
glguida: Yes please. Can we please merge this?

dgreid (approved):

src/hyp_map.rs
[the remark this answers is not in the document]
dgreid: Yes, I'm good with merging this one. Getting const page addrs was not fun, but I got something to work: https://github.com/dgreid/salus/commits/topic/const_page_addr

I can rebase after this lands.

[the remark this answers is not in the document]
mnissler-rivos: And it turns out Dylan is way ahead of me again :) Let me comment over there.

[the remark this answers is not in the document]
glg-rv: Came up with this macro. It does use unwrap (alternative is to use an unsafe function to create a raw page addr), but at least it's contained in a const_assert.

We could use this in the future. 

```
index d4431b4..f790591 100644
--- a/riscv-pages/src/page.rs
+++ b/riscv-pages/src/page.rs
@@ -169,6 +169,22 @@ pub struct PageAddr<AS: AddressSpace> {
     addr: RawAddr<AS>,
 }
 
+/// Create a PageAddr from an aligned constant.
+///
+/// Usage example:
+///
+///     use riscv_pages::*;
+///
+///     let x = const_page_addr!(0x1000, PageSize::Size4k, SupervisorVirt::default());
+#[macro_export]
+macro_rules! const_page_addr {
+    ($addr: expr, $page_size: expr, $address_space: expr) => ({
+        static_assertions::const_assert!($page_size.is_aligned($addr));
+        // Unwrap okay: static assertion right above ensures aslignment.
+        PageAddr::with_alignment(RawAddr::new($addr, $address_space), PageSize::Size4k).unwrap();
+    })
+}
+
 /// A page-aligned address in the `SupervisorPhys` address space.
 pub type SupervisorPageAddr = PageAddr<SupervisorPhys>;
 /// A page-aligned address in a `GuestPhys` address space.
@@ -781,6 +797,13 @@ mod tests {
         .is_some());
     }
 
+    #[test]
+    fn const_test() {
+        let _x = const_page_addr!(0x1000, PageSize::Size4k, SupervisorVirt::default());
+        let _x = const_page_addr!(0x200000, PageSize::Size2M, SupervisorVirt::default());
+        let _x = const_page_addr!(0x40000000, PageSize::Size1G, SupervisorVirt::default());
+    }
+
     #[test]
     fn round_phys() {
         assert!(

```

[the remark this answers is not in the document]
glg-rv: I am actually okay with adding this after it lands, or just importing your patch in the PR. I don't have strong opinion. I tried to do something useful adding a change that removed the previous unwraps with something similar that what happened before (plus adding const_assert).

The unfortunate timing is that the patch that introduced the unwrap above went in at the same time as your patch removing other unwraps, so I tried to fix this up.

u-mode/src/main.rs
abrestic-rivos: Yeah that's fine. Was just wondering what sizes to expect here. If we end up getting much larger I'd like to us remove that copy.

PR #227, opened 2023-02-08
README: Improve accuracy of the instructions

sboeuf:
While going through the `README` steps, I came across a few road blocks that I thought would be best documented directly in there.

- Fix rustup hyperlink

- Indicate libslirp must be installed
    As a prerequisite before building QEMU, libslirp must be present
    otherwise SLIRP won't be available through the QEMU binary. It
    prevents an error from being triggered when using "-netdev user".

- Indicate how to build the cross compiler
    Provide a link to the instructions on how to build and install the
    GCC RISCV64 compiler for a Linux machine.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,11 +20,13 @@ The `Makefile` provides targets for running various hosts in QEMU.
 ### Prerequisites
 
 Toolchains:
-- Rust: Install [rustup](rustup.rs), then: `rustup target add riscv64gc-unknown-none-elf`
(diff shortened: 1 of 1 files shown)


glg-rv:
Possibly squash them in a single commit?

sboeuf:
> Possibly squash them in a single commit?

Sure!

abrestic-rivos (approved):
thanks!

sameo (approved):

PR #230, opened 2023-02-08
Salus: Move certificate generation to U-mode.

glg-rv:
This PR moves the certificate generation (through the get_evidence call) in U-mode.

Some preparation:

Patch 1: as the U-mode certificate call will use both Umode guest mappings and Umode shared region, the test calls for these functions (PrintString and MemCopy) can be removed. I have left some stub implementation of the vendor extensions of salus for future use, but can be removed completely, at the cost of reimplementing it in case we have another temporary need in the future.

Patch2 allow Umode request to return a `Result<u64>` instead of a `Result<()>`, to allow returning directly values that can be mapped in `SbiReturn`.

Patch3 exposes the number of management registers to the API.

The following patches implement the call in Umode-API, U-mode and the hypervisor itself.

Last patch cleans the certificate generation code in hypervisor.

This patch is possibly still a RFC at this stage.

### DIFF
--- a/src/vm.rs
+++ b/src/vm.rs
@@ -16,7 +16,7 @@ use s_mode_utils::print::*;
 use sbi_rs::{salus::*, Error as SbiError, *};
 
 use crate::guest_tracking::{GuestStateGuard, GuestVm, Guests};
-use crate::hyp_map::UmodeSlotId;
+use crate::hyp_map::{UmodeSlotId, UmodeSlotPerm};
 use crate::umode::UmodeTask;
 use crate::vm_cpu::{ActiveVmCpu, VmCpu, VmCpuParent, VmCpuStatus, VmCpuTrap, VmCpus, VM_CPUS_MAX};
 use crate::vm_pages::Error as VmPagesError;
@@ -1512,7 +1512,7 @@ impl<'a, T: GuestStagePagingMode> FinalizedVm<'a, T> {
         slot: UmodeSlotId,
         addr: u64,
         len: u64,
-        writable: bool,
+        slot_perm: UmodeSlotPerm,
     ) -> EcallResult<(u64, GuestUmodeMapping)> {
         let base = PageSize::Size4k.round_down(addr);
         let end = addr
@@ -1524,7 +1524,7 @@ impl<'a, T: GuestStagePagingMode> FinalizedVm<'a, T> {
                 slot,
                 self.guest_addr_from_raw(base)?,
                 PageSize::num_4k_pages(end - base),
-                writable,
+                slot_perm,
             )
             .map_err(EcallError::from)?;
         let vaddr = umode_mapping.vaddr().bits() + (addr - base);
--- a/src/vm.rs
+++ b/src/vm.rs
@@ -4,20 +4,19 @@
 
 use attestation::{AttestationManager, Error as AttestationError, TcgPcrIndex};
 use core::{mem, ops::ControlFlow, slice};
-use der::Decode;
 use drivers::{imsic::*, pmu::PmuInfo};
 use page_tracking::collections::PageBox;
 use page_tracking::{LockedPageList, PageList, PageTracker, TlbVersion};
-use rice::x509::{request::CertReq, MAX_CSR_LEN};
 use riscv_page_tables::{GuestStagePageTable, GuestStagePagingMode};
 use riscv_pages::*;
 use riscv_regs::{DecodedInstruction, Exception, GprIndex, Instruction, Interrupt, Trap};
 use s_mode_utils::print::*;
 use sbi_rs::{salus::*, Error as SbiError, *};
+use u_mode_api::Error as UmodeApiError;
 
 use crate::guest_tracking::{GuestStateGuard, GuestVm, Guests};
 use crate::hyp_map::{UmodeSlotId, UmodeSlotPerm};
-use crate::umode::UmodeTask;
+use crate::umode::{Error as UmodeError, ExecError, UmodeTask};
 use crate::vm_cpu::{ActiveVmCpu, VmCpu, VmCpuParent, VmCpuStatus, VmCpuTrap, VmCpus, VM_CPUS_MAX};
 use crate::vm_pages::Error as VmPagesError;
 use crate::vm_pages::{
@@ -186,6 +185,24 @@ impl From<AttestationError> for EcallError {
             // TODO: Map individual error types.
             // InvalidParam may not be the right value for each error.
             _ => EcallError::Sbi(SbiError::InvalidParam),
+        }
+    }
+}
+
+impl From<UmodeApiError> for EcallError {
+    fn from(error: UmodeApiError) -> EcallError {
+        match error {
+            UmodeApiError::InvalidArgument => EcallError::Sbi(SbiError::InvalidParam),
+            _ => EcallError::Sbi(SbiError::Failed),
+        }
+    }
+}
+
+impl From<UmodeError> for EcallError {
+    fn from(error: UmodeError) -> EcallError {
+        match error {
+            UmodeError::Exec(ExecError::Umode(api_error)) => api_error.into(),
+            _ => EcallError::Sbi(SbiError::Failed),
         }
     }
 }
@@ -1559,7 +1576,6 @@ impl<'a, T: GuestStagePagingMode> FinalizedVm<'a, T> {
                     evidence_format,
                     cert_addr_out,
                     cert_size as usize,
-                    active_pages,
                 )
                 .into(),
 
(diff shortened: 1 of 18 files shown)


u-mode-api/src/cert.rs
```diff
@@ -0,0 +1,28 @@
+use attestation::MSMT_REGISTERS;
```
abrestic-rivos, line 1: this needs a copyright / license thingy at the top of the file

u-mode-api/src/cert.rs
```diff
@@ -0,0 +1,28 @@
...
+use data_model::DataInit;
+
+/// CDI ID length.
+pub const CDI_LEN: usize = 20;
```
abrestic-rivos, line 5: `attestation/src/manager.rs` says this is 32

u-mode/Cargo.toml
```diff
@@ -7,5 +7,12 @@ edition = "2021"
...
 data_model = { path = "../data-model" }
+der = { version = "0.6.0", features = ["derive", "flagset", "oid"] }
 libuser = { path = "../libuser" }
 u_mode_api = { path = "../u-mode-api" }
+rice = { git = "https://github.com/rivosinc/rice" }
+const-oid = { version = "0.9.0", features = ["db"] }
+sha2 = {version = "0.10", default-features = false }
+generic-array = "0.14.5"
+ed25519 = { version = "1.5.2", default-features = false, features = ["pkcs8"] }
+ed25519-dalek = { version = "1.0.1", default-features = false, features = ["u64_backend"] }
```
abrestic-rivos, line 18: nit: need newline at EOF

src/vm.rs
```diff
@@ -1621,55 +1637,47 @@ impl<'a, T: GuestStagePagingMode> FinalizedVm<'a, T> {
...
+        certout_len: usize,
     ) -> EcallResult<u64> {
-        if cert_request_size > MAX_CSR_LEN {
-            return Err(EcallError::Sbi(SbiError::InsufficientBufferCapacity));
+        // Map CSR read-only.
+        let (csr_addr, _csr_mapping) = self.map_guest_range_in_umode_slot(
+            UmodeSlotId::A,
+            csr_guest_addr,
+            csr_len as u64,
+            false,
```
abrestic-rivos, line 1654: pre-existing, and doesn't have to be part of this PR: I assume the `bool` here means RO vs RW? Can we make that an enum instead? :)

abrestic-rivos (commented):
The U-mode mechanics look reasonable to me. I'll leave it to sameo@ to review the cert specifics.

u-mode-api/src/cert.rs
glg-rv: No, that's the CDI length. The CDI ID (generated by the public key) is 20. Defined in rice/cdi.rs CDI_ID_LEN.

src/vm.rs
glg-rv: Ah that's a great suggestion. Will change.

u-mode/Cargo.toml
glg-rv: Ah yes, fixed one but not this.

src/vm.rs
glg-rv: Done.

u-mode/Cargo.toml
glg-rv: Done.

u-mode-api/src/cert.rs
glg-rv: Changed to the less confusing name CDI_ID_LEN.

u-mode-api/src/cert.rs
glg-rv: Done.

u-mode/src/cert.rs
```diff
@@ -0,0 +1,103 @@
...
+    CertificateCreationFailed(rice::Error),
+    /// Output Certificate buffer too small.
+    CertificateBufferTooSmall(usize, usize),
+}
+
+struct UmodeSigner {}
+
+impl Signer<Signature> for UmodeSigner {
+    fn try_sign(&self, _: &[u8]) -> Result<Signature, ed25519::Error> {
+        // TODO: Implement Signing of certificate.
```
sameo, line 40: So is the plan to sign from S-mode or from U-mode?
If we want to sign from U-mode, then `UmodeSigner` should carry a `KeyPair` that's basically derived from the attestation layer `CDI` (The layer's secret).

If we want to sign from S-mode, then I think we want to expand the rice `Layer` API to sign a blob with its `CDI`.

I was under the impression that we wanted to do the whole thing (including signing) from U-mode, in which case I suggest the following:

* Continue generating the CDI from the attestation manager
* Pass the CDI from the attestation manager to the U-mode task
* Have the U-mode task generate a `rice` layer for the CDI
* Generate the certificate from the U-mode task, based on the passed CSR, including signing it.

Would  generating a rice layer be complex from the U-mode task? Apologize if I previously misunderstood your intentions with https://github.com/rivosinc/rice/pull/4.

glg-rv: Current interface is generic enough that can be done both ways (passing the keys from CDI to U-mode is something I really don't want to do, but can be done), and the signing patch is upcoming. But plan is to sign in S-mode.

Plan is (half-implemented).

1. UmodeSigner will send an ecall, passing certificate data.
2. send_req will have a closure that will be able to handle specific ecalls.
3. the closure will call the attestation manager to sign (which in the end will have to expose at Layer either the key or the signing). As you said, and as I planned.

I don't think you misunderstood the reason for that patch. That patch was for certificate creation. I wanted to separate the concept of creating certificate and signing it in two separate patches exactly to find the best solution for each of these separate problems.

u-mode/src/cert.rs
```diff
@@ -0,0 +1,103 @@
...
+        Signature::from_bytes(&[0; 64])
+    }
+}
+
+pub fn get_certificate_sha384(
+    csr_input: &[u8],
+    data: GetEvidenceShared,
+    certout: &mut [u8],
```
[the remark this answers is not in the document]
glg-rv: Ack for `evidence`. 

Disagree on `cert_bytes`, which is used later in the function. the subtlety here is that `csr_input` and `certout` are guest mapped. So they need a copy. I don't want to have this copy in `src/main.rs` because I want to keep all rice-specific knowledge in this file.

So possibly `csr_guest` and `cert[out]_guest` could be better names.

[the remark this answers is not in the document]
glg-rv: Renamed to `evidence`, changed `certout` to `cert_output` to be more simmetrical with input, and fixed the comments.

u-mode/src/cert.rs
glg-rv: > > Current interface is generic enough that can be done both ways (passing the keys from CDI to U-mode is something I really don't want to do, but can be done), and the signing patch is upcoming. But plan is to sign in S-mode.
> 
> Just to confirm: You're reluctant to pass the CDI to U-mode for security reasons?

Its a good practice not to pass private keys around. Especially once it was decided that an instance of U-mode will be able to process many requests coming from different VMs, and we don't reset the stack and data at each call (which is not small given this code requirements).

> 
> > Plan is (half-implemented).
> > ```
> > 1. UmodeSigner will send an ecall, passing certificate data.
> > 
> > 2. send_req will have a closure that will be able to handle specific ecalls.
> > 
> > 3. the closure will call the attestation manager to sign (which in the end will have to expose at Layer either the key or the signing). 
> > ```
> 
> Would it be simpler to assume that when the certificate U-mode task returns (ecall, right), the `cert_out` must be signed?

Let me rephrase.

The ecall for signing will not return the U-mode task, it will be a hypervisor call. The hypervisor call that signal the end of the operation is NextOp. When this is called, then the operation returns.

What I want to do, for the reason listed above, is to use a specific hypcall (implemented only in one particular send_req place in hypervisor) to sign the key.

When the U-mode operation will return, the certificate will be signed.

But again, you seem to believe that this won't even allow something I am  opposed to: signing in U-mode. Rather than recreating a layer each time we create a certificate, suppose someone wants to do U-mode signing: it gets the key from the shared data structure from the hypervisor (again, via CDI or private key directly), and implements UmodeSigner so that it does that.

> 
> > I don't think you misunderstood the reason for that patch. That patch was for certificate creation. I wanted to separate the concept of creating certificate and signing it in two separate patches exactly to find the best solution for each of these separate problems.
> 
> I agree that they're different problems but here they're tightly linked. The certificate obviously must be signed, and it's always going to be signed by the same entity, the TSM layer. So if we reduce our scope to that, maybe we can slightly simply things.

I might be misunderstanding these words, but what you mean is this:  pass the CDI (secret), recrate the layer in U-mode at each certificate call, and leave the current rice unmodified. I understand how this could sound appealing, but I rather go towards a model where  trust in U-mode is reduced -- otherwise, why not keep it in the hypervisor in the first place?.

The reason we build the certificate `from_raw_parts` is that U-mode will trust the hypervisor to set up things so that they come from the same layer (CDI_ID and the sign hypercall). I think this is more than fair, and requires no private key sharing.

mnissler-rivos: (sorry for pressing the resolve button accidentally)

Side note re key exposure: Whatever we do, I think we should leave the door open for a world were we have DICE key derivation handled in a hardware block that binds the keys more tightly to the hardware. In that world, signing would be a call to the hardware block, providing a key handle and the data (or hash of it) to sign.

And there's one point in this discussion that might make sense to get more clarity on: What security objectives (if any) we want to achieve by separating out this code into u-mode, at least I'm personally still pretty fuzzy on that question.

glg-rv: The latter is a good question, and I think the reason why you're fuzzy is that it has been a secondary target. One of the objectives of U-mode (and I also joined relatively late in the discussion) was to help make salus minimal. The attestation code, whose interface has to have complex things like DER encoded certificates and CSR, definitely brought a lot of needed complexity and third party requirement. It became a very easy target as the first code to be ported to U-mode.

U-mode doesn't maintain state and its main objective is to try to remove third party crates from U-mode. Which is why this patch moves all the x509 handling in U-mode.

It becomes, in my opinion, extremely natural also making so that the keys are kept only in the hypervisor. It is not a strict requirement but it's always better to do so (for reasons that I think should be self-evident and outlined in broad strokes above). In this case is also relatively easy, so I really see no reason for not doing this.

But of course, I am open to hear reasons for different opinions, and @sameo is the person that knows best how the attestation will evolve in the future. Hence this discussion.

sameo: > > > Current interface is generic enough that can be done both ways (passing the keys from CDI to U-mode is something I really don't want to do, but can be done), and the signing patch is upcoming. But plan is to sign in S-mode.
> > 
> > 
> > Just to confirm: You're reluctant to pass the CDI to U-mode for security reasons?
> 
> Its a good practice not to pass private keys around. 

Of course. I was just wondering if we were trying to overcome a technical issue, but that's not the case.
> Especially once it was decided that an instance of U-mode will be able to process many requests coming from different VMs, and we don't reset the stack and data at each call (which is not small given this code requirements).
> 

Yes, that would be a no-go. Thanks for the clarification.

> > > Plan is (half-implemented).
> > > ```
> > > 1. UmodeSigner will send an ecall, passing certificate data.
> > > 
> > > 2. send_req will have a closure that will be able to handle specific ecalls.
> > > 
> > > 3. the closure will call the attestation manager to sign (which in the end will have to expose at Layer either the key or the signing). 
> > > ```
> > 
> > 
> > Would it be simpler to assume that when the certificate U-mode task returns (ecall, right), the `cert_out` must be signed?
> 
> Let me rephrase.
> 
> The ecall for signing will not return the U-mode task, it will be a hypervisor call. The hypervisor call that signal the end of the operation is NextOp. When this is called, then the operation returns.
> 
> What I want to do, for the reason listed above, is to use a specific hypcall (implemented only in one particular send_req place in hypervisor) to sign the key.
> 
> When the U-mode operation will return, the certificate will be signed.
> 

Sounds good to me.

> But again, you seem to believe that this won't even allow something I am opposed to: signing in U-mode. Rather than recreating a layer each time we create a certificate, suppose someone wants to do U-mode signing: it gets the key from the shared data structure from the hypervisor (again, via CDI or private key directly), and implements UmodeSigner so that it does that.
> 

I suppose you mean it implements the `Signer` trait, which is what `UmodeSigner` will do via an `ECALL`.
FWIW, creating a rice `Layer` basically constructs a `Signer` implementer which happens to be the `KeyPair` derived from the curent CDI. Unless you generate the next layer (which you could not from a U-mode task, lacking the next layer TCI), it does not do more than that.

But I agree with your point of not signing from U-mode, so implementing `Signer` as a call back to Salus makes sense to me. Is that something that you're planning to add as a separate PR?

> > > I don't think you misunderstood the reason for that patch. That patch was for certificate creation. I wanted to separate the concept of creating certificate and signing it in two separate patches exactly to find the best solution for each of these separate problems.
> > 
> > 
> > I agree that they're different problems but here they're tightly linked. The certificate obviously must be signed, and it's always going to be signed by the same entity, the TSM layer. So if we reduce our scope to that, maybe we can slightly simply things.
> 
> I might be misunderstanding these words, but what you mean is this: pass the CDI (secret), recrate the layer in U-mode at each certificate call, and leave the current rice unmodified. I understand how this could sound appealing, but I rather go towards a model where trust in U-mode is reduced -- otherwise, why not keep it in the hypervisor in the first place?.
> 
> The reason we build the certificate `from_raw_parts` is that U-mode will trust the hypervisor to set up things so that they come from the same layer (CDI_ID and the sign hypercall). I think this is more than fair, and requires no private key sharing.

CDI sharing is not a recommended practice, but if we move towards a model where the U-mode stack would not even be zero'ed then it's not even something we should discuss. Sorry for the noise.

sameo: > (sorry for pressing the resolve button accidentally)
> 
> Side note re key exposure: Whatever we do, I think we should leave the door open for a world were we have DICE key derivation handled in a hardware block that binds the keys more tightly to the hardware. In that world, signing would be a call to the hardware block, providing a key handle and the data (or hash of it) to sign.
> 

Makes sense, and @glg-rv abstration of the signing entity paves the way for that model.
FWIW, TCG is working on a DICE spec extension that goes in that direction as well, where each layer would get a handle to its CDI, not the CDI itself. CDIs would then be hold in one or multiple safer entities (A RoT for example) and DICE implementations would use an (undefined) API to get their payloads signed by the relevant safer entity, from the handle they'd hold. That supported threat model is identical, but it allows for moving the trust somewhere else than in the DICE layers themselves.

sameo: > (sorry for pressing the resolve button accidentally)
> 
> Side note re key exposure: Whatever we do, I think we should leave the door open for a world were we have DICE key derivation handled in a hardware block that binds the keys more tightly to the hardware. In that world, signing would be a call to the hardware block, providing a key handle and the data (or hash of it) to sign.
> 

On key exposure and storage. I'm thinking about a model where we'd generate the TVM certificate once and delete the TSM CDI after that. It's fairly typical for CC attestation user APIs to allow for unlimited evidence generation calls but from the CSPs feedback we're getting it seems that in practice workloads usually only need one evidence that they immediately use to either get a signed attestation result (a token, the passport model) that they can re-use later on or get provisioned with secrets at e.g. boot time. The model where cc workloads would for some reasons need to get evidence from different CSRs, at different point of their lifecycle does not seem to be used in practice.

glg-rv: [...]

> But I agree with your point of not signing from U-mode, so implementing `Signer` as a call back to Salus makes sense to me. Is that something that you're planning to add as a separate PR?

Yes, next PR will be about this. It will add per-req ecalls and a way for umode to send/receive data. I have two equivalent implementations, but I'd like to discuss this in the next PR.

[...]

> > 
> > I might be misunderstanding these words, but what you mean is this: pass the CDI (secret), recrate the layer in U-mode at each certificate call, and leave the current rice unmodified. I understand how this could sound appealing, but I rather go towards a model where trust in U-mode is reduced -- otherwise, why not keep it in the hypervisor in the first place?.
> > The reason we build the certificate `from_raw_parts` is that U-mode will trust the hypervisor to set up things so that they come from the same layer (CDI_ID and the sign hypercall). I think this is more than fair, and requires no private key sharing.
> 
> CDI sharing is not a recommended practice, but if we move towards a model where the U-mode stack would not even be zero'ed then it's not even something we should discuss. Sorry for the noise.

Yes, we do that already. The idea of resetting the task at each call req was shot down very quickly during review. This is why we now have a loop in umode/src/main.rs. Also stack is now 176kb so that would be a lot of `bzero`ing.

Cool, thank you for the discussion!

Any further notes/comments/change requests re this PR?

sameo (approved):

abrestic-rivos (approved):

PR #233, opened 2023-02-09
Add htimedelta support.

rajnesh-kanwal:


riscv-regs/src/csrs/mod.rs
```diff
@@ -36,6 +36,8 @@ type VsiselectReg<R, const V: u64> = IndirectReadWriteRiscvCsr<
 >;
 
 pub struct CSR {
+    pub time: ReadWriteRiscvCsr<time::Register, CSR_TIME>,
+
```
abrestic-rivos, line 40: this is already available as `CSR.hpmcounter[1]`, no?

src/vm.rs
```diff
@@ -255,6 +263,7 @@ impl<T: GuestStagePagingMode> Vm<T> {
                 const_oid::db::rfc5912::ID_SHA_384,
             )
             .map_err(Error::AttestationManagerCreationFailed)?,
+            htimedelta: Wrapping(CSR.time.get_value()).neg().0,
```
abrestic-rivos, line 266: I think we actually want to latch at the point the first vCPU in the TVM is run. If we latch at TVM creation, then the host can effectively control the start time the TVM observes (which could possibly effect, e.g., RNG seeds).

src/vm.rs
```diff
@@ -1182,6 +1195,28 @@ impl<'a, T: GuestStagePagingMode> FinalizedVm<'a, T> {
...
+        // |                   |                   |
+        // |                   |                   |
+        // _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
+        // 0 1 2 3 4 5 6 7 8 9 1 1 1 1 1 1 1 1 1 1 2 2 2 2 2 2 2 2 2 2
+        //                     0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9
+        //                 Time
+        //
+        // For host the TVM's timedelta is
+        //     g_htimedelta = -get_cycle() = -(host time) = -(time + h_htimedelta)
+        //     g_htimedelta = - (20 - 10) = -10
```
abrestic-rivos, line 1211: A possible simplification would be to not use `htimedelta` for the host VM, i.e. host VM time == real time. But this works too, so up to you.

riscv-regs/src/csrs/defs.rs
```diff
@@ -9,6 +9,13 @@ use riscv_pages::Pfn;
 use tock_registers::register_bitfields;
```
atulkharerivos, line 9: Nit: Unrelated, but typo in commit message.

src/vm.rs
atishp04: So you want to do it in tvm_finalize ? That means the delta would not include the time to create a TVM (initial configuration/memory conversion/) which may be significant. Once we have a confidential page pool in place, conversion wouldn't be factor though.

src/vm.rs
atishp04: That would be problem in migration of the hosts. No ?  
Disclaimer: I still think host in VS mode to actually be useful along with host in HS mode in the future.

src/vm_cpu.rs
```diff
@@ -408,6 +408,8 @@ impl VmCpuArchState {
         scounteren.modify(scounteren::instret.val(1));
         regs.guest_regs.scounteren = scounteren.get();
 
+        regs.vs_csrs.htimedelta = htimedelta;
```
atishp04, line 411: Instead of setting the htimedelta in VmCpuArchState during creation, define an helper function to set it ?
We don't know how migration APIs will look like right now but htimedelta needs to be updated at that time.
We may or may not need to create VmCpuArchState during that time. 

If you like to keep it this way, leave a TODO comment at least.

src/vm.rs
abrestic-rivos: Not `tvm_finalize` -- the first time the TVM is run, i.e. `tvm_vcpu_run(0)`. The intention is for the TVM to observe a start time of ~0.

src/vm.rs
abrestic-rivos: Migrating the host VM sounds scary and I'm having a hard time thinking of a good use case for it :)

riscv-regs/src/csrs/mod.rs
rajnesh-kanwal: Didn't notice that. Thanks :+1: 
I personally find it less confusing to have `time`, `cycle` and `insret` to be separate from `hpmcounter` csrs. I am leaving this as it is for now and revert to using `hpmcounter[1]`.

src/vm.rs
atishp04: I don't think there is any use case right now. May be in the future. However, the cost of maintaining the htimedelta for the host VM is not much. IMO, It follows the standard convention of keeping htimedelta for any VM. Less confusion for future. 
Anyways, I am okay with your suggestion as well. Leave a good comment to explain the reasoning either way.

.project
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
abrestic-rivos, line 1: Unintended `git add`?

src/vm.rs
```diff
@@ -492,6 +496,27 @@ impl<'a, T: GuestStagePagingMode> FinalizedVm<'a, T> {
...
+            if let Ok(vcpu) = self.vm().vcpus.get_vcpu(i as u64) {
+                vcpu.set_htimedelta(htimedelta);
+            }
+        }
+
+        htimedelta
+    }
+
+    /// Sets htimedelta for the VM.
+    pub fn set_htimedelta(&self) {
```
abrestic-rivos, line 514: Does this need to be `pub`?

abrestic-rivos (commented):
Two minor comments, otherwise LGTM.

.project
rajnesh-kanwal: Oops. Fixed. Added it to gitignore as well.

src/vm.rs
rajnesh-kanwal: Fixed.

abrestic-rivos (approved):
LGTM. I'll let @atishp04 have another look before merging. Thanks!

atishp04 (approved):

atishp04:
LGTM.

PR #234, opened 2023-02-09
Basic test system to work with docker image

stillson:
This creates a basic system for working with the docker image rivos/rivos-qemu-docker which is what we use for acceptance testing with github actions.

Macros added to declare test pass and fail which will be detected when run under the overwatch.py program on the docker image.

The idea is that this can be easily extended to other parts of the system which can println!

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -22,7 +22,7 @@
 
 use core::alloc::{Allocator, GlobalAlloc, Layout};
 use core::ptr::NonNull;
-
+use test_system::*;
 extern crate alloc;
 
 mod asm;
@@ -308,6 +308,8 @@ extern "C" fn kernel_init(hart_id: u64, fdt_addr: u64) {
     SbiConsoleV01::set_as_console();
     println!("Salus: Boot test VM");
 
+    test_declare_pass!("Salus Boot", hart_id);
+
     // Safe because we trust that the firmware passed a valid FDT.
     let hyp_fdt =
         unsafe { Fdt::new_from_raw_pointer(fdt_addr as *const u8) }.expect("Failed to read FDT");
@@ -353,7 +355,10 @@ extern "C" fn kernel_init(hart_id: u64, fdt_addr: u64) {
         // unless we have support for per-mode filtering.
         println!("Sscofpmf support present");
         if let Err(e) = PmuInfo::init() {
+            test_declare_fail!("PMU counters");
             println!("PmuInfo::init() failed with {:?}", e);
+        } else {
+            test_declare_pass!("PMU counters");
         }
     }
     if cpu_info.has_vector() {
@@ -496,6 +501,7 @@ extern "C" fn kernel_init(hart_id: u64, fdt_addr: u64) {
     UmodeTask::setup_this_cpu().expect("Could not setup umode");
     // Do a NOP request to the U-mode task to check it's functional in this CPU.
     UmodeTask::send_req(u_mode_api::UmodeRequest::Nop).expect("U-mode not executing NOP");
+    test_declare_pass!("successful return from u-mode");
 
     // Now load the host VM.
     let host = HostVmLoader::new(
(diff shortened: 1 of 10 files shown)


src/main.rs
```diff
@@ -354,7 +359,10 @@ extern "C" fn kernel_init(hart_id: u64, fdt_addr: u64) {
         // unless we have support for per-mode filtering.
         println!("Sscofpmf support present");
         if let Err(e) = PmuInfo::init() {
+            test_fail!("sstc presence");
             println!("PmuInfo::init() failed with {:?}", e);
+        } else {
+            test_pass!("sstc presence")
         }
     }
```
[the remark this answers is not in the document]
stillson: Added test_declare_pass!() and test_runtest!()
test_declare_pass!() declares and passes in one macro

test_runtest!() executes a block. There is also test_assert!()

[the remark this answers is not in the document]
abrestic-rivos: Can we convert the remaining instances of separate `test_declare!` + `test_pass!` pairs to a single invocation? Having both strings matches is brittle and I'd like to avoid checking it in...

src/main.rs
```diff
@@ -354,7 +357,10 @@ extern "C" fn kernel_init(hart_id: u64, fdt_addr: u64) {
         // unless we have support for per-mode filtering.
         println!("Sscofpmf support present");
         if let Err(e) = PmuInfo::init() {
+            test_declare_fail!("sstc presence");
```
abrestic-rivos, line 360: s/sstc/sscofpmf

test-workloads/src/bin/guestvm.rs
```diff
@@ -169,6 +171,8 @@ pub fn test_vector() {
         }
     }
 
+    test_assert!(!should_panic, "writing/reading vector registers");
+
```
abrestic-rivos, line 175: This and the cert stuff below seems like good candidates for `test_runtest!`.

u-mode/src/main.rs
```diff
@@ -121,5 +122,6 @@ extern "C" fn task_main(cpuid: u64, shared_addr: u64, shared_size: u64) -> ! {
         "umode/#{}: U-mode Shared Region: {:016x} - {} bytes",
         cpuid, shared_addr, shared_size
     );
+    test_declare_pass!("umode start", cpuid);
```
abrestic-rivos, line 125: minor: I think it would be better to do this in Salus after the U-mode task calls back into Salus for its next task in `run_loop()`.

stillson: or both.

src/main.rs
stillson: fixed

test-workloads/src/bin/guestvm.rs
stillson: These seem a bit to complicated to drop into a single block.

abrestic-rivos: Seems pretty straightforward to me. Just make `test_vector` / `test_attestation` return `Pass` / `Fail` instead of panicing on error and then: `test_runtest!("...", || test_foo())`.

stillson: fixed

src/main.rs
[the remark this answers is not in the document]
stillson: fixed

abrestic-rivos:
Not sure what's going on with your git workflow, but this branch just looks like commits from main.

stillson:
You caught it in the middle of a rebase. Trying to fix it now.

src/main.rs
```diff
@@ -22,7 +22,7 @@
 
 use core::alloc::{Allocator, GlobalAlloc, Layout};
 use core::ptr::NonNull;
-
+use test_system::{test_declare, test_declare_fail, test_declare_pass, test_fail, test_pass};
```
abrestic-rivos, line 25: I'd just do `use test_system::*`

src/main.rs
```diff
@@ -345,6 +347,7 @@ extern "C" fn kernel_init(hart_id: u64, fdt_addr: u64) {
         // We don't implement or use the SBI timer extension and thus require Sstc for timers.
         panic!("CPU does not support Sstc");
     }
+
```
abrestic-rivos, line 350: unrelated whitespace change

abrestic-rivos (approved):

src/main.rs
stillson: fixed

PR #238, opened 2023-02-13
salus: Move the rice crate into a submodule

sameo:
In order to have a clearly defined dependency on a specific revision.

Signed-off-by: Samuel Ortiz <sameo@rivosinc.com>


abrestic-rivos (approved):
thanks!

PR #242, opened 2023-02-15
Salus: allow call-specific hypercalls in U-mode

glg-rv:
In preparation for modifying Attestation Manager to sign certificates, this change implements a stub hypervisor sign ecall for u-mode, that is accessible only during a guest request to get an evidence certificate.

Patch1: Fixes and makes error handling in U-mode requests a bit saner: task.run() will return an ExecError in case U-mode execution failed unexpectedly, and in the non error case will return the result of the operation (which is a Result<u64, UmodeApiError>). This helps simplify things and allows for an easier following of which error code we're handling.

### DIFF
--- a/src/umode.rs
+++ b/src/umode.rs
@@ -15,7 +15,7 @@ use riscv_elf::ElfMap;
 use riscv_regs::{Exception::UserEnvCall, GeneralPurposeRegisters, GprIndex, Readable, Trap, CSR};
 use s_mode_utils::print::*;
 use spin::Once;
-use u_mode_api::{Error as UmodeApiError, HypCall, TryIntoRegisters, UmodeRequest};
+use u_mode_api::{Error as UmodeApiError, HypCall, OpResult, TryIntoRegisters, UmodeRequest};
 
 /// Host GPR and which must be saved/restored when entering/exiting U-mode.
 #[derive(Default)]
@@ -233,6 +233,8 @@ global_asm!(
 pub enum Error {
     /// U-mode task returned an error while running.
     Exec(ExecError),
+    /// U-mode task completed execution but returned an error.
+    Request(UmodeApiError),
     /// Error while sharing data.
     HypMap(HypMapError),
 }
@@ -244,8 +246,8 @@ pub enum ExecError {
     UnexpectedTrap,
     /// Umode called panic.
     Panic,
-    /// Error in umode.
-    Umode(UmodeApiError),
+    /// U-mode API error.
+    Api(UmodeApiError),
 }
 
 // Entry for umode task.
@@ -267,13 +269,13 @@ impl UmodeTask {
     pub fn setup_this_cpu() -> Result<(), Error> {
         let arch = UmodeCpuArchState::default();
         let mut task = UmodeTask { arch };
-        task.reset().map_err(Error::Exec)?;
+        task.reset()?;
         // Install umode cpu state in the current cpu.
         PerCpu::this_cpu().set_umode_task(task);
         Ok(())
     }
 
-    fn reset(&mut self) -> Result<(), ExecError> {
+    fn reset(&mut self) -> Result<(), Error> {
         // Initialize umode CPU state to run at ELF entry.
         let mut arch = UmodeCpuArchState::default();
         // Unwrap okay: this is called after `Self::init()`.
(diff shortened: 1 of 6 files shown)


src/umode.rs
```diff
@@ -300,7 +300,18 @@ impl UmodeTask {
         Ok(())
     }
 
-    fn execute_request<T: DataInit>(
+    /// Execute a NOP request.
+    pub fn execute_nop() -> Result<(), Error> {
+        Self::execute_request::<u8>(UmodeRequest::Nop, None)?;
```
abrestic-rivos: I forgot to mention this before (and it's not really related to this patch), but the need for a dummy generic here is kind of silly. I guess that's because `T` must be `DataInit`, and the unit type isn't `DataInit`?

src/umode.rs
```diff
@@ -356,6 +372,14 @@ impl UmodeTask {
...
                 }
                 HypCall::NextOp(result) => ControlFlow::Break(Ok(result)),
+                hypc => {
+                    // Handle the rest through the caller-supplied `hypc_handler`.
+                    // All errors will immediately stop and reset U-mode.
+                    match hypc_handler(hypc) {
+                        Ok(()) => ControlFlow::Continue(()),
+                        Err(e) => ControlFlow::Break(Err(e)),
+                    }
+                }
```
abrestic-rivos, line 382: The intention of this is to only allow certain hypercalls for certain requests, right? e.g. only allowing signing for requests that have a need for it. Instead of a clousre, could you have a `UmodeRequest` method that takes a `HypCall` saying whether or not that `HypCall` is allowed? (or vice-versa)

abrestic-rivos (commented):
Overall LGTM, just wondering if there's a simpler way to dictate which hypercalls are allowed for a particular request.

src/umode.rs
mnissler-rivos: I also have a feeling that a function as a handler to filter what is allowed might not be the best way to manage the set of available hypervisor calls. I think you'll probably want something that's easy to compose (in many cases, you'll probably want the standard hypercalls plus a small number of additional calls). Perhaps we can learn from similar problems elsewhere, and do something similar to a syscall table here? In its simplest form it's just a map of HypCall to optional handler function, and it'd be simple enough to compose (start with the standard handler table, add in the signing one).

Now, you since a handler function is maximally generic, you can do all I say above with what you have, I guess I'm just wondering how this will look like in practice when we have more than one  umode request with more than one special-purpose restricted hypercalls...

(As often with my comments, just some unsolicited thinking-ahead, please feel free to ignore/defer until it becomes more relevant)

src/vm.rs
```diff
@@ -1680,7 +1680,18 @@ impl<'a, T: GuestStagePagingMode> FinalizedVm<'a, T> {
...
+                u_mode_api::HypCall::ExtSign {
+                    msg_addr: _,
+                    msg_size: _,
+                    sign_addr: _,
+                    sign_size: _,
+                } => {
+                    println!("Signing!");
+                    Ok(())
+                }
+                _ => Err(ExecError::UnhandledHypCall(hypc)),
```
mnissler-rivos, line 1693: So the signing request can no longer panic since HypCall::Panic would take this branch? Or am I missing something?

src/umode.rs
glg-rv: It's silly and annoying, agree. Option None needs a type, and that type must be DataInit. Any suggestion on how to get around this in a way that feels acceptable is okay. Whenever I meet an idiotic rust stance I go for the pragmatic way.

src/umode.rs
glg-rv: UmodeRequest is meant for the API and not for the machinery of the U-mode loop. So that will require a differfent type.

Having said this, I strongly prefer to use a closure because it will define data specific to the call. For example, which key to sign, how to get the key, etc., without requiring the umode module to be aware of all the internal specifics of complex requests.

We could be a bit more precise and define a "External Hypcall" type, so to better specify that this closure only specifies extra hypercalls, and doesn't filter the standard ones.

This is not filtering, this is defining some extra hypercalls that are more complex than simple management of the U-mode execution state.

src/vm.rs
glg-rv: ? An error return by this function will panic U-mode. I think you got this backward. I am not filtering all hypercalls to go through here. The external handler is called only after the basic hypercalls (panic, nextop, etc) checked.

mnissler-rivos: Ah, so that is what I had missed!

PR #245, opened 2023-02-16
Add hgatp emulation.

rajnesh-kanwal:
We need bare minimum hgatp emulation to allow the host to detect the VMIDLEN and the paging mode supported. All the writes are basically ignored and not conveyed to hardware csr.

Related change in Host: https://github.com/rajnesh-kanwal/linux/commit/85245a9bc1f390dc8de3e1ef50a7f54e0224610a

### DIFF
--- a/src/vm_cpu.rs
+++ b/src/vm_cpu.rs
@@ -16,7 +16,7 @@ use spin::{Mutex, MutexGuard, Once, RwLock};
 
 use crate::smp::PerCpu;
 use crate::vm::{MmioOpcode, MmioOperation, VmExitCause};
-use crate::vm_id::VmId;
+use crate::vm_id::*;
 use crate::vm_interrupts::{self, VmCpuExtInterrupts};
 use crate::vm_pages::{ActiveVmPages, FinalizedVmPages, PinnedPages};
 use crate::vm_pmu::VmPmuState;
@@ -105,6 +105,7 @@ pub struct GuestVsCsrs {
 pub struct GuestVirtualHsCsrs {
     hie: u64,
     hgeie: u64,
+    hgatp: u64,
 }
 
 /// CSRs written on an exit from virtualization that are used by the hypervisor to determine the cause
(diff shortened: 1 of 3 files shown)


src/vm_id.rs
```diff
@@ -57,6 +57,11 @@ impl VmIdTracker {
         self.current_version
     }
 
+    /// Returns the VMID length.
+    pub fn get_vmid_len(&self) -> u64 {
+        *self.vmid_bits.call_once(get_vmid_bits)
+    }
```
abrestic-rivos, line 63: Should we just put the # of VMID bits in a global `Once`?

src/vm_cpu.rs
```diff
@@ -969,6 +970,29 @@ impl<'vcpu, 'pages, 'host, T: GuestStagePagingMode> ActiveVmCpu<'vcpu, 'pages, '
                         Ok(0)
                     }
                 }
+                CSR_HGATP => {
+                    let vmid_len = PerCpu::this_cpu().vmid_tracker().get_vmid_len();
+                    let mode = self.active_pages().get_hgatp_mode();
```
abrestic-rivos, line 975: The generic `T` for guest stage paging mode is in scope here, so no need for a method call.

src/vm_id.rs
rajnesh-kanwal: Makes sense given it shouldn't change between cores. I will change this in next push.

src/vm_cpu.rs
rajnesh-kanwal: Fixed.

src/vm_id.rs
```diff
@@ -62,10 +62,15 @@ impl VmIdTracker {
         self.current_version
     }
 
+    /// Returns the VMID length.
+    pub fn get_vmid_len(&self) -> u64 {
+        // Unwrap okay: this is called after `init()`
+        *VMIDLEN.get().unwrap()
+    }
```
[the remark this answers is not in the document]
rajnesh-kanwal: My bad. Fixed.

abrestic-rivos (approved):

Issue #249, opened 2023-02-17
configure `reuse lint` pre-commit hook


kbroch-rivosinc:
https://reuse.software/dev/#pre-commit-hook

kbroch-rivosinc:
closed with https://github.com/rivosinc/salus/pull/252

PR #252, opened 2023-02-17
add `reuse lint` pre-commit hook check #249

kbroch-rivosinc:


.pre-commit-config.yaml
```diff
@@ -22,6 +22,12 @@ repos:
       - id: trailing-whitespace
         args: [--markdown-linebreak-ext=md]
 
+  # pull mirror of https://github.com/fsfe/reuse-tool
+  - repo: https://gitlab.ba.rivosinc.com/pub/it/reuse-tool
+    rev: v1.1.2
+    hooks:
+      - id: reuse
```
abrestic-rivos, line 29: This is still behind VPN, right? I guess the above one is too, so this was already broken for external users.

kbroch-rivosinc: Yes.  Probably makes sense to move all the URLs to the publically availble ones.   I'll change that.

abrestic-rivos (approved):

PR #255, opened 2023-02-20
VmInterrupts: Allow interrupt injection to new file while rebinding

rajnesh-kanwal:
When VCPU is not bound we sent those interrupts to swfile to be later copied to hw-interrupt file for injection.
When rebinding, the `rebind_imsic_begin()` sets the state to Rebinding but swfile is not set up for use as it's supposed to be setup on the previous PCPU which is done by `rebind_imsic_clone()`.

Given sw_file is used only between `rebind_imsic_clone()` and `rebind_imsic_finish()` it's possible that we miss interrupts between `rebind_imsic_begin()` and `rebind_imsic_clone()`.

Given we have the new interrupt file wired up, we can send interrupts directly to the new interrupt file in the Rebinding state and after transitioning to Cloned state, we are safe to use swfile.

Reported-by: Atish Patra <atishp@rivosinc.com>

### DIFF
--- a/src/vm_interrupts.rs
+++ b/src/vm_interrupts.rs
@@ -25,6 +25,7 @@ pub type Result<T> = core::result::Result<T, Error>;
 enum BindStatus {
     Binding(CpuId, ImsicFileId),
     Bound(CpuId, ImsicFileId),
+    // Rebinding has old CpuId and old ImsicFileId first and new ones later.
     Rebinding(CpuId, ImsicFileId, CpuId, ImsicFileId),
     Cloned(CpuId, ImsicFileId),
     Unbinding(CpuId, ImsicFileId),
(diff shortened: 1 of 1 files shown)


atishp04 (approved):

src/vm_interrupts.rs
```diff
@@ -25,6 +25,7 @@ pub type Result<T> = core::result::Result<T, Error>;
 enum BindStatus {
     Binding(CpuId, ImsicFileId),
     Bound(CpuId, ImsicFileId),
+    // Rebinding has old CpuId and old ImsicFileId first and new ones later.
     Rebinding(CpuId, ImsicFileId, CpuId, ImsicFileId),
```
abrestic-rivos, line 29: optional: you could name the fields to make this more obvious, e.g.:

```
Rebinding { old_cpu: CpuId, old_file: ImsicFileId, new_cpu: CpuId, new_file: ImsicFileId }
```

abrestic-rivos (approved):
Filed #256 for the CI breakage.

src/vm_interrupts.rs
rajnesh-kanwal: In my opinion I find it better as a tuple. With Rebinding being struct it will create mismatch between enum fields and the way it's accessed through out the code.

rajnesh-kanwal:
> Filed #256 for the CI breakage.

Thanks :+1:

PR #260, opened 2023-02-23
Test system for running unit tests for Salus.

stillson:
Adds `salus_test` and `run_salus_test` targets. These generate and run a binary under qemu which execute any function label with `#[test_case]` for unit testing. This allows unit testing under riscv emulation.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -19,6 +19,18 @@
     is_some_and,
     negative_impls
 )]
+#![feature(custom_test_frameworks)]
+#![test_runner(crate::test_runner)]
+#![reexport_test_harness_main = "test_main"]
+#![cfg_attr(test, allow(unused))]
+
+#[cfg(test)]
+fn test_runner(tests: &[&dyn Fn()]) {
+    println!("Running {} tests\n", tests.len());
+    for test in tests {
+        test();
+    }
+}
 
 use core::alloc::{Allocator, GlobalAlloc, Layout};
 use core::ptr::NonNull;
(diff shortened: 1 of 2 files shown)


src/asm.rs
```diff
@@ -4,4 +4,8 @@
 
 use core::arch::global_asm;
 
+#[cfg(not(test))]
 global_asm!(include_str!("start.S"));
+
+#[cfg(test)]
+global_asm!(include_str!("test_start.S"));
```
abrestic-rivos, line 11: do we really need a separate `start.S`?

src/main.rs
```diff
@@ -299,6 +310,28 @@ fn check_vector_width() {
     CSR.sstatus.read_and_clear_bits(sstatus::vs::Dirty.value);
 }
 
+/// The entry point for the test runner
+#[cfg(test)]
+#[no_mangle]
+extern "C" fn test_runner_kernel(_hart_id: u64, _fdt_addr: u64) {
```
abrestic-rivos, line 316: just call this `kernel_init` and put a `#[cfg(not(test))]` over the other `kernel_init`.

abrestic-rivos (commented):
cool!

src/asm.rs
stillson: I want to make sure that there is no way for testing code to end up in the shipping binary, and there is no way to use cfg in inline assembly.

src/main.rs
stillson: in that case, maybe we don't need another start.S

Will fix

abrestic-rivos (approved):

PR #261, opened 2023-02-24
Move to a more explicit memory sharing model

sboeuf:
In anticipation for the huge pages support, the way to share/unshare memory from a TVM is updated with a more explicit model that expects the host to drive the entire operation. Relying on some new TEE-Host functions, the host can invalidate and remove a specific range of memory based on the request from the TVM.

### DIFF
--- a/page-tracking/src/page_info.rs
+++ b/page-tracking/src/page_info.rs
@@ -152,6 +152,14 @@ impl PageInfo {
         matches!(self.state, PageState::Reserved)
     }
 
+    pub fn is_blocked(&self) -> bool {
+        matches!(self.state, PageState::Blocked(_))
+    }
+
+    pub fn is_blocked_shared(&self) -> bool {
+        matches!(self.state, PageState::BlockedShared(_, _))
+    }
+
     /// Returns the page type.
     pub fn mem_type(&self) -> MemType {
         self.mem_type
@@ -322,6 +330,17 @@ impl PageInfo {
         }
     }
 
+    /// Returns if the page is Blocked or BlockedShared and can complete the removal at
+    /// the given `tlb_version`.
+    pub fn is_removable(&self, tlb_version: TlbVersion) -> bool {
+        use PageState::*;
+        match self.state {
+            Blocked(version) => version < tlb_version,
+            BlockedShared(_, version) => version < tlb_version,
+            _ => false,
+        }
+    }
+
     /// Returns if the page is Shared
     pub fn is_shared(&self) -> bool {
         use PageState::*;
@@ -385,6 +404,64 @@ impl PageInfo {
             }
             // TODO: Reclaim pages that are converting but not yet fully converted?
             _ => Err(PageTrackingError::PageNotReclaimable),
+        }
+    }
+
+    /// Invalidates the page for promotion/demotion or removal from the TVM's address space.
+    pub fn block(&mut self, tlb_version: TlbVersion) -> PageTrackingResult<()> {
+        use PageState::*;
+        match self.state {
+            Mapped => {
+                self.state = Blocked(tlb_version);
+                Ok(())
+            }
+            Shared(rc) => {
+                self.state = BlockedShared(rc, tlb_version);
+                Ok(())
+            }
+            _ => Err(PageTrackingError::PageNotBlockable),
+        }
+    }
+
+    /// Unblocks an invalidated page.
+    pub fn unblock(&mut self) -> PageTrackingResult<()> {
+        use PageState::*;
+        match self.state {
+            Blocked(_) => {
+                self.state = Mapped;
+                Ok(())
+            }
+            BlockedShared(rc, _) => {
+                self.state = Shared(rc);
+                Ok(())
+            }
+            _ => Err(PageTrackingError::PageNotBlocked),
+        }
+    }
+
+    /// Removes an invalidated page.
+    pub fn remove(&mut self) -> PageTrackingResult<()> {
+        use PageState::*;
+        match self.state {
+            Blocked(_) => {
+                if self.owners.is_empty() {
+                    Err(PageTrackingError::OwnerUnderflow) // Can't pop the last owner.
+                } else {
+                    self.owners.pop().unwrap();
+                    self.state = Converted;
+                    Ok(())
+                }
+            }
+            BlockedShared(rc, _) => {
+                // Shared pages start with a RC of 1, so RC is always > 0 here
+                if let Some(rc) = rc.checked_sub(1) {
+                    self.state = if rc == 0 { Mapped } else { Shared(rc) };
+                    Ok(())
+                } else {
+                    Err(PageTrackingError::RefCountUnderflow)
+                }
+            }
+            _ => Err(PageTrackingError::PageNotBlocked),
         }
     }
 
--- a/page-tracking/src/page_info.rs
+++ b/page-tracking/src/page_info.rs
@@ -56,6 +56,14 @@ pub enum PageState {
 
     /// A Converted page that has been locked exclusively for assignment or reclaim.
     ConvertedLocked,
+
+    /// A Mapped page invalidated by the host at the given TLB version in the current owner's
+    /// address space.
+    Blocked(TlbVersion),
+
+    /// A Shared page invalidated by the host at the given TLB version in the current owner's
+    /// address space. The reference count of the page is recorded.
+    BlockedShared(u64, TlbVersion),
 }
 
 /// The maximum length for an ownership chain. Enough for the host VM to assign to a guest VM
@@ -117,15 +125,14 @@ impl PageInfo {
     pub fn owner(&self) -> Option<PageOwnerId> {
         use PageState::*;
         match self.state {
-            Converting(_) | Unassigning(_) | Converted | ConvertedLocked | Mapped | VmState
-            | Shared(_) => {
+            Reserved | Free | HypState => None,
+            _ => {
                 if !self.owners.is_empty() {
                     Some(self.owners[self.owners.len() - 1])
                 } else {
                     Some(PageOwnerId::hypervisor())
                 }
             }
-            _ => None,
         }
     }
 
@@ -151,10 +158,11 @@ impl PageInfo {
     }
 
     /// Pops the current owner if there is one, returning the page to the previous owner.
-    pub fn release(&mut self) -> PageTrackingResult<()> {
+    /// Takes into account the special case that the page was invalidated.
+    pub fn release(&mut self, was_invalidated: bool) -> PageTrackingResult<()> {
         use PageState::*;
         match self.state {
-            Mapped | VmState | Converted | Converting(_) | Unassigning(_) => {
+            Mapped | VmState | Converted | Converting(_) | Unassigning(_) | Blocked(_) => {
                 if self.owners.is_empty() {
                     Err(PageTrackingError::OwnerUnderflow) // Can't pop the last owner.
                 } else {
@@ -164,11 +172,18 @@ impl PageInfo {
                 }
             }
             ConvertedLocked => Err(PageTrackingError::PageLocked),
-            Shared(rc) => {
+            BlockedShared(rc, tlbv) if !was_invalidated => {
+                if let Some(rc) = rc.checked_sub(1) {
+                    self.state = BlockedShared(rc, tlbv);
+                    Ok(())
+                } else {
+                    Err(PageTrackingError::RefCountUnderflow)
+                }
+            }
+            Shared(rc) | BlockedShared(rc, _) => {
                 // Shared pages start with a RC of 1, so RC is always > 0 here
                 if let Some(rc) = rc.checked_sub(1) {
                     self.state = if rc == 0 { Mapped } else { Shared(rc) };
-                    // Unwrap ok: Shared pages must have an owner
                     Ok(())
                 } else {
                     Err(PageTrackingError::RefCountUnderflow)
--- a/page-tracking/src/page_tracker.rs
+++ b/page-tracking/src/page_tracker.rs
@@ -355,6 +355,31 @@ impl PageTracker {
         info.complete_unassignment(tlb_version)
     }
 
+    /// Move the page to a blocked state.
+    pub fn block_page<P: InvalidatedPhysPage>(
+        &self,
+        page: P,
+        tlb_version: TlbVersion,
+    ) -> Result<()> {
+        let mut page_tracker = self.inner.lock();
+        let info = page_tracker.get_mut(page.addr()).unwrap();
+        info.block(tlb_version)
+    }
+
+    /// Move the page back to the state it was before being blocked.
+    pub fn unblock_page(&self, addr: SupervisorPageAddr) -> Result<()> {
+        let mut page_tracker = self.inner.lock();
+        let info = page_tracker.get_mut(addr).unwrap();
+        info.unblock()
+    }
+
+    /// Remove the page, moving it to either "Mapped" or "Converted" state.
+    pub fn remove_page(&self, addr: SupervisorPageAddr) -> Result<()> {
+        let mut page_tracker = self.inner.lock();
+        let info = page_tracker.get_mut(addr).unwrap();
+        info.remove()
+    }
+
     /// Releases an exclusive reference to a "ConvertedLocked" page.
     pub fn unlock_page<P: ConvertedPhysPage>(&self, page: P) -> Result<()> {
         let mut page_tracker = self.inner.lock();
@@ -480,6 +505,61 @@ impl PageTracker {
             info.owner() == Some(owner)
                 && info.mem_type() == MemType::Ram
                 && info.state() == PageState::VmState
+        } else {
+            false
+        }
+    }
+
+    /// Returns true if `addr` is a page owned by `owner` and it's in a "Blocked" state with
+    /// type `mem_type`, or if `addr` is a page not owned by `owner` and it's in a "SharedBlocked"
+    /// state with type `mem_type`.
+    pub fn is_blocked_page(
+        &self,
+        addr: SupervisorPageAddr,
+        owner: PageOwnerId,
+        mem_type: MemType,
+    ) -> bool {
+        let mut page_tracker = self.inner.lock();
+        if let Ok(info) = page_tracker.get(addr) {
+            info.mem_type() == mem_type
+                && ((info.owner() == Some(owner) && info.is_blocked())
+                    || (info.owner() != Some(owner) && info.is_blocked_shared()))
+        } else {
+            false
+        }
+    }
+
+    /// Returns true if the page is considered blockable.
+    pub fn is_blockable_page(
+        &self,
+        addr: SupervisorPageAddr,
+        owner: PageOwnerId,
+        mem_type: MemType,
+    ) -> bool {
+        let mut page_tracker = self.inner.lock();
+        if let Ok(info) = page_tracker.get(addr) {
+            info.mem_type() == mem_type
+                && ((info.owner() == Some(owner) && info.state() == PageState::Mapped)
+                    || (info.owner() != Some(owner) && info.is_shared()))
+        } else {
+            false
+        }
+    }
+
+    /// Returns true if the page is considered removable.
+    pub fn is_removable_page(
+        &self,
+        addr: SupervisorPageAddr,
+        owner: PageOwnerId,
+        mem_type: MemType,
+        tlb_version: TlbVersion,
+    ) -> bool {
+        let mut page_tracker = self.inner.lock();
+        if let Ok(info) = page_tracker.get(addr) {
+            info.is_removable(tlb_version)
+                && info.mem_type() == mem_type
+                && ((info.owner() == Some(owner) && info.is_blocked())
+                    || (info.owner() == Some(PageOwnerId::host()) && info.is_blocked_shared()))
         } else {
             false
         }
--- a/page-tracking/src/page_tracker.rs
+++ b/page-tracking/src/page_tracker.rs
@@ -57,6 +57,10 @@ pub enum Error {
     RefCountOverflow,
     /// The ref count was already 0.
     RefCountUnderflow,
+    /// This page is not in a state it can be blocked.
+    PageNotBlockable,
+    /// Attempt to unblock, promote, demote or remove a page that is not blocked.
+    PageNotBlocked,
 }
 
 /// Holds the result of page tracking operations.
@@ -241,7 +245,7 @@ impl PageTracker {
     pub fn release_page<P: PhysPage>(&self, page: P) -> Result<()> {
         let mut page_tracker = self.inner.lock();
         let info = page_tracker.get_mut(page.addr()).unwrap();
-        info.release()?;
+        info.release(false)?;
         Ok(())
     }
 
@@ -254,7 +258,25 @@ impl PageTracker {
         if info.owner() != Some(owner) && !info.is_shared() {
             return Err(Error::OwnerMismatch);
         }
-        info.release()?;
+        info.release(false)?;
+        Ok(())
+    }
+
+    /// Releases the page at `addr` back to its previous owner if it's currently owned by `owner`
+    /// and is in a releasable state. Takes into account the special case where the page was
+    /// invalidated.
+    pub fn release_invalidated_page_by_addr(
+        &self,
+        addr: SupervisorPageAddr,
+        owner: PageOwnerId,
+    ) -> Result<()> {
+        let mut page_tracker = self.inner.lock();
+        let info = page_tracker.get_mut(addr)?;
+        // Shared pages might be owned by the parent
+        if info.owner() != Some(owner) && !info.is_shared() {
+            return Err(Error::OwnerMismatch);
+        }
+        info.release(true)?;
         Ok(())
     }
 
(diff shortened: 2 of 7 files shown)


sboeuf:
This PR depends on https://github.com/rivosinc/sbi-rs/pull/12 where the extended TEE-Host API is defined.
Also, this requires some additional tests in `tellus`/`guestvm` to validate it's functional, but I posted the PR to get some early feedback.

sboeuf:
> Also, this requires some additional tests in `tellus`/`guestvm` to validate it's functional, but I posted the PR to get some early feedback.

I've just updated the tellus VM so that it follows the new model for completing memory sharing and memory unsharing from the guest.

src/vm_pages.rs
```diff
@@ -1264,6 +1319,75 @@ impl<'a, T: GuestStagePagingMode> FinalizedVmPages<'a, T> {
...
+        .ok_or(Error::UnalignedAddress)?;
+        // Check the address range lies within a "ConfidentialRemovable" region of guest
+        // physical address space and update the region type to "Shared".
+        self.inner.regions.write().update_in_place(
+            page_addr,
+            end,
+            VmRegionType::ConfidentialRemovable,
+            VmRegionType::Shared,
+            true,
+        )?;
```
abrestic-rivos, line 1246: need to check that all the pages have been removed first

page-tracking/src/page_tracker.rs
```diff
@@ -338,6 +338,31 @@ impl PageTracker {
...
+        let info = page_tracker.get_mut(addr).unwrap();
+        info.unblock()
+    }
+
+    /// Remove the page, moving it to either "Mapped" or "Converted" state.
+    pub fn remove_page(&self, addr: SupervisorPageAddr) -> Result<()> {
+        let mut page_tracker = self.inner.lock();
+        let info = page_tracker.get_mut(addr).unwrap();
+        info.remove()
+    }
```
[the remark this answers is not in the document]
sboeuf: If you confirm you're okay replacing the `PageInfo::remove()` behavior with `PageInfo::release()` behavior, then of course I'll replace `remove_page()` with `release_page()`.

riscv-page-tables/src/page_table.rs
```diff
@@ -1046,6 +1053,66 @@ impl<T: PagingMode> GuestStagePageTable<T> {
...
+            match inner.walk(va.into()) {
+                Invalidated(pte) => {
+                    if !pte.level().is_leaf() {
+                        return Err(Error::PageSizeNotSupported(pte.level().leaf_page_size()));
+                    }
+                    if !pred(pte.page_addr()) {
+                        return Err(Error::PredicateFailed);
+                    }
+                }
+                Unused(_) => continue,
```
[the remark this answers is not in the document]
sboeuf: Oh so do you also expect that we should be blocking a fully populated range? I'm asking because then I would use the non-sparse version of `invalidate_range()`.

src/vm_pages.rs
```diff
@@ -1751,6 +1752,100 @@ impl<'a, T: GuestStagePagingMode> FinalizedVmPages<'a, T> {
         Ok(())
     }
 
+    /// Invalidates a page range.
+    pub fn block_pages(&self, page_addr: GuestPageAddr, len: u64) -> Result<()> {
+        let invalidated = self
+            .inner
+            .root
+            .invalidate_range_sparse(page_addr, len, |addr| {
```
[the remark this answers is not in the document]
sboeuf: Got it! That makes sense!

src/vm_pages.rs
```diff
@@ -1751,6 +1752,100 @@ impl<'a, T: GuestStagePagingMode> FinalizedVmPages<'a, T> {
...
+            .validate_range(page_addr, len, |addr| {
+                self.inner.page_tracker.is_blocked_page(
+                    addr,
+                    self.inner.page_owner_id,
+                    MemType::Ram,
+                ) || self
+                    .inner
+                    .page_tracker
+                    .is_blocked_shared_page(addr, MemType::Ram)
+            })
```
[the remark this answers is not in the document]
sboeuf: No that's the only place where we use it, so I can merge this into a single helper. And I think this will be the same for `promote`/`demote` functions, all they'll care about is if the page is blocked, doesn't matter if it really is `BlockedShared` or `Blocked`.

src/vm_pages.rs
```diff
@@ -1751,6 +1752,100 @@ impl<'a, T: GuestStagePagingMode> FinalizedVmPages<'a, T> {
...
+            .inner
+            .regions
+            .read()
+            .contains(page_addr, end, VmRegionType::ConfidentialRemovable)
+            && !self
+                .inner
+                .regions
+                .read()
+                .contains(page_addr, end, VmRegionType::SharedRemovable)
+        {
```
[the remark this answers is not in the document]
sboeuf: Yes I'll improve this.

src/vm_pages.rs
```diff
@@ -1751,6 +1752,100 @@ impl<'a, T: GuestStagePagingMode> FinalizedVmPages<'a, T> {
...
+    pub fn block_pages(&self, page_addr: GuestPageAddr, len: u64) -> Result<()> {
+        let invalidated = self
+            .inner
+            .root
+            .invalidate_range_sparse(page_addr, len, |addr| {
+                self.inner
+                    .page_tracker
+                    .is_mapped_page(addr, self.inner.page_owner_id, MemType::Ram)
+                    || (self.inner.page_tracker.is_shared_page(addr, MemType::Ram)
+                        && self.inner.page_tracker.is_owned(addr, PageOwnerId::host()))
```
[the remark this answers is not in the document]
sboeuf: Out of curiosity, can you elaborate why? Do we expect a shared page to be owned by something else than the host?

src/vm_pages.rs
```diff
@@ -1751,6 +1752,100 @@ impl<'a, T: GuestStagePagingMode> FinalizedVmPages<'a, T> {
...
+            .root
+            .unmap_range(page_addr, len, |addr| {
+                self.inner.page_tracker.is_removable_page(
+                    addr,
+                    self.inner.page_owner_id,
+                    MemType::Ram,
+                    tlb_version,
+                )
+            })
+            .map_err(Error::Paging)?;
```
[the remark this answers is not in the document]
sboeuf: Oh so you suggest I should lock the regions at the top of this function with `let regions = self.inner.regions.read()` so that no update can happen before the unmapping is performed?

src/vm.rs
```diff
@@ -517,6 +519,47 @@ impl<'a, T: GuestStagePagingMode> FinalizedVm<'a, T> {
             .call_once(|| self.set_vcpu_htimedelta());
     }
 
+    // Completes any pending operation from the host for this VM.
+    fn complete_pending_op(&self, active_vcpu: &mut ActiveVmCpu<T>) -> EcallResult<u64> {
+        if let Some(pending_op) = active_vcpu.take_pending_op() {
```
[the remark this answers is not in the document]
sboeuf: Ok I'll try that!
What's the clean way to retrieve the value of the A0 register with this approach? I was using the `active_vcpu` but with your proposal I won't have access to it when needed.

src/vm_pages.rs
```diff
@@ -487,6 +487,67 @@ impl VmRegionList {
...
+            to_reserve += 1;
+        }
+        if to_split.end != end {
+            to_reserve += 1;
+        }
+        if self.regions.remaining_capacity() < to_reserve {
+            return Err(Error::InsufficientVmRegionSpace);
+        }
+
+        // Now do the split, marking the range to be updated as 'Updating'.
```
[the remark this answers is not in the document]
sboeuf: Yes I'll remove it.

src/vm_pages.rs
```diff
@@ -1264,6 +1319,75 @@ impl<'a, T: GuestStagePagingMode> FinalizedVmPages<'a, T> {
...
+            page_addr,
+            end,
+            VmRegionType::Confidential,
+            VmRegionType::ConfidentialRemovable,
+            false,
+        )?;
+
+        // Should we go further here, by getting the actual list of pages that will be
+        // invalidated, so that we can know for sure if we can return to the guest.
+        Ok(self.inner.tlb_tracker.current_version().increment())
```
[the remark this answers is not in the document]
sboeuf: I kept incrementing the TLB (following what was already there in `share_mem_region_begin()`) because that's the way to tell the calling function `handle_tee_guest_msg()` that we have a blocking ecall. But I guess with the new design we can always assume a `ShareMemory` or `UnshareMemory` call is a blocking ecall. And that would allow to get rid of this logic here.

page-tracking/src/page_info.rs
```diff
@@ -117,8 +125,15 @@ impl PageInfo {
...
-            | Shared(_) => {
+            Converting(_)
+            | Unassigning(_)
+            | Converted
+            | ConvertedLocked
+            | Mapped
+            | VmState
+            | Shared(_)
+            | Blocked(_)
+            | BlockedShared(_, _) => {
```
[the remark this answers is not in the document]
sboeuf: Done!

src/vm_pages.rs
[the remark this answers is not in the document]
sboeuf: Ah I see! thanks for the clarification.

src/vm_pages.rs
sboeuf: Ah yes of course, given we've cleared the PTE bits through `remove_pages`, we expect every PTE from the range to be marked as `Unused`.

src/vm.rs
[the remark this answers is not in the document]
sboeuf: IIUC, the thing you want to avoid is having A0 being read twice, right?

Could we pass a reference of `FinalizedVm` to `VmCpu::complete_pending_op()`? This way, we would invoke `complete_share_mem_region()` from inside VmCpu.
That being said, I'm not a big fan of this approach because `FinalizedVm` is the superset that contains `VmCpu` and I'm not sure it's clean to pass a reference down to `VmCpu`.

Another approach would be to call `VmCpu::complete_pending_op()` first, and expect it to return the value of the A0 register along with the `pending_op`, so that from `vm.rs`, we have all the information we need to call into `complete_share_mem_region()`. The only thing that is unclear to me with this approach is how we can make the return type of `VmCpu::complete_pending_op()` generic enough so that we wouldn't make it specific to the case we're trying to solve here.

src/vm_pages.rs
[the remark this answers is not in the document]
sboeuf: I patched this and then realized the `tellus`/`guestvm` test was broken. Basically what happens is that when the `guestvm` first tries to share some untouched memory, then `tellus` invokes `block_pages()`, but then `salus` only finds `Unused` PTEs from the PT walk.
Do I need to reuse the logic that was present in `share_mem_region_begin()` so that trying to share a range if the page is not mapped would not cause a breaking ecall but instead a resumable one? Or instead, do you think the page should have been mapped by `tellus` beforehand?

src/vm.rs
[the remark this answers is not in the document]
abrestic-rivos: > IIUC, the thing you want to avoid is having A0 being read twice, right?

to be more precise, having the A0 that's being acted upon be different from the one that ends up getting written to the guest's register file. but yes, only reading it once would prevent that.

> That being said, I'm not a big fan of this approach because FinalizedVm is the superset that contains VmCpu and I'm not sure it's clean to pass a reference down to VmCpu.

agreed.

> Actually we can keep the take_pending_op() logic associated with calling VmCpu::complete_pending_op() first. That would simplify what needs to be returned by VmCpu::complete_pending_op() (still need to figure out how to make it generic enough).

i suppose that works too. another thing that comes to mind is using a closure that `complete_pending_op()` invokes on the ecall and the contents of A0. something like:

```
pub fn try_complete_pending_op<F, E>(&mut self, ecall_completion: F) -> Result<(), E>
where
    F: Fn(&SbiMessage, SbiReturnType) -> Result<(), E>
{
    match self.arch.pending_op {
        Some(Mmio(mmio_op)) => {
            // handle this same as before
        }
        Some(Ecall(msg)) => {
            let sbi_ret = ...;
            ecall_completion(&msg, sbi_ret)?;
            self.set_ecall_result(sbi_ret);
        }
        None => (),
    }
    self.arch.pending_op = None;
    Ok(())
}
```

and then call `try_complete_pending_op()` before entering the `active_vcpu.run()` loop in `run_vcpu()`.

src/vm_pages.rs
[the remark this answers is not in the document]
abrestic-rivos: ah, in the case that the region is totally unpopulated to begin with, the host does not have to do the block/remove sequence since there's nothing to remove. so in that case the host can just put the return value in A0 and run the vCPU again.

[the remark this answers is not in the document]
sboeuf: Alright!

src/vm.rs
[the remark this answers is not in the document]
sboeuf: Okay I'll implement this way!

sboeuf:
@abrestic-rivos I think I addressed all of your comments.
BTW, are we planning on merging https://github.com/rivosinc/sbi-rs/pull/12? I'm asking because right now this PR relies on my personal fork of `sbi-rs`.

atishp04:
I haven't looked into the PR in details. As per my understanding, this PR will break the existing setup running TVMs as the share memory design is revamped. I was planning to send the RFC with the existing sharing APIs soon. Can we not merge this until host side implementation is also complete ?

The other option is to wait for the sharing APIs to settle down and complete implementation on host/salus side before sending the RFC.

page-tracking/src/page_info.rs
```diff
@@ -164,7 +171,7 @@ impl PageInfo {
                 }
             }
             ConvertedLocked => Err(PageTrackingError::PageLocked),
-            Shared(rc) => {
+            Shared(rc) | BlockedShared(rc, _) => {
```
abrestic-rivos, line 182: so there's a bit of a wrinkle here i just realized...

we have two ways to end up with a `SharedBlock` page in `release()`:
1. a TVM that had the page blocked was killed before the page was unblocked or removed
2. a TVM that had the shared page mapped (while it was blocked in another TVM) was killed

the behavior here is correct for case (1). for case (2), we want to leave it as `BlockedShared` but decrement the reference count. seems like we need to know whether or not the page was invalidated (indicating it was blocked) when releasing the page in order to differentiate between the two cases.

src/vm_pages.rs
```diff
@@ -1334,6 +1463,79 @@ impl<'a, T: GuestStagePagingMode> FinalizedVmPages<'a, T> {
         Ok(())
     }
 
+    /// Converts the specified memory region from shared to confidential. Returns the TLB version
+    /// at which the conversion will be completed.
+    pub fn unshare_mem_region(&self, page_addr: GuestPageAddr, len: u64) -> Result<()> {
```
abrestic-rivos, line 1278: the `*unshare_mem_region` methods look pretty similar to the `*share_mem_region` ones except the region types are flipped. is it possible to deduplicate these a bit?

src/vm.rs
```diff
@@ -2359,14 +2397,13 @@ impl<'a, T: GuestStagePagingMode> FinalizedVm<'a, T> {
...
-                    Ok(tlbv) if tlbv > self.vm_pages().min_tlb_version() => EcallAction::Break(
-                        VmExitCause::BlockingEcall(SbiMessage::TeeGuest(guest_func), tlbv),
-                        SbiReturn::success(0),
-                    ),
                     Ok(_) => EcallAction::Break(
-                        VmExitCause::ResumableEcall(SbiMessage::TeeGuest(guest_func)),
+                        VmExitCause::BlockingEcall(
+                            SbiMessage::TeeGuest(guest_func),
+                            self.vm_pages().min_tlb_version(),
+                        ),
```
abrestic-rivos, line 2406: now that the host is responsible for explicitly blocking & removing the pages, we should remove this logic to block until a specific tlb version is met.

src/vm_pages.rs
```diff
@@ -304,12 +304,6 @@ enum VmRegionType {
     Imsic,
```
abrestic-rivos, line 304: something to consider, not necessarily for this PR, is to adapt the whole IMSIC bind/unbind/rebind bits to internally use the new block/remove flow. that would allow us to completely get rid of the existing "unassignment" flow.

page-tracking/src/page_info.rs
sboeuf: I see... Should we carry the TVM id through the `BlockedShared()` state? That could help us know what to do. Or did you have another idea about how to do it?

sboeuf:
> I haven't looked into the PR in details. As per my understanding, this PR will break the existing setup running TVMs as the share memory design is revamped. I was planning to send the RFC with the existing sharing APIs soon. Are we planning to merge this PR soon ? We need to point to a working salus branch instead of main in the docs in that case.
> 
> The other option is to wait for the sharing APIs to settle down and complete implementation on host/salus side before sending the RFC. That works for me as well as it may be little odd to change the APIs between between RFC and v1 :)

IMO it'd be best to update the RFC to match the new design, otherwise we might confuse the reviewers upstream, and this might increase the delay in getting the RFC agreed on.

page-tracking/src/page_info.rs
abrestic-rivos: i did think about that, but it grows the `PageInfo` struct by 64-bits. that's a non-trivial amount of overhead considering the struct is repeated for every page of memory in the system.

another option is to have separate, e.g.`release_present()` vs `release_invalidated()`, calls for the two paths.

src/vm.rs
abrestic-rivos: i think we can drop the `Blocked` state entirely now that we handle that logic in `complete_pending_op()`.

page-tracking/src/page_info.rs
sboeuf: >another option is to have separate, e.g.release_present() vs release_invalidated(), calls for the two paths.

Ok I'm gonna try this approach

src/vm.rs
sboeuf: Sounds good

page-tracking/src/page_info.rs
sboeuf: I've added the commit https://github.com/rivosinc/salus/pull/261/commits/f1bdd023ea5d41f82b377909d1724c664ad54e66 to handle this issue, please let me know if this is covering what you had in mind.

page-tracking/src/page_info.rs
```diff
@@ -159,6 +159,24 @@ impl PageInfo {
 
     /// Pops the current owner if there is one, returning the page to the previous owner.
     pub fn release(&mut self) -> PageTrackingResult<()> {
```
abrestic-rivos, line 169: nit: i think it'd be slightly cleaner if you had a single release with a `was_invalid: bool` param that was used to determine the behavior in the `BlockedShared` state, rather than having `release()` fall through to `release_invalidated()`.

page-tracking/src/page_info.rs
abrestic-rivos: yes, that looks right to me. squash with this commit?

src/vm_pages.rs
```diff
@@ -1751,6 +1750,93 @@ impl<'a, T: GuestStagePagingMode> FinalizedVmPages<'a, T> {
...
+            .root
+            .invalidate_range(page_addr, len, |addr| {
+                self.inner
+                    .page_tracker
+                    .is_mapped_page(addr, self.inner.page_owner_id, MemType::Ram)
+                    || (self.inner.page_tracker.is_shared_page(addr, MemType::Ram)
+                        && !self
+                            .inner
+                            .page_tracker
+                            .is_owned(addr, self.inner.page_owner_id))
```
abrestic-rivos, line 1668: kind of pre-existing and not something that needs to be fixed in this PR, but this logic should probably be in a single `PageTracker` call to avoid bouncing of that lock so much.

src/vm_pages.rs
```diff
@@ -1262,6 +1318,151 @@ impl<'a, T: GuestStagePagingMode> FinalizedVmPages<'a, T> {
...
+    pub fn share_mem_region(&self, page_addr: GuestPageAddr, len: u64) -> Result<()> {
+        self.do_share_mem_region(page_addr, len, true)
+    }
+
+    /// Converts the specified memory region from shared to confidential.
+    pub fn unshare_mem_region(&self, page_addr: GuestPageAddr, len: u64) -> Result<()> {
+        self.do_share_mem_region(page_addr, len, false)
+    }
+
+    pub fn do_share_mem_region(
```
abrestic-rivos, line 1214: nit: maybe `do_convert_*` for these helpers instead of `do_share_*`? though i guess "convert" is overloaded...

src/vm_cpu.rs
```diff
@@ -820,10 +817,10 @@ impl<'vcpu, 'pages, 'host, T: GuestStagePagingMode> ActiveVmCpu<'vcpu, 'pages, '
 
         use VmExitCause::*;
         match cause {
-            ResumableEcall(msg) | FatalEcall(msg) | BlockingEcall(msg, _) => {
+            ResumableEcall(msg) | FatalEcall(msg) => {
                 self.report_ecall_exit(msg);
             }
-            ForwardedEcall(msg) => {
+            ForwardedEcall(msg) | BlockingEcall(msg) => {
```
abrestic-rivos, line 823: i think we can totally get rid of the `BlockingEcall` exit cause, right?

src/vm.rs
```diff
@@ -533,6 +533,44 @@ impl<'a, T: GuestStagePagingMode> FinalizedVm<'a, T> {
         let mut active_vcpu = vcpu
             .activate(self.vm_pages(), host_context)
             .map_err(|_| EcallError::Sbi(SbiError::InvalidParam))?;
+
+        // Complete a possible pending operation
+        active_vcpu.try_complete_pending_op(|sbi_msg, sbi_ret| -> EcallResult<()> {
```
abrestic-rivos, line 538: maybe hoist the contents of the closure into it's own function?

abrestic-rivos (commented):
just a couple minor comments left, otherwise lgtm -- thanks!

src/vm_pages.rs
sboeuf: Yes that's fine I can use `do_convert_region_*`

src/vm.rs
sboeuf: This will be cleaner yes.

src/vm_cpu.rs
sboeuf: We use it for setting up the `pending_op`. Should I replace it with `ForwardedEcall` so that I can fully remove it?

sboeuf: As per our offline conversation, I've replaced BlockingEcall with ForwardedEcall.

src/vm_pages.rs
```diff
@@ -616,7 +615,7 @@ impl<'a, T: GuestStagePagingMode, M> VmPagesMapper<'a, T, M> {
             .checked_add_pages(num_pages)
             .ok_or(Error::AddressOverflow)?;
         let regions = vm_pages.regions.read();
-        if !regions.contains(page_addr, end, region_type) {
+        if !regions.contains(page_addr, end, |r| r == region_type) {
             return Err(Error::InvalidMapRegion);
         }
         let mapper = vm_pages
```
abrestic-rivos, line 566: per our offline conversation, i think we should allow page faults in `*Removable` regions to continue to be serviceable as if they were the original region type. in other words:
- page faults taken in `*Removable` regions should not be fatal
- host should be allowed to add shared pages to `SharedRemovable` regions, and confidential zero pages to `ConfidentialRemovable`.

sboeuf: Sounds good! What about measured pages?

sboeuf: Please see the new commit handling this: https://github.com/rivosinc/salus/pull/261/commits/7c997b79e15b00b9a701a501286145f07b947054

abrestic-rivos: measured pages can only be added prior to tvm finalization, so it's impossible for those to be in a `*Removable` region anyway (which you've handled correctly.)

abrestic-rivos (approved):
lgtm! @atishp04 / @rajnesh-kanwal how do you want to handle submitting this? do you want to take a stab at getting the kernel side working first?

rajnesh-kanwal:
> lgtm! @atishp04 / @rajnesh-kanwal how do you want to handle submitting this? do you want to take a stab at getting the kernel side working first?

The change overall looks good to me. I don't think there will be problems while supporting it in KVM. Just doing a final review and once done I think we can merge this. I will then start to add host side support in KVM.

test-workloads/src/bin/tellus.rs
```diff
@@ -954,11 +979,26 @@ extern "C" fn kernel_init(hart_id: u64, fdt_addr: u64) {
...
+                                .is_some()
+                                && unshared_spa_range.is_none()
+                            {
+                                unshared_spa_range = Some(Range {
+                                    start: zero_pages_base + zero_pages_added * PAGE_SIZE_4K,
+                                    end: zero_pages_base
+                                        + zero_pages_added * PAGE_SIZE_4K
+                                        + PAGE_SIZE_4K,
+                                });
+                            }
```
rajnesh-kanwal, line 1000: Can you just add a small comment for this part explaining why we are saving unshared_spa_range here.
As per my understanding this is so that we can check if we need a fence or not and run the if block on line `810`. Right?
Seemed confusing to me at the start.

rajnesh-kanwal (approved):
One minor comment otherwise LGTM.

test-workloads/src/bin/tellus.rs
sboeuf: Yes you understood that well. Of course I can add a comment to make things clearer.

PR #220, opened 2023-01-27
Initial bazel build system for salus

stillson:
Created initial bazel based build system for salus. Will elaborate further later, just is mainly to get feedback.

PR #262, opened 2023-02-27
Consolidate 'spin' dependencies

abrestic-rivos:
This should reduce some of the unnecessary boilerplate in #220 due to pulling in `spin` with different features in different crates.


stillson:
LGTM
ship it

PR #264, opened 2023-02-28
VmCpu: Fix rd and rs2 encoding for load and store instructions.

rajnesh-kanwal:
For load instruction, rd is encoded in bits 7:11 and for store instruction, rs2 is encoded in bits 20:24.


abrestic-rivos (approved):
whoops, good catch!

PR #270, opened 2023-03-03
Added script to format all rust files

stillson:
Also, `rustfmt` can be called directly.

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -124,11 +124,11 @@ objcopy_to_object(
 rust_binary(
     name = "salus",
     srcs = glob(["src/*.rs"]),
-    data = glob(["src/*.S"]) + [":umode_to_object"],
-    linker_script = ":l_rule",
+    data = glob(["src/*.S"]) + [":umode_to_object", ":l_rule"],
     rustc_flags = [
         "-Ctarget-feature=+v",
         "--codegen=link-arg=-nostartfiles",
+        "-Clink-arg=-T$(location //:l_rule)"
     ],
     deps = [
         "//attestation",


abrestic-rivos:
can't this be done with a `bazel run`? see: https://bazelbuild.github.io/rules_rust/rust_fmt.html#setup

stillson:
No. It's broken for no_std builds, and It seems wrong for bazel to actually modify the source code. I did try this, and tried writing a rule, but realized that it's not obvious how to get bazel to modify source in place.

abrestic-rivos:
> No. It's broken for no_std builds, and It seems wrong for bazel to actually modify the source code. I did try this, and tried writing a rule, but realized that it's not obvious how to get bazel to modify source in place.

If you look at the page I linked it doesn't suggest writing a rule to do the formatting, rather it says to use `bazel run` to build and run the `rustfmt` that's part of the toolchain that bazel has installed. The `no_std` issue you're seeing is I think because that `bazel run` command tries to build `rustfmt` using the target arch (`riscv64gc`) rather than the host arch. If that can be fixed, then it should work just fine.

sboeuf:
>I think because that bazel run command tries to build rustfmt using the target arch (riscv64gc) rather than the host arch.

Yes that's exactly the problem I had trying to run the rust-analyzer generator tool. The problem comes from `build --platforms="@rules_rivos//platforms:riscv64-none"` in `.bazelrc`.

stillson:
This should be fixed, and much simple.

abrestic-rivos (approved):

PR #271, opened 2023-03-06
Improve first time bazel experience

rbradford:
- README: Add instruction to update submodules
- Makefile: Make build targets depend on sbirs target

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,35 +5,35 @@
 
 # Do not add dependencies to the bazel targets here.
 # The dependencies are handled by bazel.
-all: sbirs
+all: update_submodules
 	bazel build //:salus-all
 
-check: sbirs
+check: update_submodules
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -5,32 +5,32 @@
...
 
-tellus:
+tellus: sbirs
 	bazel build //test-workloads:tellus
 
-umode:
+umode: sbirs
 	bazel build //umode:
 
 sbirs:
```
abrestic-rivos, line 36: this target should really be called `update_submodules` or `submodules`, but i suppose this is just matching what we previously had.

abrestic-rivos (approved):
thanks!

Makefile
rbradford: I think so too, PR updated.

sboeuf (approved):

PR #272, opened 2023-03-06
Implement a simple stack dump on panic.

stillson:
This does no symbol resolution, just walks up the stack frame and prints the return addresses, to a maximum of 10 calls.

### DIFF
--- a/src/backtrace.rs
+++ b/src/backtrace.rs
@@ -1,0 +1,126 @@
+// SPDX-FileCopyrightText: 2023 Rivos Inc.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+/// Get an array of backtrace addresses.
+///
+/// This needs `force-frame-pointers` enabled for rustc
+use crate::hyp_layout::{HYP_STACK_BOTTOM, HYP_STACK_TOP};
+use alloc::fmt::{Display, Formatter, Result};
+use core::arch::asm;
+use core::mem::size_of;
+
+#[derive(Copy, Clone)]
+pub enum BTReturnAddress {
+    ReturnAddress(u64),
+    InvalidFramePointer(u64),
+    InvalidReturnAddress(u64),
+}
+
+impl Display for BTReturnAddress {
+    fn fmt(&self, f: &mut Formatter<'_>) -> Result {
+        match self {
+            BTReturnAddress::ReturnAddress(addr) => {
+                writeln!(f, "0x{addr:x}")
+            }
+            BTReturnAddress::InvalidFramePointer(addr) => {
+                writeln!(f, "Invalid frame pointer: 0x{addr:x}")
+            }
+            BTReturnAddress::InvalidReturnAddress(addr) => {
+                writeln!(f, "Null address at frame pointer: 0x{addr:x}")
+            }
+        }
+    }
+}
+
+pub struct BackTrace {
+    fp: Option<u64>,
+    stack_start: u64,
+    stack_end: u64,
+    stack: &'static [u64],
+}
+
+impl BackTrace {
+    fn new(fp: u64) -> Self {
+        let stack_start = HYP_STACK_BOTTOM;
+        let stack_end = HYP_STACK_TOP;
+        if fp < stack_start || fp > stack_end {
+            return Self {
+                fp: None,
+                stack_start: 0_u64,
+                stack_end: 0_u64,
+                stack: &[],
+            };
+        }
+
+        // Safe because we only access memory that should be paged in
+        // and we never write to it
+        let stack = unsafe {
+            core::slice::from_raw_parts(
+                stack_start as *const u64,
+                (stack_end - stack_start) as usize,
+            )
+        };
+
+        Self {
+            fp: Some(fp),
+            stack_start,
+            stack_end,
+            stack,
+        }
+    }
+
+    fn next_address(&self, fp: u64) -> Option<(u64, u64)> {
+        // If offset is out of bounds get will return none, and the whole function
+        // will return none, except for the small gap from stack_start to stack_start + 1
+
+        // Guarantee alignment and check that gap.
+        if !(fp as *const u64).is_aligned() {
+            return None;
+        }
+        if fp <= self.stack_start + size_of::<u64>() as u64 {
+            return None;
+        }
+
+        let offset = ((fp - self.stack_start) / size_of::<u64>() as u64) as usize;
+        let address = *self.stack.get(offset - 1)?;
+        let current_fp = *self.stack.get(offset - 2)?;
+        Some((address, current_fp))
+    }
+}
+
+impl Iterator for BackTrace {
+    type Item = BTReturnAddress;
+    fn next(&mut self) -> Option<Self::Item> {
+        if let Some(mut fp) = self.fp {
+            if fp < self.stack_start || fp > self.stack_end {
+                self.fp = None;
+                return Some(BTReturnAddress::InvalidFramePointer(fp));
+            }
+            if let Some((address, current_fp)) = self.next_address(fp) {
+                fp = current_fp;
+                self.fp = Some(fp);
+                if address == 0 {
+                    self.fp = None;
+                    return Some(BTReturnAddress::InvalidReturnAddress(fp));
+                }
+                return Some(BTReturnAddress::ReturnAddress(address));
+            }
+            self.fp = None;
+            Some(BTReturnAddress::InvalidReturnAddress(fp))
+        } else {
+            None
+        }
+    }
+}
+
+pub(crate) fn backtrace() -> BackTrace {
+    // Safe because we are just reading a register
+    let fp = unsafe {
+        let mut tmp: u64;
+        asm!("mv {0}, fp", out(reg) tmp);
+        tmp
+    };
+
+    BackTrace::new(fp)
+}
--- a/src/main.rs
+++ b/src/main.rs
@@ -23,6 +23,7 @@
 #![test_runner(crate::tests::test_runner)]
 #![reexport_test_harness_main = "test_main"]
 #![cfg_attr(test, allow(unused))]
+#![feature(pointer_is_aligned)]
 
 use core::alloc::{Allocator, GlobalAlloc, Layout};
 use core::fmt::Display;
@@ -31,6 +32,7 @@ use test_system::*;
 extern crate alloc;
 
 mod asm;
+mod backtrace;
 mod guest_tracking;
 mod host_vm;
 mod hyp_layout;
@@ -45,6 +47,7 @@ mod vm_interrupts;
 mod vm_pages;
 mod vm_pmu;
 
+use backtrace::backtrace;
 use device_tree::{DeviceTree, DeviceTreeError, Fdt};
 use drivers::{
     imsic::Imsic, iommu::Iommu, pci::PcieRoot, pmu::PmuInfo, reset::ResetDriver, uart::UartDriver,
@@ -72,6 +75,22 @@ use umode::UmodeTask;
 #[panic_handler]
 fn panic(info: &core::panic::PanicInfo) -> ! {
     println!("panic : {:?}", info);
+    if let Some(location) = info.location() {
+        println!(
+            "panic occurred in file '{}' at line {}",
+            location.file(),
+            location.line()
+        );
+    }
+
+    println!("panic backtrace:");
+    // Skip the first 2 calls on the stack.
+    // They are always the same, and are from the
+    // panic handling code.
+    backtrace().skip(2).for_each(|frame| {
+        print!("{}", frame);
+    });
+
     abort()
 }
 
--- a/.bazelrc
+++ b/.bazelrc
@@ -4,6 +4,8 @@
 
 # By default, build optimized
 build --compilation_mode=opt
+# add frame pointers to make tracing the stack of a crash easier
+build --@rules_rust//:extra_rustc_flag=-Cforce-frame-pointers
 
 # Use toolchain resolution to find the cc toolchain.
 # This is required to ensure cross compilation operates with the correct toolchain.
@@ -21,4 +23,3 @@ build --@rules_rust//:clippy_flags=-Dmissing-docs,-Dwarnings
 # Use the workspace status feature to include the git commit in the build:
 # https://bazel.build/docs/user-manual#workspace-status
 build --workspace_status_command=bazel/workspace_status_command.sh
-
--- a/scripts/resolve_stack_addr.sh
+++ b/scripts/resolve_stack_addr.sh
@@ -1,0 +1,38 @@
+#!/bin/bash
+# SPDX-FileCopyrightText: 2023 Rivos Inc.
+#
+# SPDX-License-Identifier: Apache-2.0
+
+# This script is used to resolve addresses from a stack dump to
+# symbolic names. It can be used two ways:
+#
+# 1. With a single argument it will resolve the address represented by the argument
+#
+# > ./scripts/resolve_stack_addr.sh 0x00000000802d7a2a
+#   const_oid::arcs::Arcs::try_next at external/rice-index__const-oid-0.9.1/src/arcs.rs:51
+#
+# 2. with no arguments it will read from stdin, which allows you to cut and paste a full stack
+#
+# > ./scripts/resolve_stack_addr.sh
+#   0x000000008034757a
+#   0x00000000802fc6be
+#   0x00000000802f3b72
+# ^D <-- type a literal control-D when done to specify EOF
+#   <core::array::iter::IntoIter<T,_> at /rustc/1db9c061df216e2da/library/core/src/array/iter.rs:249
+#   <core::array::iter::IntoIter<T,_> as core::next::{{closure}} at /rustc/116e2da/library/core/src/array/iter.rs:249
+#   hyp_alloc::arena::Arena<T,A>::ids::{{closure}} at hyp-alloc/src/arena.rs:118
+
+. scripts/common_variables
+
+if [ x$1 == x ]; then
+
+    stack=$(</dev/stdin)
+
+    for s in $stack
+    do
+        addr2line -fiCpe ${SALUS_BINS}salus $s
+    done
+
+else
+    addr2line -fiCpe ${SALUS_BINS}salus $1
+fi


src/backtrace.rs
```diff
@@ -0,0 +1,78 @@
...
+
+extern "C" {
+    static _stack_start: u8;
+    static _stack_end: u8;
+}
+
+pub(crate) fn backtrace() -> [Option<usize>; MAX_BACKTRACE_ADDRESSES] {
+    // Safe because we are just reading a register
+    let fp = unsafe {
+        let mut _tmp: u64;
```
atulkharerivos, line 20: Nit: _ should be used only for unused variables (unless it's being used to pacify the compiler in this context).

src/backtrace.rs
```diff
@@ -0,0 +1,78 @@
...
+    let mut skip = 2;
+
+    // Safe because we trust the linker placed these symbols correctly.
+    let stack_start = unsafe { core::ptr::addr_of!(_stack_start) as u64 };
+    let stack_end = unsafe { core::ptr::addr_of!(_stack_end) as u64 };
+
+    let mut fp = fp;
+    let mut old_address = 0;
+    loop {
+        // safe because we are only reading in the valid stack trace area
```
atulkharerivos, line 40: Nit: s/safe/Safe/

atulkharerivos (commented):
Nice...is this safe to use even if `-Cforce-frame-pointers` is accidentally omitted?

src/backtrace.rs
stillson: fixed

src/backtrace.rs
stillson: fixed

stillson:
I believe it's ok if you don't force frame-pointers. The only dangerous thing it does is pull memory off the stack, and it check if it is in stack bounds first.

src/backtrace.rs
```diff
@@ -0,0 +1,67 @@
...
+    // Safe because we trust the linker placed these symbols correctly.
+    let stack_start = unsafe { core::ptr::addr_of!(_stack_start) as u64 };
+    let stack_end = unsafe { core::ptr::addr_of!(_stack_end) as u64 };
+
+    let mut fp = fp;
+    loop {
+        // Safe because we are only reading in the valid stack trace area
+        // Also, this is only called on panic
+
+        if fp < stack_start || fp > stack_end {
```
rbradford, line 42: Would it be interesting to know when the fp points outside the expected range (stack corruption?) could there be a `println!()` of a message here (or maybe return a `Result<...>` and handle it elsewhere.)

src/backtrace.rs
```diff
@@ -0,0 +1,67 @@
...
+        tmp
+    };
+
+    backtrace_internal(fp)
+}
+
+fn backtrace_internal(fp: u64) -> [Option<usize>; MAX_BACKTRACE_ADDRESSES] {
+    let mut result = [None; MAX_BACKTRACE_ADDRESSES];
+    let mut index = 0;
+    let mut skip = 2;
```
rbradford, line 31: Naive question: is this here to skip including this code itself?

src/backtrace.rs
```diff
@@ -0,0 +1,67 @@
...
+
+        if fp < stack_start || fp > stack_end {
+            break;
+        }
+
+        let address = unsafe { (fp as *const u64).offset(-1).read_volatile() }; // RA/PC
+        fp = unsafe { (fp as *const u64).offset(-2).read_volatile() }; // next FP
+
+        if address == 0 {
+            break;
```
rbradford, line 50: Again, is this something that we could usefully report an error out on?

attestation/BUILD
```diff
@@ -27,6 +27,9 @@ rust_library(
         "@salus-index//:hex",
         "@salus-index//:typenum",
     ],
+    rustc_flags = [
+        "-Cforce-frame-pointers",
+    ],
```
abrestic-rivos, line 32: instead of putting this in every single `BUILD` file, could you put it in the `.bazelrc` or similar?

abrestic-rivos (commented):
i'm not convinced we should be rolling our own backtrace here... have you tried getting something like `mini-backtrace` to work?

rbradford:
> mini-backtrace

I also found https://docs.rs/unwinding/latest/unwinding/ which doesn't pull in the libunwind (i.e. it's pure Rust.)

src/backtrace.rs
stillson: That, and the panic code that is always the same.

attestation/BUILD
stillson: Well examine

src/backtrace.rs
stillson: this might be best handled with a enum for all the error types

src/backtrace.rs
stillson: again, enum

src/backtrace.rs
```diff
@@ -0,0 +1,77 @@
...
+    let fp = unsafe {
+        let mut tmp: u64;
+        asm!("mv {0}, x8", out(reg) tmp);
+        tmp
+    };
+
+    backtrace_internal(fp)
+}
+
+fn backtrace_internal(fp: u64) -> [BTReturn; MAX_BACKTRACE_ADDRESSES] {
```
dgreid, line 36: 'fp' is dereferenced as a raw pointer, so this function shouldn't be safe.

stillson: fixed.

src/backtrace.rs
stillson: fixed

src/backtrace.rs
stillson: fixed

src/backtrace.rs
```diff
@@ -0,0 +1,77 @@
...
+
+pub(crate) fn backtrace() -> [BTReturn; MAX_BACKTRACE_ADDRESSES] {
+    // Safe because we are just reading a register
+    let fp = unsafe {
+        let mut tmp: u64;
+        asm!("mv {0}, x8", out(reg) tmp);
+        tmp
+    };
+
+    unsafe { backtrace_internal(fp) }
```
atulkharerivos, line 33: Needs a comment explaining why it's safe, and also corresponding documentation on the caller requirements in `backtrace_internal()`. 

Probably something like:

```
// # Safety
// The caller must ensure that fp is a valid address on the stack frame.
unsafe fn backtrace_internal(fp: u64) -> [BTReturn; MAX_BACKTRACE_ADDRESSES] {
```

src/backtrace.rs
```diff
@@ -0,0 +1,77 @@
...
+    let mut index = 0;
+    let mut skip = 2;
+
+    // Safe because we trust the linker placed these symbols correctly.
+    let stack_start = unsafe { core::ptr::addr_of!(_stack_start) as u64 };
+    let stack_end = unsafe { core::ptr::addr_of!(_stack_end) as u64 };
+
+    let mut fp = fp;
+    loop {
+        // Safe because we are only reading in the valid stack trace area
```
atulkharerivos, line 47: Nit: Comment should be moved a few lines below.

src/backtrace.rs
```diff
@@ -0,0 +1,77 @@
...
+    static _stack_start: u8;
+    static _stack_end: u8;
+}
+
+#[derive(Copy, Clone)]
+pub(crate) enum BTReturn {
+    Some(usize),
+    None,
+    BadFp(usize),
+    BadAddress(usize),
```
atulkharerivos, line 22: Nit: Since both u64 and usize are are 64-bit, using u64 will avoid the requirement for casting.

attestation/BUILD
stillson: Fixed. added to .bazelrc

src/backtrace.rs
stillson: fixed

src/backtrace.rs
stillson: fixed

src/backtrace.rs
stillson: fixed

src/backtrace.rs
```diff
@@ -0,0 +1,77 @@
...
+pub const MAX_BACKTRACE_ADDRESSES: usize = 10;
+
+extern "C" {
+    static _stack_start: u8;
+    static _stack_end: u8;
+}
+
+#[derive(Copy, Clone)]
+pub(crate) enum BTReturn {
+    Some(u64),
```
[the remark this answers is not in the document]
stillson: Makes sense. fixed

src/main.rs
```diff
@@ -77,6 +79,31 @@ use umode::UmodeTask;
...
+        let bt = backtrace();
+        println!(
+            "panic occurred in file '{}' at line {}",
+            location.file(),
+            location.line()
+        );
+        println!("panic backtrace:");
+        for frame in bt {
+            match frame {
+                BTReturn::Some(x) => {
```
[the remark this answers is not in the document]
stillson: yes, good idea. fixed.

src/backtrace.rs
```diff
@@ -0,0 +1,77 @@
...
+
+pub const MAX_BACKTRACE_ADDRESSES: usize = 10;
+
+extern "C" {
+    static _stack_start: u8;
+    static _stack_end: u8;
+}
+
+#[derive(Copy, Clone)]
+pub(crate) enum BTReturn {
```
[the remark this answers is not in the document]
stillson: Fixed

src/backtrace.rs
```diff
@@ -0,0 +1,95 @@
+// SPDX-FileCopyrightText: 2023 Rivos Inc.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+use alloc::fmt::{Display, Formatter, Result};
+/// Get an array of backtrace addresses.
```
rbradford, line 6: nit: Not sure if the project enforces a particular style for use statements (grouping by external/internal) but I doesn't look right to have a comment in the middle here.

stillson: fixed

src/backtrace.rs
```diff
@@ -0,0 +1,95 @@
...
+    let mut fp = fp;
+    loop {
+        if fp < stack_start || fp > stack_end {
+            result[index] = BTReturnAddress::InvalidFramePointer(fp);
+            break;
+        }
+
+        // Safe because we are only reading in the valid stack trace area
+        let address = unsafe { (fp as *const u64).offset(-1).read_volatile() }; // RA/PC
+        fp = unsafe { (fp as *const u64).offset(-2).read_volatile() }; // next FP
```
atulkharerivos, line 74: Sorry if this a duplicate, but my Github comment rendering has been really weird.

My test could be flawed, but what if fp == stack_start when we begin execution? Basically, we are effectively addressing *(fp - 16) on line 74, meaning that we can underflow the bounds of the stack?

```
fn assert_stack() {
     // Safe because we trust the linker placed these symbols correctly.
     let stack_start = unsafe { core::ptr::addr_of!(_stack_start) as u64 };
     let stack_end = unsafe { core::ptr::addr_of!(_stack_end) as u64 };
     let fp = stack_start;
     //  Same as if fp < stack_start || fp > stack_end on 67
     if fp >= stack_start && fp <= stack_end {
        let address = unsafe { (fp as *const u64).offset(-1) as u64 };
        let address = unsafe { (fp as *const u64).offset(-2) as u64 };
        assert!(address < stack_end);
        assert!(address > stack_start);
     }
}
```

atulkharerivos: On a somewhat related note, can we treat the stack as an as `&u64` slice that begins with stack_start, and with `len == (stack_end - stack_start) / 8`? That way, we can do the unsafe{} cast once, and will get bounds check for free, instead of relying on multiple unsafe checks.

atulkharerivos: Actually, there's no math involved, except for fp - stack_start. You can probably even mark the method as safe (but that's a separate discussion). 

Also, you simplify the code by replacing the loop with a `while index < MAX_BACKTRACE_ADDRESSES`.

For example, the following appears to be the equivalent of your code:

```
// Safety
// The caller must ensure that fp is a valid address on the stack frame.
unsafe fn backtrace_internal(fp: u64) -> [BTReturnAddress; MAX_BACKTRACE_ADDRESSES] {
    use s_mode_utils::print::CONSOLE;
    use s_mode_utils::print::*;
    let mut result = [BTReturnAddress::End; MAX_BACKTRACE_ADDRESSES];
    let mut index = 0;
    let mut skip = 2;

    // Safe because we trust the linker placed these symbols correctly.
    let stack_start = unsafe { core::ptr::addr_of!(_stack_start) as u64 };
    let stack_end = unsafe { core::ptr::addr_of!(_stack_end) as u64 };
    let mut fp  = fp;
    let stack = unsafe {core::slice::from_raw_parts(stack_start as *const u64, (stack_end - stack_start) as usize)};
    loop {  
        if fp < (stack_start+2) || fp > stack_end {
            result[index] = BTReturnAddress::InvalidFramePointer(fp);
            break;
        }

        let offset = ((fp - stack_start - 2)/8) as usize;
        if offset > 0 && let Some(address) = stack.get(offset) && let Some(current_fp) = stack.get(offset-1) {
            fp = *current_fp;
            let address = *address;     
            if address == 0 {
                result[index] = BTReturnAddress::InvalidReturnAddress(fp);
                break;
            }
            
            if skip > 0 {
                skip -= 1;
                continue;
            }
    
            result[index] = BTReturnAddress::ReturnAddress(address);
            index += 1;
    
            if index >= MAX_BACKTRACE_ADDRESSES {
                break;
            }        
        } else {
            result[index] = BTReturnAddress::InvalidReturnAddress(fp);
            break;
        }
    }

    result
}

```

rbradford:
I think this needs revisiting with per-vCPU VA mapped stack work that is in #290

src/main.rs
```diff
@@ -78,6 +80,21 @@ use umode::UmodeTask;
...
 fn panic(info: &core::panic::PanicInfo) -> ! {
+    if let Some(location) = info.location() {
+        println!(
+            "panic occurred in file '{}' at line {}",
+            location.file(),
+            location.line()
+        );
+    } else {
+        println!("panic occurred but can't get location information...");
+    }
```
glg-rv, line 91: This information is already printed on the "info" down later.

glg-rv:
> I think this needs revisiting with per-vCPU VA mapped stack work that is in #290

Yes AFAICT though the main thing to change is where to get the stack end/start. the rest should be hopefully fine.

src/backtrace.rs
```diff
@@ -0,0 +1,95 @@
...
+
+// Safety
+// The caller must ensure that fp is a valid address on the stack frame.
+unsafe fn backtrace_internal(fp: u64) -> [BTReturnAddress; MAX_BACKTRACE_ADDRESSES] {
+    let mut result = [BTReturnAddress::End; MAX_BACKTRACE_ADDRESSES];
+    let mut index = 0;
+    let mut skip = 2;
+
+    // Safe because we trust the linker placed these symbols correctly.
+    let stack_start = unsafe { core::ptr::addr_of!(_stack_start) as u64 };
```
atulkharerivos, line 62: I still think that using a &[u64] slice is much better approach than trying to shift bits and check for boundary conditions. Using a slice only requires _stack_end and _stack_start to be right (and it's a given implicit assumption in both cases). However, when using slices, reviewers (and in the future maintainers) don't have to worry about getting anything else right  in the unsafe blocks (since there are none), and all checks are enforced by the compiler.

rbradford: And with the slice approach you can even iterate over it rather than the current `loop {}` with the `offset` calculation which is more of an index

atulkharerivos: Furthermore, it will make it possible to hoist the skip logic by simply returning the slice, and then using slice.iter().skip(2) at the caller site, which simplifies the code even further.

src/backtrace.rs
stillson: fixed

src/main.rs
stillson: fixed

src/backtrace.rs
atulkharerivos: No, we don't need to create a stack iterator because you get it for free after creating a slice. For example, I think you can simply return `[BTReturnAddress; MAX_BACKTRACE_ADDRESSES]` and then use iter() in the return value.

For example:

`
let stack_trace = unsafe {backtrace_internal(fp)};
stack_trace.to_iter().skip(2).for_each(|entry_type|) {
   // Handle entry type.
}
`

Also, notice how we use unsafe{} exactly once below with the linker defined symbols, and then rely on the compiler to enforce the rest without having to worry about whether we got it right.

/```
/ Safety
// The caller must ensure that fp is a valid address on the stack frame.
unsafe fn backtrace_internal(fp: u64) -> [BTReturnAddress; MAX_BACKTRACE_ADDRESSES] {
    let mut result = [BTReturnAddress::End; MAX_BACKTRACE_ADDRESSES];
    let mut index = 0;
    let mut skip = 2;

    // Safe because we trust the linker placed these symbols correctly.
    let stack_start = unsafe { core::ptr::addr_of!(_stack_start) as u64 };
    let stack_end = unsafe { core::ptr::addr_of!(_stack_end) as u64 };
    let mut fp  = fp;
    let stack = unsafe {core::slice::from_raw_parts(stack_start as *const u64, (stack_end - stack_start) as usize)};
    loop {  
        if fp < (stack_start+2) || fp > stack_end {
            break;
        }

        let offset = ((fp - stack_start - 2)/8) as usize;
        if offset > 0 && let Some(address) = stack.get(offset) && let Some(current_fp) = stack.get(offset-1) {
            fp = *current_fp;
            let address = *address;     
            if address == 0 {
                break;
            }
            
            result[index] = BTReturnAddress::ReturnAddress(address);
            index += 1;
    
            if index >= MAX_BACKTRACE_ADDRESSES {
                break;
            }        
        } else {
            break;
        }
    }

    result
}
```

File: src/backtrace.rs [hunk 18]
```diff
@@ -0,0 +1,104 @@
...
+            stack_start,
+            stack_end,
+            stack,
+        }
+    }
+}
+
+impl Iterator for BackTrace {
+    type Item = BTReturnAddress;
+    fn next(&mut self) -> Option<Self::Item> {
```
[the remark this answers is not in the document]
stillson: fixed

atulkharerivos (approved):

src/backtrace.rs
```diff
@@ -0,0 +1,105 @@
...
+        }
+    }
+}
+
+//pub(crate) fn backtrace() -> [BTReturnAddress; MAX_BACKTRACE_ADDRESSES] {
+pub(crate) fn backtrace() -> BackTrace {
+    // Safe because we are just reading a register
+    let fp = unsafe {
+        let mut tmp: u64;
+        asm!("mv {0}, x8", out(reg) tmp);
```
abrestic-rivos, line 100: s/x8/fp

src/backtrace.rs
```diff
@@ -0,0 +1,105 @@
...
+            }
+            self.fp = None;
+            Some(BTReturnAddress::InvalidReturnAddress(fp))
+        } else {
+            None
+        }
+    }
+}
+
+//pub(crate) fn backtrace() -> [BTReturnAddress; MAX_BACKTRACE_ADDRESSES] {
```
abrestic-rivos, line 95: stale comment?

src/backtrace.rs
```diff
@@ -0,0 +1,105 @@
...
+            stack,
+        }
+    }
+}
+
+impl Iterator for BackTrace {
+    type Item = BTReturnAddress;
+    fn next(&mut self) -> Option<Self::Item> {
+        if let Some(mut fp) = self.fp {
+            if fp < self.stack_start + 2 || fp > self.stack_end {
```
abrestic-rivos, line 71: why the `+2`?

src/backtrace.rs
```diff
@@ -0,0 +1,105 @@
...
+                self.fp = None;
+                return Some(BTReturnAddress::InvalidFramePointer(fp));
+            }
+
+            let offset = ((fp - self.stack_start - 2) / 8) as usize;
```
abrestic-rivos, line 76: replace the 8 with `size_of::<u64>()`

src/backtrace.rs
```diff
@@ -0,0 +1,105 @@
...
+            if offset > 0 && let Some(address) = self.stack.get(offset) && let Some(current_fp) = self.stack.get(offset-1) {
```
abrestic-rivos, line 77: should we enforce that the stack only go up? i.e. next fp must be > previous fp

rbradford, line 77: Perhaps this could be a method

```rust
fn next_address(&self, fp: u64) -> Option<(u64, u64> {
  let offset = ((fp - self.stack_start - 2) / 8) as usize;
  if offset == 0 {
     return None;
  }
  let address = self.stack.get(offset)?;
  let current_fp = self.stack.get(offset - 1)?;

  Some(address, current_fp)
```

That means you can then just do `if let Some(address, current_fp) = self.next_address()`

src/backtrace.rs
stillson: yes, fixed

src/backtrace.rs
stillson: fixed

src/backtrace.rs
stillson: Not needed anymore. gone.

src/backtrace.rs
stillson: fixed

src/backtrace.rs
stillson: I can see some cases of asynchronous code where that might not be true. You might catch some strange bugs. Worth doing for now.

stillson: fixed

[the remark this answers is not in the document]
stillson: fixed

stillson: Actually, kernel_init uses an odd spot in the stack. leaving out for now, since this seems to cause an issue.

stillson: The problem with this is that you lose information (if offset or offset -1 are not in the stack).

rbradford: I'm not sure I follow? If either of `offset` or `offset - 1` some part of the if statement will be None and not executed. This function does the same thing by prematurely returning `None` via ?

src/backtrace.rs
```diff
@@ -0,0 +1,107 @@
...
+impl Iterator for BackTrace {
+    type Item = BTReturnAddress;
+    fn next(&mut self) -> Option<Self::Item> {
+        if let Some(mut fp) = self.fp {
+            if fp < self.stack_start || fp > self.stack_end {
+                self.fp = None;
+                return Some(BTReturnAddress::InvalidFramePointer(fp));
+            }
+
+            let offset = ((fp - self.stack_start - 2) / (size_of::<u64>() as u64)) as usize;
```
abrestic-rivos, line 76: why the `- 2`?

src/backtrace.rs
```diff
@@ -0,0 +1,107 @@
...
+            if offset > 0 &&
```
abrestic-rivos, line 77: this is going to be true unless `fp == self.stack_start`. i think if you actually want to catch an `fp` below `stack_start` you need to do a `checked_sub()` above

src/main.rs
```diff
@@ -79,6 +81,19 @@ use umode::UmodeTask;
...
+    if let Some(location) = info.location() {
+        println!(
+            "panic occurred in file '{}' at line {}",
+            location.file(),
+            location.line()
+        );
+    }
+
+    println!("panic backtrace:");
+    backtrace().skip(2).for_each(|frame| {
```
abrestic-rivos, line 93: put a comment about the `skip(2)`

src/backtrace.rs
atulkharerivos: Think that as originally suggested, the condition protected against the subtract underflow in ` (fp - stack_start - 2)`. However, with this change, we do need to use `checked_sub()`. One way to do it might be to use the following instead, which will set offset to 0 if we underflow:

` let offset = ((fp.checked_sub(self.stack_start).unwrap_or(2) - 2) / (size_of::<u64>() as u64)) as usize;`

```
    if fp < (stack_start+2) || fp > stack_end {
        break;
    }
```

src/backtrace.rs
atulkharerivos: Yeah, seems like we will encounter this because we removed the check for ` fp - 2` that guarded against the underflow. Using something like the following should set offset to 0 without a panic, but we may want to check for the underflow and simply bail.

` let offset = ((fp.checked_sub(self.stack_start).unwrap_or(2) - 2) / (size_of::<u64>() as u64)) as usize;`

src/backtrace.rs
stillson: My bad. You are right, although I had to add a few *'s

src/main.rs
stillson: fixed

src/backtrace.rs
```diff
@@ -0,0 +1,113 @@
...
+        Self {
+            fp: Some(fp),
+            stack_start,
+            stack_end,
+            stack,
+        }
+    }
+
+    fn next_address(&self, fp: u64) -> Option<(u64, u64)> {
+        let offset = (fp - self.stack_start - 2) as usize / size_of::<u64>();
```
atulkharerivos, line 67: Needs to guard against potential underflows:

`let offset = fp.checked_sub(self.stack_start).unwrap_or(2) .checked_sub(2).unwrap_or(0) as usize / sizeof::u64();`

src/backtrace.rs
```diff
@@ -0,0 +1,113 @@
...
+        if offset == 0 {
```
atulkharerivos, line 68: Nit: Can elide return with: `if offset > 0 {}...else { None}`

src/backtrace.rs
stillson: It uses slice.get(), which returns None on out of bounds.

atulkharerivos: Think we are referring to different portions of code unless the code needs to be refreshed. In the below, it seems offset is computed unconditionally, which can cause an underflow if `fp == self.stack_start` for example. The first checked_sub() guards against an underflow with `(fp - self.stack_start)`, and the second guards against the case where `'fp - self.stack_start'` evaluates to 0 (which will cause a underflow with the -2).

In either of those cases, offset will be 0.

   ```suggestion
         fn next_address(&self, fp: u64) -> Option<(u64, u64)> {
             let offset = (fp.checked_sub(self.stack_start).unwrap_or(2).checked_sub(2).unwrap_or(0) as usize / size_of::<u64>();
```

File: src/backtrace.rs [hunk 29]
```diff
@@ -0,0 +1,110 @@
...
+            stack_end,
+            stack,
+        }
+    }
+
+    fn next_address(&self, fp: u64) -> Option<(u64, u64)> {
+        // if offset is out of bounds get will return none, and the whole function
+        // will return none
+        let offset = ((fp - self.stack_start) / size_of::<u64>() as u64) as usize;
+        let address = *self.stack.get(offset - 1)?;
```
atulkharerivos, line 70: The issue here is the unchecked subtract.

To illustrate the problem, here's a sample program that demonstrates what happens if offset is 0 for example (it's silly to prevent the compiler from detecting the attempted subtract from 0 at compile time):

```
fn get_offset(b : bool) -> usize {
    if b {
        0
     } else { 4 }
}

fn main() {
    let stack = &[0, 1, 2, 3];
    let address = stack.get(foo(true) - 2);
    println!("{:?}", address);
}
```

src/backtrace.rs
```diff
@@ -0,0 +1,106 @@
...
+    fp: Option<u64>,
+    stack_start: u64,
+    stack_end: u64,
+    stack: &'static [u64],
+}
+
+impl BackTrace {
+    fn new(fp: u64) -> Self {
+        let stack_start = HYP_STACK_BOTTOM;
+        let stack_end = HYP_STACK_TOP;
```
glg-rv, line 46: Now that the stack address is fixed, the stack_start and stack_end could _possibly_ be removed. The slice would be fixed, everything would be more static.

src/backtrace.rs
```diff
@@ -0,0 +1,106 @@
...
+        let stack = unsafe {
+            core::slice::from_raw_parts(
+                stack_start as *const u64,
+                (stack_end - stack_start) as usize,
+            )
+        };
+
+        Self {
+            fp: Some(fp),
```
glg-rv, line 55: I think the best thing to do is to check whether fp is between HYP_STACK_TOP and HYP_STACK_BOTTOM. If not, disable the backtrace.

This will both fix the problem that is unaddressed right now, the fact that before we switch to the VA stack (or when we are on the overflow stack) this would fail, and it would also make the calculation of the offset below safe.

glg-rv (changes requested):
As discussed, this still doesn't handle the problem that stack during the early boot (when we could panic).

src/backtrace.rs
stillson: Since this is really just 2 addresses (stack_start and stack_start + 8) I guaranteed alignment and check for the addresses. I think that is enough of a check to fix the problem.

src/backtrace.rs
stillson: we need stack start for bounds checking and calculating the offset, but we don't need stack end.

stillson: actually, we use stack_end too. we don't really need it, but I'd rather fail early.

src/backtrace.rs
stillson: I put a check in, and if it's out of bounds I just return a "NULL" object.

atulkharerivos (commented):
Looks good, with the new check for  `if fp <= self.stack_start + size_of::<u64>() as u64 `. BTW, since the stack start/end are constants, can we add a test to check for a boundary conditions? Basically, something like constructing BackTrace with ` fp == HYP_STACK_BOTTOM` and perhaps, 'fp > HYP_STACK_TOP', and verifying that  `.next()` returns `None`.

atulkharerivos (approved):
Looks like we can't use the actual Salus stack constant values for BT testing, so a test is moot unless we can run it under QEMU.

stillson:
For what it's worth, I started making this testable in another branch, and it work fine.

src/backtrace.rs
```diff
@@ -0,0 +1,115 @@
+// SPDX-FileCopyrightText: 2023 Rivos Inc.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+use crate::hyp_layout::{HYP_STACK_BOTTOM, HYP_STACK_TOP};
```
abrestic-rivos, line 5: nit: this `use` should go below with the others.

src/main.rs
abrestic-rivos: not fixed?

abrestic-rivos (commented):
maybe i'm missing something, but it still doesn't look like this is addressing @glg-rv's point about panics that occur in early boot, before the stack is mapped at `HYP_STACK_BOTTOM`.

src/main.rs
stillson: My bad. I forgot to push out the latest version.

src/backtrace.rs
```diff
@@ -0,0 +1,115 @@
...
+    stack_start: u64,
+    stack_end: u64,
+    stack: &'static [u64],
+}
+
+impl BackTrace {
+    fn new(fp: u64) -> Self {
+        let stack_start = HYP_STACK_BOTTOM;
+        let stack_end = HYP_STACK_TOP;
+        let stack = unsafe {
```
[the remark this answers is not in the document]
stillson: Somehow, I had done this and lost it again. Fixed.

glg-rv (approved):
> maybe i'm missing something, but it still doesn't look like this is addressing @glg-rv's point about panics that occur in early boot, before the stack is mapped at `HYP_STACK_BOTTOM`.

This has since been fixed. Essentially now we check that we are in the HYP_STACK range and if so we attempt the stack backtrace. It is a bit flimsy but it's really hard to know from an exception handler whether we're still in the init phase. (Can't trust PerCpu, etc).

Something we could do in the future is to actually have two different rust exception handler, one for the init (early?) phase and the other from when we switched to hyp paging. We already have two different assembler entry points for the phase so that should be okay.

The reason I prefer not to have a stack trace in the init stage is because it's really hard to know what cpu we're running (could use SBI, but then we have to map it to CpuId and find what stack range we're using) and we really have no protection on the stack, so could also be not trusted.

Regarding this patch, A few things might be done simpler (no reason to keep stack_start and stack_end as structure of backtrace), and also should return none in case stack is not in range. But these are minor details and possibly be left there for trying to support the `init` phase (which I think we don't need).

PR #277, opened 2023-03-08
build: Enable Rust Analyzer support

rbradford:
The metadata can be generated with:

bazel run @rules_rust//tools/rust_analyzer:gen_rust_project

Users of VScode can automate this with:

https://bazelbuild.github.io/rules_rust/rust_analyzer.html#vscode

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -11,6 +11,7 @@ build --incompatible_enable_cc_toolchain_resolution
 
 build --platforms="@rules_rivos//platforms:riscv64-none"
 test --platforms=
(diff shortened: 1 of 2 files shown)


rbradford:
I chose not to include the .vscode/ file in the tree, but am happy to do that if folks would like to have it rather than having their own local version.

sboeuf (approved):
This is one step in the proper direction, but still missing the fact that rust-analyzer doesn't check for the validity of the code, and doesn't show unused variables for instance.

PR #278, opened 2023-03-08
Ignore rust-project.json file

sboeuf:
In order to use rust-analyzer with the project and given we moved away from Cargo, we need to generated a rust-project.json file so that rust-analyzer can run properly. This file should be ignored.


rbradford (approved):
Yes, the documentation recommends adding this to the .gitignore file.

abrestic-rivos (approved):

PR #280, opened 2023-03-09
Align the TEE[HGI]extension names with CoVE

atishp04:
This needs to be merged after
https://github.com/rivosinc/sbi-rs/pull/15

As mentioned in that PR,
This patch only updates the extension name & ID as per the latest naming changes in the spec.
We can modify all the reference of "tee" to cove as well to align it well or leave it as it is for now.


abrestic-rivos (approved):

rajnesh-kanwal:
Just to confirm, we are not merging this because sbi-rs is not pointing to the correct commit. right?

atishp04:
> > Just to confirm, we are not merging this because sbi-rs is not pointing to the correct commit. right?
> 
> right

It is updated now.

Issue #180, opened 2022-12-10
Map salus in the top half of the VA space.


glg-rv:
Currently Salus supervisor VA is mapped in the lower half of the virtual address. With the umode changes, we now map the umode in the top half.

We want to change this to a more _canonical_ way: supervisor in the top half and user in the lower half.

See: https://github.com/rivosinc/salus/pull/178#discussion_r1044744264

PR #284, opened 2023-03-09
HypMap clean-up

glg-rv:
This PR serves both to clean up my conscience of U-mode PRs leftovers and to prepare HypMap to support virtual mapped stack more easily.

Sharing before the stack patch to gather opinions.

patch1: U-mode Shared Region is renamed to U-mode Input Region, as it is not writable from U-mode.
patch2: All layout const definitions are moved to a new module: hyp_layout.
patch3: Private/Shared regions (per-cpu, global mappings) is untenable as it would complicate the types too much. Rename them for what they are: hwmap regions and umode_elf regions.


src/hyp_map.rs
```diff
@@ -317,6 +255,21 @@ impl PrivateRegion {
...
 }
 
+/// Generic Id names for each of the U-mode mapping slots.
+/// There is no mandated use for each of the slots, and caller can decide to map each of them
+/// readable or writable based on the requirement.
+#[derive(Copy, Clone)]
+pub enum UmodeSlotId {
+    A,
+    B,
+}
```
abrestic-rivos, line 264: should this go in `hyp_layout.rs`?

src/hyp_map.rs
```diff
@@ -19,21 +19,21 @@ use riscv_pages::{
...
 }
 
-// Maximum number of regions unique to every pagetable (private).
-const MAX_PRIVATE_REGIONS: usize = 32;
-// Maximum number of regions shared across all pagetables.
-const MAX_SHARED_REGIONS: usize = 32;
+// Maximum number of U-mode ELF regions.
+const MAX_UMODE_ELF_REGIONS: usize = 32;
+// Maximum number of hardware map regions.
+const MAX_HW_MAP_REGIONS: usize = 32;
```
abrestic-rivos, line 31: these regions are that are identity mapped, right? might be more obvious if this were called the "linear map" or "identity map"

abrestic-rivos (approved):
just a couple nits; otherwise lgtm.

src/hyp_map.rs
glg-rv: Was future-proofing for #180, and in general other possible identity mapped might require a different structure, and this change was exactly to create a region type for each source.

src/hyp_map.rs
glg-rv: Could go. Thought about it for about 30 seconds. SlotIDs and Slot Addrs are a specific of the implementation. But yeah, I'll move it, would make things cleaner.

glg-rv: Moved.

src/hyp_map.rs
glg-rv: Haven't changed this. Let me know if you have strong opinions about this.

PR #285, opened 2023-03-13
Vm: Use EcallAction::Forward for share/unshare calls from guest.

rajnesh-kanwal:
Given we do post processing after the share/unshare calls from guest in `try_complete_pending_op` and use `set_ecall_result` to set the ECALL results returned from the host, we end up incrementing sepc second time. First time when we call `set_ecall_result` from `run_vcpu` under EcallAction::Break case.

This skips the next instruction(s) (fp load and stack pointer adjustment instructions in case of kvm) and leads to strange issues.


abrestic-rivos (approved):
good catch!

sboeuf:
Thanks for fixing this @rajnesh-kanwal

PR #286, opened 2023-03-13
riscv-page-tables: Update GuestStagePageTable to support all page sizes

sboeuf:
All the functions updating the page table entries must be able to deal
with pages of different sizes across an entire range of memory. That's
why the loops and iterators previously used had to be updated to support
incrementing each iteration with a dynamic page size.


riscv-page-tables/src/page_table.rs
```diff
@@ -1296,3 +1333,65 @@ impl<'a, T: PagingMode> Drop for GuestStageMapper<'a, T> {
...
+    Proceed(PageSize, SupervisorPageAddr),
+    Skip(PageSize),
+    End,
+}
+
+/// Generate a supervisor page iterator for the given range
+struct SupervisorPageIter<
+    T: PagingMode,
+    F: FnMut(PageAddr<T::MappedAddressSpace>) -> IteratorAction,
+> {
```
abrestic-rivos, line 1348: nit: given how verbose these types are, i'd use `where` to spell out the trait bounds

riscv-page-tables/src/page_table.rs
```diff
@@ -989,37 +1020,32 @@ impl<T: PagingMode> GuestStagePageTable<T> {
...
-                    }
-                }
-                _ => {
-                    return Err(Error::PageNotMapped);
-                }
+        let mut va = vaddr;
+        loop {
+            let pte = inner.get_mapped_leaf(va)?;
+            let page_size = pte.level().leaf_page_size();
+            if !pred(pte.page_addr()) {
```
abrestic-rivos, line 1031: not necessary for this PR, but it's likely that we will want to pass the page size to the predicate function as well.

riscv-page-tables/src/page_table.rs
```diff
@@ -977,6 +988,26 @@ impl<T: PagingMode> GuestStagePageTable<T> {
         Ok(mapper)
     }
 
+    fn check_and_increment_vaddr(
+        mut va: PageAddr<T::MappedAddressSpace>,
+        end: PageAddr<T::MappedAddressSpace>,
+        page_size: PageSize,
+    ) -> Result<Option<PageAddr<T::MappedAddressSpace>>> {
```
abrestic-rivos, line 995: nit: i find `Result<Option<>` to be an awkward return value. instead of returning `Ok(None)` in the `va == end` case, could you instead have the outer loop in which this called go until `va == end`?

riscv-page-tables/src/page_table.rs
```diff
@@ -1296,3 +1333,65 @@ impl<'a, T: PagingMode> Drop for GuestStageMapper<'a, T> {
...
+        loop {
+            if self.current >= self.end {
+                return None;
+            }
+
+            match (self.update)(self.current) {
+                IteratorAction::Proceed(page_size, paddr) => {
+                    self.current = self.current.checked_add_pages_with_size(1, page_size)?;
+                    if self.current > self.end {
+                        return None;
```
abrestic-rivos, line 1385: is this case even possible to hit given the validation we did prior to constructing the iterator?

abrestic-rivos (commented):
a couple minor comments, otherwise lgtm

riscv-page-tables/src/page_table.rs
sboeuf: Indeed! Given the predicate is usually going through the page tracker which maintains pages at a 4k granularity, we will need this information.

riscv-page-tables/src/page_table.rs
sboeuf: No I agree it should never happen given we already ran all the needed checks beforehand.
But that means we assume this iterator must always be used after such validation has been performed.

riscv-page-tables/src/page_table.rs
sboeuf: I agree, this is much simpler and much nicer :)

abrestic-rivos (approved):
thanks!

PR #288, opened 2023-03-17
Fixes and improvement for Salus unit testing

stillson:
* Fixed build so that salus and unit tests share dependencies
* added Testable trait to print which test is being run
* Added FailedAt to TestResult enumeration
* Expaned macros where needed
* Expanded test runner function
* Added multiple example tests
* Added two tests (MmioOpcodeTest and  CounterMaskIterTest) to show how testing works

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -147,19 +147,7 @@ objcopy_to_object(
     out = "umode.o",
 )
 
-rust_binary(
-    name = "salus",
-    srcs = glob(["src/*.rs"]),
-    compile_data = glob(["src/*.S"]) + [
-        ":umode_to_object",
-        ":l_rule",
-    ],
-    rustc_flags = [
-        "-Ctarget-feature=+v",
-        "--codegen=link-arg=-nostartfiles",
-        "-Clink-arg=-T$(location //:l_rule)",
-    ],
-    deps = [
+salus_deps = [
         "//attestation",
         "//data-model",
         "//device-tree",
@@ -188,7 +176,21 @@ rust_binary(
         "@salus-index//:arrayvec",
         "@salus-index//:memoffset",
         "@salus-index//:static_assertions",
+]
+
+rust_binary(
+    name = "salus",
+    srcs = glob(["src/*.rs"]),
+    compile_data = glob(["src/*.S"]) + [
+        ":umode_to_object",
+        ":l_rule",
     ],
+    rustc_flags = [
+        "-Ctarget-feature=+v",
+        "--codegen=link-arg=-nostartfiles",
+        "-Clink-arg=-T$(location //:l_rule)",
+    ],
+    deps = salus_deps,
 )
 
 rust_clippy(
@@ -216,33 +218,5 @@ rust_test(
         "-Clink-arg=-Tsrc/salus-test.lds",
         "--codegen=link-arg=-nostartfiles",
     ],
-    deps = [
-        "//attestation",
-        "//data-model",
-        "//device-tree",
-        "//drivers",
-        "//hyp-alloc",
-        "//mtt",
-        "//page-tracking",
-        "//rice",
-        "//riscv-elf",
-        "//riscv-page-tables",
-        "//riscv-pages",
-        "//riscv-regs",
-        "//s-mode-utils",
-        "//sbi-rs",
-        "//sync",
-        "//test-system",
-        "//u-mode-api",
-        "@rice-index//:const-oid",
-        "@rice-index//:der",
-        "@rice-index//:digest",
-        "@rice-index//:ed25519-dalek",
-        "@rice-index//:generic-array",
-        "@rice-index//:hkdf",
-        "@rice-index//:sha2",
-        "@salus-index//:arrayvec",
-        "@salus-index//:memoffset",
-        "@salus-index//:static_assertions",
-    ],
+    deps = salus_deps,
 )
--- a/src/main.rs
+++ b/src/main.rs
@@ -20,17 +20,9 @@
     negative_impls
 )]
 #![feature(custom_test_frameworks)]
-#![test_runner(crate::test_runner)]
+#![test_runner(crate::tests::test_runner)]
 #![reexport_test_harness_main = "test_main"]
 #![cfg_attr(test, allow(unused))]
-
-#[cfg(test)]
-fn test_runner(tests: &[&dyn Fn()]) {
-    println!("Running {} tests\n", tests.len());
-    for test in tests {
-        test();
-    }
-}
 
 use core::alloc::{Allocator, GlobalAlloc, Layout};
 use core::ptr::NonNull;
@@ -338,13 +330,6 @@ extern "C" fn primary_init(_hart_id: u64, _fdt_addr: u64) -> CpuParams {
 #[no_mangle]
 extern "C" fn primary_main() {}
 
-#[test_case]
-fn example_test() {
-    println!("example test");
-    assert!(true);
-    println!("OK\n");
-}
-
 /// Rust Entry point for the bootstrap CPU. Bootstraps salus, starts secondary CPUs and builds the
 /// HOST VM.
 ///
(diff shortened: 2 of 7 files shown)


atulkharerivos (approved):
Nice; LGTM.

src/vm.rs
```diff
@@ -73,6 +75,34 @@ impl MmioOpcode {
     }
 }
 
+#[test_case]
+fn MmioOpcodeTest() -> TestResult {
```
abrestic-rivos, line 79: is the purpose of the `TestResult` return type to avoid crashing the runner when a unit test panics, e.g. when an `assert!` fires? If so, should we be returning `Fail` instead of asserting below?

src/vm_pmu.rs
```diff
@@ -10,6 +10,9 @@ use sbi_rs::{
     PmuEventType, Result as SbiResult,
 };
 
+#[cfg(test)]
+use test_system::TestResult;
```
abrestic-rivos, line 14: can we keep all the test related stuff in each module in a `test` sub-module like one does with ordinary rust unit tests? that would reduce the conditional compilation and keep all the test stuff in one place.

src/main.rs
```diff
@@ -76,6 +68,34 @@ use smp::PerCpu;
...
+pub trait Testable {
+    fn run(&self) -> TestResult;
+}
+
+#[cfg(test)]
+impl<T> Testable for T
+where
+    T: Fn() -> TestResult,
+{
+    fn run(&self) -> TestResult {
```
abrestic-rivos, line 81: it would make it easier to write tests if `TestResult` were actually a result, because then you could use the `?` operator.

abrestic-rivos (commented):
this should probably be 2-3 separate commits:
- build deps fix
- `TestResult` return for unit tests
- adding example test cases

src/vm.rs
stillson: fix

src/vm_pmu.rs
stillson: done

src/main.rs
stillson: done

atulkharerivos: Think you pushed the wrong branch -- here's how it looks like for me: `fn run(&self) -> TestResult;` (instead of  `fn run(&self) -> Result<blah>;`

BUILD
```diff
@@ -187,8 +175,21 @@ rust_binary(
         "@rice-index//:signature",
         "@salus-index//:arrayvec",
         "@salus-index//:memoffset",
-        "@salus-index//:static_assertions",
+        "@salus-index//:static_assertions", ]
```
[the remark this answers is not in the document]
stillson: fixed

src/main.rs
stillson: TestResult is a type in test_system 
```
pub type TestResult = Result<(), TestFailure>;
```

stillson:
Done, needs to be split into commits

test-system/src/lib.rs
```diff
@@ -135,8 +138,55 @@ macro_rules! test_runtest {
...
+            Err(TestFailure::FailedAt(s)) => test_fail!($n, s),
+        }
+    };
+}
+
+/// Test if an expression is true, with a name for the test
+/// and a boolean to carry results from multiple test
+/// ```
+/// use test_system::*;
+/// let results = true;
```
atulkharerivos, line 152: Nit: Doc test will fail because results needs to mutable: `let mut results = true;`

test-system/src/lib.rs
```diff
@@ -135,8 +138,55 @@ macro_rules! test_runtest {
...
+            test_fail!($n)
+        }
+    };
+}
+
+/// Test if an expression is false, with a name for the test
+/// and a boolean to carry results from multiple test
+/// ```
+/// use test_system::*;
+/// let results = true;
```
atulkharerivos, line 175: Nit: Doc test will fail because results needs to mutable: let mut results = true;

src/vm.rs
stillson: Still?

test-system/src/lib.rs
stillson: fixed

src/main.rs
```diff
@@ -76,6 +68,49 @@ use smp::PerCpu;
 use sync::Once;
 use umode::UmodeTask;
 
+#[cfg(test)]
+mod tests {
```
abrestic-rivos, line 72: nit: can we put this `mod` at the bottom of the file so it's out of the way?

src/main.rs
```diff
@@ -76,6 +68,49 @@ use smp::PerCpu;
...
+
+    #[test_case]
+    fn example_fail() -> TestResult {
+        Err(TestFailure::Fail)
+    }
+
+    #[test_case]
+    fn example_fail_at() -> TestResult {
+        Err(TestFailure::FailedAt("example_fail_at"))
+    }
```
abrestic-rivos, line 112: drop these? you can put examples in a doc comment

test-system/src/lib.rs
```diff
@@ -135,8 +138,55 @@ macro_rules! test_runtest {
...
+        test_declare!($n);
+        let res: bool = $e;
+        $f = $f && res;
+        if res {
+            test_pass!($n);
+        } else {
+            test_fail!($n)
+        }
+    };
+}
```
abrestic-rivos, line 169: instead of accumulating test results in a `bool`, can this instead evaluate to a `TestResult` that we can use the `?` operator on?

e.g.

```
fn test_fn() -> TestResult {
    test_result_true!(1 == 1, "foo")?;
    Ok(())
}
```

src/vm.rs
```diff
@@ -73,6 +73,45 @@ impl MmioOpcode {
     }
 }
 
+#[cfg(test)]
+mod tests {
```
abrestic-rivos, line 77: similar to previous commit, the `mod tests` generally goes at the bottom of the file.

src/main.rs
stillson: fixed

src/main.rs
stillson: fixed

src/main.rs
stillson: fixed

test-system/src/lib.rs
stillson: fixed. (btw I don't think you can use the ? operator on a macro)

src/vm.rs
stillson: fixed

test-system/src/lib.rs
atulkharerivos: There shouldn't be inherent limitation on the use of the ? operator inside macros, and long as the encompassing scope can handle it in conjunction with ?. 

For example, the macro itself can return the `Result<>` as an expression, meaning that we can do something like:

```
#[macro_export]
macro_rules! test_result_true {
    ($e:expr, $n: expr) => {
       {
        test_declare!($n);
        let res: bool = $e;
        if res {
            test_pass!($n);
            Ok(())
        } else {
            test_fail!($n);
           Err(())
        }
    }
    }
}

test_result_true!(1 == 1, "One equals")?;
test_result_true(2 == 2, "Two equals")?
```

Anyway, I am not sure whether it's worth modifying what you have already implemented since it seems to have a similar effect.

test-system/src/lib.rs
```diff
@@ -135,8 +138,29 @@ macro_rules! test_runtest {
...
+            Ok(()) => test_pass!($n),
+            Err(TestFailure::Fail) => test_fail!($n),
+            Err(TestFailure::FailedAt(s)) => test_fail!($n, s),
         }
     };
 }
+
+/// Go from boolean to a TestResult.
+/// True will result in Ok, false will give a FailedAt.
+pub fn test_result_true(expr: bool, name: &'static str) -> TestResult {
```
abrestic-rivos, line 150: if you keep these as macros, you can use the expression itself as the message for `TestFailure::FailedAt()`, similar to `assert!`.

e.g. `test_result_true!(2 == 3)` becomes `FailedAt("2 == 3")`.

abrestic-rivos (commented):
just one ergonomic comment, otherwise lgtm

abrestic-rivos (approved):

PR #289, opened 2023-03-20
Add huge pages support

sboeuf:
Adding the huge pages support so that a guest can rely on pages greater than 4k in size.


sboeuf:
@abrestic-rivos Here is the first draft for supporting huge pages. It updates the existing code to handle huge pages and prevent errors from being returned. It also implements page promotion/demotion logic based on the new TEE-Host API.
Please let me know if there's an aspect of the huge pages support that I might have missed and that needs to be addressed. In the meantime, I'll need to come up with some testing in tellus to validate this can work as expected.

page-tracking/src/page_tracker.rs
```diff
@@ -516,79 +708,144 @@ impl PageTracker {
...
+                continue;
+            }
+            // We don't need to touch the destination, but we need to make sure it isn't linked.
+            let dst = page_tracker.get(next)?;
+            if dst.next().is_some() {
+                return Err(Error::PageAlreadyLinked);
+            }
+            let src = page_tracker.get_mut(prev)?;
+            src.link(next)?;
+            prev = next;
```
abrestic-rivos, line 914: so is the idea that pages in the list are always linked at 4k and caller has to decide what granularity they are unlinked at? in that case, the size should probably be part of `PageList`. also could use some documentation.

drivers/src/iommu/msi_page_table.rs
abrestic-rivos, line 1: please include a description of the change in the commit message. this is a pretty large patch and there should be some documentation in the commit message about what's changing and why.

src/vm_pages.rs
```diff
@@ -1481,14 +1481,18 @@ impl<'a, T: GuestStagePagingMode> FinalizedVmPages<'a, T> {
...
-                    P::mem_type(),
-                )
-            })
+            .invalidate_range(
+                page_addr,
+                num_pages * PageSize::Size4k as u64,
+                |addr, ps| {
+                    self.inner.page_tracker.is_mapped_page(
+                        addr,
+                        ps,
```
abrestic-rivos, line 1492: we always convert pages at 4k granularity since the host is exclusively mapped with 4k pages.

riscv-page-tables/src/page_table.rs
```diff
@@ -737,6 +779,26 @@ impl<T: PagingMode> PageTableInner<T> {
             _ => Err(Error::PageNotInvalidated),
         }
     }
+
+    /// Promotes the table entry at `vaddr` as a valid leaf.
+    fn promote_leaf(
```
abrestic-rivos, line 784: `promote_pte()`? we're not actually promoting a leaf; we're turning a table into a leaf pte.

riscv-page-tables/src/page_table.rs
```diff
@@ -737,6 +779,26 @@ impl<T: PagingMode> PageTableInner<T> {
...
+        while let Table(t) = entry {
+            if t.level().leaf_page_size() != page_size {
+                entry = t.table().entry_for_addr_mut(vaddr);
+            } else {
+                let table_addr = t.table_addr();
+                let page: Page<InternalDirty> = unsafe { Page::new(table_addr) };
+                return Ok((t.leaf(), page.clean()));
+            }
+        }
+        Err(Error::UpdatePteAsLeaf)
```
abrestic-rivos, line 800: `PteNotPromotable` might be a more descriptive error code

riscv-page-tables/src/page_table.rs
```diff
@@ -1207,6 +1269,137 @@ impl<T: PagingMode> GuestStagePageTable<T> {
...
+    /// the same state, either "Blocked" and uniquely owned by the TVM, or "BlockedShared" and not
+    /// owned by the TVM. Last thing to check is that each page must have been invalidated with a
+    /// TLB version that's older than the current one.
+    pub fn promote_page(
+        &self,
+        vaddr: PageAddr<T::MappedAddressSpace>,
+        requested_page_size: PageSize,
+        expected_page_size: PageSize,
+        page_tracker: PageTracker,
+        owner: PageOwnerId,
```
abrestic-rivos, line 1284: these are members of `GuestStagePageTable`, why do we need to pass them in here?

riscv-page-tables/src/page_table.rs
```diff
@@ -1207,6 +1269,137 @@ impl<T: PagingMode> GuestStagePageTable<T> {
...
+    /// smaller than the requested page size. Also checks the pages are physically contiguous and
+    /// suitably aligned to the requested page size. It's important that all the pages must be in
+    /// the same state, either "Blocked" and uniquely owned by the TVM, or "BlockedShared" and not
+    /// owned by the TVM. Last thing to check is that each page must have been invalidated with a
+    /// TLB version that's older than the current one.
+    pub fn promote_page(
+        &self,
+        vaddr: PageAddr<T::MappedAddressSpace>,
+        requested_page_size: PageSize,
+        expected_page_size: PageSize,
```
abrestic-rivos, line 1282: i'd only pass one of these in in since one can be inferred from the other

riscv-page-tables/src/page_table.rs
```diff
@@ -1207,6 +1269,137 @@ impl<T: PagingMode> GuestStagePageTable<T> {
...
+            if !page_tracker.is_promotable_page(
+                paddr,
+                page_size,
+                owner,
+                MemType::Ram,
+                prev_page_state,
+                tlb_version,
+            ) {
+                return Err(Error::PageNotPromotable);
+            }
```
abrestic-rivos, line 1338: to be consistent with the other `PageTable` methods we should do this `PageTracker` checking in a closure provided by the caller. and the caller can determine whether the page to be promoted lies within a confidential or non-confidential memory region, so they need only check whether it's blocked at an older tlb version (i.e. no need to check that the state of each page is the same).

riscv-page-tables/src/page_table.rs
```diff
@@ -1207,6 +1269,137 @@ impl<T: PagingMode> GuestStagePageTable<T> {
...
+        for va in vaddr
+            .iter_from_with_size(requested_page_size)
+            .unwrap()
+            .take(512)
+        {
+            let entry = page_table.entry_for_addr_mut(va.into());
+            use TableEntryType::*;
+            let u = match entry {
+                Unused(u) => u,
+                _ => return Err(Error::PteNotUnused),
```
abrestic-rivos, line 1396: should be `unreachable()!` since the page table must've been zeroed out.

riscv-page-tables/src/page_table.rs
```diff
@@ -1207,6 +1269,137 @@ impl<T: PagingMode> GuestStagePageTable<T> {
...
+        // Installs a non-leaf PTE pointing the newly-allocated page-table page in place of the huge-page leaf PTE
+        // Safety: `page.addr()` references a page uniquely owned by the root `GuestStagePageTable`.
+        let table_entry = unsafe { pte.map_table(free_page.addr()) };
+        let mut page_table = table_entry.table();
+        // Fills in the free page-table page mapping the sub-pages of the huge page to be split,
+        // marking each PTE as present.
+        for va in vaddr
+            .iter_from_with_size(requested_page_size)
+            .unwrap()
+            .take(512)
```
abrestic-rivos, line 1390: `ENTRIES_PER_PAGE`

riscv-page-tables/src/page_table.rs
sboeuf: Agreed.

riscv-page-tables/src/page_table.rs
sboeuf: :+1:

riscv-page-tables/src/page_table.rs
sboeuf: Ok I'll do the calculation directly inside this function.

riscv-page-tables/src/page_table.rs
sboeuf: We don't :)
I'll fix it!

riscv-page-tables/src/page_table.rs
sboeuf: Ah ok I understand. The reason I couldn't use the closure design was because I wanted to check every page state was the same, but if we rely on memory region for that, then that should solve the problem.

riscv-page-tables/src/page_table.rs
sboeuf: Agreed.

riscv-page-tables/src/page_table.rs
sboeuf: :+1:

riscv-page-tables/src/pte.rs
```diff
@@ -153,10 +153,15 @@ impl Pte {
...
     /// Returns `true` if the entry is a leaf.
-    pub fn leaf(&self) -> bool {
+    pub fn is_leaf(&self) -> bool {
         self.bits() & MASK_RWX != 0
     }
 
+    /// Marks the entry as a leaf.
+    pub fn leaf(&mut self) {
+        self.0 |= MASK_RWX;
+    }
```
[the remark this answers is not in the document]
sboeuf: Yes I can use `set()` if you prefer, but it is `unsafe`, which is why I wasn't too keen to use it.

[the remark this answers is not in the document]
sboeuf: Also, `set()` expects the PFN, which we don't need to update.

page-tracking/src/page_tracker.rs
abrestic-rivos: yeah it does. we should try to optimize this by only linking the "head" page together, but that can be deferred til later.

src/vm_pages.rs
```diff
@@ -1387,12 +1404,16 @@ impl<'a, T: GuestStagePagingMode> FinalizedVmPages<'a, T> {
         // Lock the pages for assignment.
         let mut locked_pages = LockedPageList::new(self.inner.page_tracker());
         for paddr in converted {
+            if !paddr.is_aligned(page_size) {
+                continue;
+            }
+
```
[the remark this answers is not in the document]
abrestic-rivos: ah ok, that makes sense now. maybe we could make this a little more concise using `for paddr in converted.filter(...)` ?

src/vm_pages.rs
abrestic-rivos: well, we should reject any `ps` that's not 4k

src/vm_pages.rs
```diff
@@ -1403,7 +1403,7 @@ impl<'a, T: GuestStagePagingMode> FinalizedVmPages<'a, T> {
 
         // Lock the pages for assignment.
         let mut locked_pages = LockedPageList::new(self.inner.page_tracker());
-        for paddr in converted {
+        for (paddr, _) in converted {
             if !paddr.is_aligned(page_size) {
                 continue;
             }
```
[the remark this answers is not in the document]
abrestic-rivos: yeah makes sense now

riscv-page-tables/src/page_table.rs
```diff
@@ -1207,6 +1269,137 @@ impl<T: PagingMode> GuestStagePageTable<T> {
...
+        let mut inner = self.inner.lock();
+        let pte = inner.get_invalidated_leaf(vaddr)?;
+        let paddr = pte.page_addr();
+        let page_size = pte.level().leaf_page_size();
+        // Check the page matches the expected page size
+        if page_size != expected_page_size {
+            return Err(Error::InvalidPageSize);
+        }
+        // Check the page state
+        if !page_tracker.is_demotable_page(paddr, page_size, owner, MemType::Ram, tlb_version) {
```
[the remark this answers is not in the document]
abrestic-rivos: yes right, the comments in `promote()` apply here as well.

riscv-page-tables/src/page_table.rs
```diff
@@ -183,6 +199,13 @@ impl<'a, T: PagingMode> UnusedPte<'a, T> {
         self.pte.lock();
         LockedUnmappedPte::new(self.pte, self.level)
     }
+
+    /// Makes the entry a valid leaf.
+    fn leaf(self) -> LeafPte<'a, T> {
+        self.pte.leaf();
+        self.pte.mark_valid();
+        LeafPte::new(self.pte, self.level)
+    }
```
abrestic-rivos, line 208: how can we turn an unused PTE into a valid leaf? the PFN is zero at this point, so we need to pass in the PFN to be set.

riscv-page-tables/src/page_table.rs
```diff
@@ -301,6 +337,12 @@ impl<'a, T: PagingMode> PageTablePte<'a, T> {
         //    `GuestStagePageTable` and have their lifetime bound to the root.
         unsafe { PageTable::from_pte(self.pte, self.level.next().unwrap()) }
     }
+
+    /// Makes the entry a valid leaf.
+    fn leaf(self) -> LeafPte<'a, T> {
+        self.pte.leaf();
+        LeafPte::new(self.pte, self.level)
+    }
```
abrestic-rivos, line 345: similar to the "unused" case, i don't think this is correct. the PFN of this PTE is currently set to the PFN of the table page. when promoting, we want to set the PFN to the base of the huge page that this PTE is supposed to map.

riscv-page-tables/src/pte.rs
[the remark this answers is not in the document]
abrestic-rivos: forgot to comment on this the first time, but we definitely do need to update the PFN. see my comments on where `leaf()` is used.

riscv-page-tables/src/page_table.rs
sboeuf: Gotcha, something I had completely missed.

sboeuf: Fixed. PTAL

riscv-page-tables/src/page_table.rs
sboeuf: Fixed. PTAL

src/vm_pages.rs
```diff
@@ -1390,10 +1390,10 @@ impl<'a, T: GuestStagePagingMode> FinalizedVmPages<'a, T> {
...
             .root
-            .get_invalidated_pages(page_addr, num_pages * page_size as u64, |addr| {
+            .get_invalidated_pages(page_addr, num_pages * page_size as u64, |addr, ps| {
                 self.inner.page_tracker.is_converted_page(
                     addr,
-                    PageSize::Size4k,
+                    ps,
                     self.inner.page_owner_id,
                     P::mem_type(),
                     version,
```
abrestic-rivos: for any of these functions where we're trying to create a hugepage out of 4k pages, we need to check that the pages that are mapped are contiguous and suitably aligned. i.e. the base GPA and SPA that it maps are aligned to the hugepage size, and that for each SPA we encounter in the page table it is contiguous to the one previous.

src/vm_pages.rs
abrestic-rivos: this applies to everywhere else we're getting pages from the host.

page-tracking/src/page_tracker.rs
```diff
@@ -733,6 +733,39 @@ impl PageTracker {
         true
     }
 
+    /// Returns true if `addr` is a page owned by `owner` and it's in a "Blocked" state with
+    /// type `mem_type`, or if `addr` is a page not owned by `owner` and it's in a "SharedBlocked"
+    /// state with type `mem_type`. Additionally, the TLB version of the page must be older than
+    /// the version provided by the caller.
+    pub fn is_blocked_page_with_older_tlb(
```
abrestic-rivos, line 740: this is basically the same check as `is_removable_page`. can we combine them into a single function?

riscv-page-tables/src/page_table.rs
```diff
@@ -204,10 +232,23 @@ impl<'a, T: PagingMode> InvalidatedPte<'a, T> {
...
     }
+
+    /// Marks this invalid PTE as valid and maps it to a next-level page table at `table_paddr`.
+    /// Returns this entry as a valid table entry.
+    ///
+    /// # Safety
+    ///
+    /// The caller must guarantee that `table_paddr` references a page-table page uniquely owned by
+    /// the root `GuestStagePageTable`.
+    unsafe fn map_table(self, table_paddr: SupervisorPageAddr) -> PageTablePte<'a, T> {
```
abrestic-rivos, line 248: i'd call this `demote()`

riscv-page-tables/src/page_table.rs
```diff
@@ -301,6 +342,20 @@ impl<'a, T: PagingMode> PageTablePte<'a, T> {
...
         unsafe { PageTable::from_pte(self.pte, self.level.next().unwrap()) }
     }
+
+    /// Makes the entry a valid leaf.
+    ///
+    /// # Safety
+    ///
+    /// The caller must guarantee that `paddr` references a page uniquely owned by
+    /// the root `GuestStagePageTable`.
+    unsafe fn leaf(self, paddr: SupervisorPageAddr) -> LeafPte<'a, T> {
```
abrestic-rivos, line 352: ... and this `promote()`

riscv-page-tables/src/page_table.rs
```diff
@@ -737,6 +792,27 @@ impl<T: PagingMode> PageTableInner<T> {
...
+    ) -> Result<(LeafPte<T>, Page<InternalClean>)> {
+        let mut entry = PageTable::from_root(self).entry_for_addr_mut(vaddr);
+        use TableEntryType::*;
+        while let Table(t) = entry {
+            if t.level().leaf_page_size() != page_size {
+                entry = t.table().entry_for_addr_mut(vaddr);
+            } else {
+                let table_addr = t.table_addr();
+                let page: Page<InternalDirty> = unsafe { Page::new(table_addr) };
+                return Ok((unsafe { t.leaf(paddr) }, page.clean()));
```
abrestic-rivos, line 811: calls to unsafe functions need "Safety" comments

riscv-page-tables/src/page_table.rs
```diff
@@ -1207,6 +1283,137 @@ impl<T: PagingMode> GuestStagePageTable<T> {
...
+    where
+        F: FnMut(SupervisorPageAddr, PageSize) -> bool,
+    {
+        let expected_page_size = requested_page_size
+            .size_down()
+            .ok_or(Error::InvalidPageSize)?;
+        let num_pages = PageSize::num_4k_pages(requested_page_size as u64);
+        let end = vaddr
+            .checked_add_pages(num_pages)
+            .ok_or(Error::AddressOverflow)?;
```
abrestic-rivos, line 1307: need to check that `vaddr` is aligned. if it is, you're guaranteed not to overflow.

riscv-page-tables/src/page_table.rs
```diff
@@ -1207,6 +1283,137 @@ impl<T: PagingMode> GuestStagePageTable<T> {
...
+    where
+        F: FnMut(SupervisorPageAddr, PageSize) -> bool,
+    {
+        let expected_page_size = requested_page_size
+            .size_up()
+            .ok_or(Error::InvalidPageSize)?;
+        vaddr
+            .checked_add_pages_with_size(1, expected_page_size)
+            .ok_or(Error::AddressOverflow)?;
+        let mut inner = self.inner.lock();
```
abrestic-rivos, line 1377: same comment as above. check if vaddr is aligned first, and then you know it can't overflow.

abrestic-rivos (commented):
btw, another thing you'll need to do for this to actually work that i realized we hadn't done yet is to map the host such that it maps physical memory in 2M chunks. in other words, for each 2M-aligned chunk of host VM GPA space, the base of that chunk should map to a 2M-aligned SPA, and all the 4k pages within that 2M chunk should be contiguous in SPA space.

currently we only do 16kB chunks so that it can create the base of the G-stage page-tables.

page-tracking/src/page_tracker.rs
```diff
@@ -235,30 +250,52 @@ impl PageTracker {
...
-        let info = page_tracker.get_mut(page.addr()).unwrap();
-        info.release(false)?;
+        for paddr in page
+            .addr()
+            .iter_from()
+            .take(PageSize::num_4k_pages(page.size() as u64) as usize)
+        {
+            let info = page_tracker.get_mut(paddr).unwrap();
+            info.release(false)?;
+        }
```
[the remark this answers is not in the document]
sboeuf: I figured out how to simplify it creating two new functions accepting closures.

sboeuf:
@abrestic-rivos 
>btw, another thing you'll need to do for this to actually work that i realized we hadn't done yet is to map the host such that it maps physical memory in 2M chunks. in other words, for each 2M-aligned chunk of host VM GPA space, the base of that chunk should map to a 2M-aligned SPA, and all the 4k pages within that 2M chunk should be contiguous in SPA space.

Just to clarify, when you say 2M, that's because you assume 2M huge pages, but I assume for 1G huge page, the host must map in 1G chunks, is that correct?

>currently we only do 16kB chunks so that it can create the base of the G-stage page-tables.

Can you point me to the place where this is performed in the code?

src/vm_pages.rs
sboeuf: Ah yes we do need to check it's contiguous and suitably aligned. I think this check should be handled directly inside `get_invalidated_pages()`, very similar to what I did with `promote_page()`. Or do you have another approach in mind?

abrestic-rivos:
> Just to clarify, when you say 2M, that's because you assume 2M huge pages, but I assume for 1G huge page, the host must map in 1G chunks, is that correct?

Right. But mapping the host in 1G-aligned chunks would waste a ton of memory, especially since we're only booting with ~4GB in our current QEMU config. Using 2M chunks still gives us the ability to exercise hugepages without being too much of a burden.

> Can you point me to the place where this is performed in the code?

See `PageTracker::from()`, and then where those pages are used in `HostVm::add_measured_pages()` and `HostVm::add_zero_pages()`.

src/vm_pages.rs
sboeuf: If I do it in the closure, that means I have to update the closure signature so that it would provide the previous address along with the current one. That means the logic of checking pages are contiguous and aligned must be implemented in multiple places. I was thinking about passing an extra parameter to the function `get_invalidated_pages()` and `get_mapped_pages()` so that they would know if the extra check for contiguous pages must be performed.

sboeuf:
> > Just to clarify, when you say 2M, that's because you assume 2M huge pages, but I assume for 1G huge page, the host must map in 1G chunks, is that correct?
> 
> Right. But mapping the host in 1G-aligned chunks would waste a ton of memory, especially since we're only booting with ~4GB in our current QEMU config. Using 2M chunks still gives us the ability to exercise hugepages without being too much of a burden.

Got it!

> 
> > Can you point me to the place where this is performed in the code?
> 
> See `PageTracker::from()`, and then where those pages are used in `HostVm::add_measured_pages()` and `HostVm::add_zero_pages()`.

Thanks!

src/vm_pages.rs
abrestic-rivos: > If I do it in the closure, that means I have to update the closure signature so that it would provide the previous address along with the current one.

can't the closure mutably capture a previous phys addr?

sboeuf: Let me have a try!

src/host_vm.rs
```diff
@@ -282,11 +282,12 @@ impl<T: GuestStagePagingMode> HostVmLoader<T> {
...
         }
 
-        // Host guarantees that the host pages it returns start at T::TOP_LEVEL_ALIGN-aligned block,
-        // and because we built the HwMemMap with a minimum region alignment of T::TOP_LEVEL_ALIGN
+        // Host guarantees that the host pages it returns start at 2M-aligned block,
+        // and because we built the HwMemMap with a minimum region alignment of 2M,
         // any discontiguous ranges are also guaranteed to be aligned.
         //
         // Now fill in the address space, inserting zero pages around the kernel/initramfs/FDT.
+        assert!(self.guest_ram_base.is_aligned(PageSize::Size2M));
```
sboeuf, line 290: @abrestic-rivos looking at the code that comes after this, I'm wondering if we should ensure that `num_pages`, `num_kernel_pages`, etc are covering a 2M size multiple. Because if that's not the case, we might end up with `current_gpa` not being aligned on 2M, and that would mean some `zero_ranges` range might not be properly aligned. What do you think?

src/host_vm.rs
```diff
@@ -167,15 +167,15 @@ impl<T: GuestStagePagingMode> HostVmLoader<T> {
         mut page_alloc: HypPageAlloc,
     ) -> Self {
         // Reserve a contiguous chunk for the host's FDT. We assume it will be no bigger than the
-        // size of the hypervisor's FDT and we align it to `T::TOP_LEVEL_ALIGN` to maintain the
-        // contiguous mapping guarantee from GPA -> HPA mentioned above.
+        // size of the hypervisor's FDT and we align it to 2M to maintain the contiguous mapping
```
abrestic-rivos, line 171: the `2M` should be a `const` somewhere instead of hard-coding `PageSize::Size2M` everywhere.

src/host_vm.rs
abrestic-rivos: yes, the kernel & initramfs regions should be 2M-aligned and rounded up to 2M in length.

src/host_vm.rs
```diff
@@ -305,11 +306,15 @@ impl<T: GuestStagePagingMode> HostVmLoader<T> {
         };
         self.vm
             .add_measured_pages(current_gpa, kernel_pages.into_iter());
-        current_gpa = current_gpa.checked_add_pages(num_kernel_pages).unwrap();
+        current_gpa = current_gpa
+            .checked_add_pages_with_size(
+                (num_kernel_pages / 512) + ((num_kernel_pages % 512 != 0) as u64),
+                PageSize::Size2M,
+            )
```
abrestic-rivos, line 313: when we create the `HwMemRegion` for the kernel, we should just round up the size to 2M so that we don't have to do this. same for initramfs.

src/host_vm.rs
```diff
@@ -336,7 +346,12 @@ impl<T: GuestStagePagingMode> HostVmLoader<T> {
         let num_fdt_pages = fdt_pages.len();
         self.vm
             .add_measured_pages(current_gpa, fdt_pages.into_iter());
-        current_gpa = current_gpa.checked_add_pages(num_fdt_pages).unwrap();
+        current_gpa = current_gpa
+            .checked_add_pages_with_size(
+                (num_fdt_pages / 512) + ((num_fdt_pages % 512 != 0) as u64),
+                PageSize::Size2M,
+            )
+            .unwrap();
```
abrestic-rivos, line 354: just allocate a multiple of 2M FDT pages up front.

riscv-page-tables/src/page_table.rs
```diff
@@ -1207,6 +1284,141 @@ impl<T: PagingMode> GuestStagePageTable<T> {
...
+    /// suitably aligned to the requested page size. It's important that all the pages must be in
+    /// the same state, either "Blocked" and uniquely owned by the TVM, or "BlockedShared" and not
+    /// owned by the TVM. Last thing to check is that each page must have been invalidated with a
+    /// TLB version that's older than the current one.
+    pub fn promote_page<F>(
+        &self,
+        vaddr: PageAddr<T::MappedAddressSpace>,
+        requested_page_size: PageSize,
+        mut pred: F,
+    ) -> Result<(SupervisorPageAddr, PageSize, Page<InternalClean>)>
```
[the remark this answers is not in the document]
sboeuf: You're right, that's not needed.

riscv-page-tables/src/page_table.rs
```diff
@@ -1207,6 +1284,141 @@ impl<T: PagingMode> GuestStagePageTable<T> {
...
+            if !pred(paddr, page_size) {
+                return Err(Error::PredicateFailed);
+            }
+
+            // Move to next page
+            va = Self::check_and_increment_vaddr(va, end, page_size)?;
+        }
+
+        // Promotes the non-leaf PTE and retrieves the now-unused page-table page
+        let paddr = inner.get_invalidated_leaf(vaddr)?.page_addr();
```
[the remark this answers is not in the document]
sboeuf: Sure!

riscv-page-tables/src/page_table.rs
```diff
@@ -1207,6 +1284,141 @@ impl<T: PagingMode> GuestStagePageTable<T> {
...
+        let (leaf, free_page) = inner.promote_pte(vaddr.into(), paddr, requested_page_size)?;
+
+        Ok((leaf.page_addr(), leaf.level().leaf_page_size(), free_page))
+    }
+
+    /// Verifies the virtual address range is covering only one invalidated page that matches the
+    /// expected page size. Also checks the page state is either "Blocked" and uniquely owned by
+    /// the TVM, or "BlockedShared" and not owned by the TVM. The page must have been invalidated
+    /// with a TLB version that's older than the current one.
+    pub fn demote_page<F>(
```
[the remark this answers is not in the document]
sboeuf: I've updated the comment, but I need to return the page size given it's gonna different than the requested page size in that case.

riscv-page-tables/src/page_table.rs
```diff
@@ -1207,6 +1284,141 @@ impl<T: PagingMode> GuestStagePageTable<T> {
...
+        &self,
+        vaddr: PageAddr<T::MappedAddressSpace>,
+        requested_page_size: PageSize,
+        free_page: Page<InternalClean>,
+        mut pred: F,
+    ) -> Result<(SupervisorPageAddr, PageSize)>
```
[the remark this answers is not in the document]
sboeuf: Yes we do because it's different from `promote_page()`, in that case the page size returned isn't the `requested_page_size`.

src/host_vm.rs
sboeuf: Agreed, this will simplify the code.

riscv-page-tables/src/page_table.rs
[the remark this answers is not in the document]
sboeuf: Yes I can drop the PageSize from the return value for both functions

sboeuf:
Weird. I've updated tellus/guestvm through the commit `test-workloads: Test huge pages support`, but the change isn't part of the test build. I thought the docker image `rivosinc/rivos-run-salus-qemu@v1` was created on the fly and therefore any change to `test-workloads` would be propagated. Did I miss something?

glg-rv:
These changes look good to me. Admit I missed most of the discussions happening on this PR and haven't checked every single part of the page size adds to parameters, but it is much simpler than what I was fearing.

glg-rv:
> > Admit I missed most of the discussions happening on this PR and haven't checked every single part of the page size adds to parameters, but it is much simpler than what I was fearing.
> 
> Well maybe it's simpler than you thought because we're not supporting huge pages everywhere. It's only the ability to map huge pages for TVMs and not to use them for internal states or other purposes.

Heh. Well, I saw that. I think it's a good first step.

Regarding the internal mappings, we might need to sync over the plans because I have a few upcoming changes on the hypervisor mappings.

riscv-page-tables/src/page_table.rs
```diff
@@ -737,6 +792,33 @@ impl<T: PagingMode> PageTableInner<T> {
...
+        vaddr: RawAddr<T::MappedAddressSpace>,
+        paddr: SupervisorPageAddr,
+        page_size: PageSize,
+    ) -> Result<(LeafPte<T>, Page<InternalClean>)> {
+        let mut entry = PageTable::from_root(self).entry_for_addr_mut(vaddr);
+        use TableEntryType::*;
+        while let Table(t) = entry {
+            if t.level().leaf_page_size() != page_size {
+                entry = t.table().entry_for_addr_mut(vaddr);
+            } else {
```
abrestic-rivos, line 813: nit: this is a bit easier to read if you do `if <page size matches> { promote } else { next PTE }` than the other way around.

src/vm_pages.rs
```diff
@@ -1805,6 +1811,80 @@ impl<'a, T: GuestStagePagingMode> FinalizedVmPages<'a, T> {
...
+    // Promotes a page range, creating one page of the next larger page size.
+    pub fn promote_page(&self, page_addr: GuestPageAddr, page_size: PageSize) -> Result<()> {
+        let end = page_addr
+            .checked_add_pages_with_size(1, page_size)
+            .ok_or(Error::AddressOverflow)?;
+        let regions = self.inner.regions.read();
+        if !(regions.contains(page_addr, end, |r| r == VmRegionType::Confidential)
+            || regions.contains(page_addr, end, |r| r == VmRegionType::ConfidentialRemovable)
+            || regions.contains(page_addr, end, |r| r == VmRegionType::Shared)
+            || regions.contains(page_addr, end, |r| r == VmRegionType::SharedRemovable))
```
abrestic-rivos, line 1823: nit: check the possible region types in the closure rather than calling `contains()` 4 times. same goes for `demote_page()`

sboeuf: That's actually on purpose that I'm calling 4 times the `contains()` function, because I want to make sure the whole region contains the same type. If I move to a single check with the closure handling all the types at once, that means we're okay promoting a page on a range that might be of different types.

sboeuf:
@abrestic-rivos I've addressed your comments and I've added a few unit tests.

src/vm_pages.rs
abrestic-rivos: `contains()` won't span regions; the range must fall entirely in one region.

sboeuf: Ah good point :)

sboeuf:
PR has been rebased on latest main branch.

abrestic-rivos (approved):

PR #290, opened 2023-03-22
Map Salus stack in a fixed VA address.

### LINKED ISSUE #240: Insert guard pages below per-CPU stacks
@rajnesh-kanwal encountered a panic in #239 that turned out to be due to a stack overflow on one CPU scribbling over the `PerCpu` struct of another CPU. We should really be inserting guard pages between the CPU stacks so that we can get an easy to diagnose page-fault.

glg-rv:
Small patches, profound changes. Fixes #240 

This PR maps salus stack in a VA space. Initialization happens now in two stages.

The first stage (init) is when everything (including stack) is mapped in a contiguous physical mapping in the VA.
The second stage (launch) is called once the page table and the va-mapped stack has been setup.

Ideally, we whould like to push as much as possible into the second stage. But this is future work and requires more consideration.

### DIFF
--- a/src/hyp_layout.rs
+++ b/src/hyp_layout.rs
@@ -83,21 +83,16 @@ pub const UMODE_INPUT_SIZE: u64 = 4 * 1024;
 const_assert!(PageSize::Size4k.is_aligned(UMODE_INPUT_SIZE));
 
 /// Address of the hypervisor stack top.
-#[allow(dead_code)]
 pub const HYP_STACK_TOP: u64 = 0xffff_ffff_ffe0_0000;
 // Align stack top to 2M to ease huge page mappings.
 const_assert!(PageSize::Size2M.is_aligned(HYP_STACK_TOP));
 /// Number of 4k-pages reserved for each CPU as their stack.
-#[allow(dead_code)]
 pub const HYP_STACK_PAGES: u64 = 0x80;
 /// Size of stack in bytes.
-#[allow(dead_code)]
 pub const HYP_STACK_SIZE: u64 = HYP_STACK_PAGES * PageSize::Size4k as u64;
 /// End of stack (lower address)
-#[allow(dead_code)]
 pub const HYP_STACK_BOTTOM: u64 = HYP_STACK_TOP - HYP_STACK_SIZE;
 /// Page address of the stack bottom.
-#[allow(dead_code)]
 pub const HYP_STACK_BOTTOM_PAGE_ADDR: PageAddr<SupervisorVirt> =
     PageAddr::new_const::<HYP_STACK_BOTTOM>();
 
--- a/src/hyp_layout.rs
+++ b/src/hyp_layout.rs
@@ -22,6 +22,12 @@
 // +-------------------------+ UMODE_INPUT_START
 // | Umode Input Area        |
 // +-------------------------+ +UMODE_INPUT_SIZE
+// | (unused)                |
+// +-------------------------+ HYP_STACK_BOTTOM (HYP_STACK_TOP - HYP_STACK_SIZE)
+// | Hypervisor Stack        |
+// +-------------------------+ HYP_STACK_TOP (0xffff_ffff_ffe0_0000)
+// | (unused 2Mb)            |
+// +-------------------------+ End of Address Space.
 
 use riscv_pages::{PageAddr, PageSize, SupervisorVirt};
 use static_assertions::const_assert;
@@ -75,6 +81,25 @@ const_assert!(PageSize::Size4k.is_aligned(UMODE_INPUT_START));
 /// Size of the U-mode Input Region.
 pub const UMODE_INPUT_SIZE: u64 = 4 * 1024;
 const_assert!(PageSize::Size4k.is_aligned(UMODE_INPUT_SIZE));
+
+/// Address of the hypervisor stack top.
+#[allow(dead_code)]
+pub const HYP_STACK_TOP: u64 = 0xffff_ffff_ffe0_0000;
+// Align stack top to 2M to ease huge page mappings.
+const_assert!(PageSize::Size2M.is_aligned(HYP_STACK_TOP));
+/// Number of 4k-pages reserved for each CPU as their stack.
+#[allow(dead_code)]
+pub const HYP_STACK_PAGES: u64 = 0x80;
+/// Size of stack in bytes.
+#[allow(dead_code)]
+pub const HYP_STACK_SIZE: u64 = HYP_STACK_PAGES * PageSize::Size4k as u64;
+/// End of stack (lower address)
+#[allow(dead_code)]
+pub const HYP_STACK_BOTTOM: u64 = HYP_STACK_TOP - HYP_STACK_SIZE;
+/// Page address of the stack bottom.
+#[allow(dead_code)]
+pub const HYP_STACK_BOTTOM_PAGE_ADDR: PageAddr<SupervisorVirt> =
+    PageAddr::new_const::<HYP_STACK_BOTTOM>();
 
 // Returns true if `addr` is contained in the U-mode binary area.
 fn is_umode_binary_addr(addr: u64) -> bool {
--- a/src/hyp_map.rs
+++ b/src/hyp_map.rs
@@ -329,7 +329,6 @@ pub struct HypStackRegion {
     paddr: PageAddr<SupervisorPhys>,
 }
 
-#[allow(dead_code)]
 impl HypStackRegion {
     fn new(stack_pages: SequentialPages<InternalDirty>) -> Result<Self, Error> {
         let page_count = stack_pages.len();
@@ -577,7 +576,11 @@ impl HypMap {
     }
 
     /// Creates a new page table based on this memory map.
-    pub fn new_page_table(&self, hyp_mem: &mut HypPageAlloc) -> HypPageTable {
+    pub fn new_page_table(
+        &self,
+        hyp_mem: &mut HypPageAlloc,
+        stack_pages: SequentialPages<InternalDirty>,
+    ) -> HypPageTable {
         // Create empty sv48 page table
         // Unwrap okay: we expect to have at least one page free.
         let root_page = hyp_mem
@@ -597,6 +600,10 @@ impl HypMap {
         for r in &self.umode_elf_regions {
             r.map(&sv48, hyp_mem);
         }
+        // Alloc and map the hypervisor stack for this page-table.
+        HypStackRegion::new(stack_pages)
+            .expect("Hypervisor stack allocation failed")
+            .map(&sv48, hyp_mem);
         // Alloc and map the U-mode Input Region for this page-table.
         let umode_input = UmodeInputRegion::map(&sv48, hyp_mem);
         // Alloc pte_pages for U-mode mappings.
--- a/src/hyp_map.rs
+++ b/src/hyp_map.rs
@@ -89,7 +89,6 @@ impl HwMapRegion {
                 Some(PteLeafPerms::R)
             }
             HwMemRegionType::Reserved(HwReservedMemType::HypervisorHeap)
-            | HwMemRegionType::Reserved(HwReservedMemType::HypervisorPerCpu)
             | HwMemRegionType::Reserved(HwReservedMemType::PageMap)
             | HwMemRegionType::Mmio(_) => Some(PteLeafPerms::RW),
         };
--- a/src/hyp_map.rs
+++ b/src/hyp_map.rs
@@ -35,6 +35,9 @@ type UmodeElfRegionsVec = ArrayVec<UmodeElfRegion, MAX_UMODE_ELF_REGIONS>;
 // Hw map regions vector.
 type HwMapRegionsVec = ArrayVec<HwMapRegion, MAX_HW_MAP_REGIONS>;
 
+// Global reference to the Hypervisor Map.
+static HYPMAP: Once<HypMap> = Once::new();
+
 /// Errors returned by creating or modifying hypervisor mappings.
 #[derive(Debug)]
 pub enum Error {
@@ -254,6 +257,72 @@ impl UmodeElfRegion {
     }
 }
 
+/// The U-mode Input Region is a private region of the page table used by the hypervisor to pass
+/// data to the current U-mode operation. This avoids mapping hypervisor data directly in
+/// U-mode. This area is written by the hypervisor and mapped read-only in U-mode.
+struct UmodeInputRegion {
+    /// Volatile Slice used to write to this area from the hypervisor.
+    vslice: VolatileSlice<'static>,
+}
+
+impl UmodeInputRegion {
+    /// Allocate hypervisor pages and map them read-only in U-mode VA space.
+    fn map(sv48: &FirstStagePageTable<Sv48>, hyp_mem: &mut HypPageAlloc) -> Self {
+        // Allocate pages.
+        let num_pages = PageSize::num_4k_pages(UMODE_INPUT_SIZE);
+        let pages = hyp_mem.take_pages_for_hyp_state(num_pages as usize);
+        let start = pages.base();
+
+        // Map the allocated pages as read-only in U-mode at UMODE_INPUT_START.
+        // UMODE_INPUT_START is 4k aligned (enforced by static assertion), round_down wil be a no-op.
+        let vaddr = PageAddr::with_round_down(
+            RawAddr::supervisor_virt(UMODE_INPUT_START),
+            PageSize::Size4k,
+        );
+        let pte_fields = PteFieldBits::leaf_with_perms(PteLeafPerms::UR);
+        // Unwrap okay: this is called once when we are creating the page-table. Guaranteed to be
+        // unmapped.
+        let mapper = sv48
+            .map_range(vaddr, PageSize::Size4k, num_pages, &mut || {
+                hyp_mem.take_pages_for_hyp_state(1).into_iter().next()
+            })
+            .unwrap();
+        for (virt, phys) in vaddr
+            .iter_from()
+            .zip(pages.base().iter_from())
+            .take(num_pages as usize)
+        {
+            // Safety: These pages are mapped read-only in the VA area reserved for the U-mode
+            // input region mappings. Hypervisor will write to these pages using the physical
+            // mappings and U-mode will read them through this mapping. Safe because these are
+            // per-CPU mappings, and when the hypervisor will be writing to these pages via the
+            // physical mappings no CPU will be able to access these pages through the U-mode
+            // mappings.
+            unsafe {
+                mapper.map_addr(virt, phys, pte_fields).unwrap();
+            }
+        }
+        // Safety: the range `(start..max_addr)` is mapped in U-mode as read-only and was uniquely
+        // claimed for this area from the hypervisor map above.
+        let vslice = unsafe {
+            VolatileSlice::from_raw_parts(start.raw().bits() as *mut u8, UMODE_INPUT_SIZE as usize)
+        };
+        Self { vslice }
+    }
+
+    // Writes `data` in the current CPU's U-mode Input Region.
+    fn store<T: DataInit>(&mut self, data: T) -> Result<(), Error> {
+        let vref = self.vslice.get_ref(0).map_err(Error::UmodeInput)?;
+        vref.store(data);
+        Ok(())
+    }
+
+    // Clears the U-mode Input Region.
+    fn clear(&mut self) {
+        self.vslice.write_bytes(0);
+    }
+}
+
 /// Mapping permission for a U-mode mapping slot.
 pub enum UmodeSlotPerm {
     Readonly,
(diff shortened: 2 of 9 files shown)


src/main.rs
```diff
@@ -329,17 +335,22 @@ extern "C" fn kernel_init(_hart_id: u64, _fdt_addr: u64) {
     poweroff();
 }
 
+#[cfg(test)]
+#[no_mangle]
+extern "C" fn primary_launch() {}
```
rbradford, line 340: Would you consider using the naming `{primary,secondary}_stage{1,2}` as then you don't need to remember if init comes before launch?

src/hyp_map.rs
```diff
@@ -35,6 +35,9 @@ type UmodeElfRegionsVec = ArrayVec<UmodeElfRegion, MAX_UMODE_ELF_REGIONS>;
 // Hw map regions vector.
 type HwMapRegionsVec = ArrayVec<HwMapRegion, MAX_HW_MAP_REGIONS>;
 
+// Global reference to the Hypervisor Map.
```
rbradford, line 38: nit: Perhaps this commit message should start with `hyp_map`. Also would you consider adding why moving it is a good idea?

src/main.rs
glg-rv: Not having strong opinion on  names, but believing that launch happens before init is a bit of a stretch.

Thought initially about stage0,1 but decided to do this essentially for fishing for names.

src/smp.rs
```diff
@@ -41,23 +41,14 @@ static PER_CPU_BASE: Once<SupervisorPageAddr> = Once::new();
 impl PerCpu {
     /// Initializes the `PerCpu` structures for each CPU, taking memory from `mem_map`. This (the
     /// boot CPU's) per-CPU area is initialized and loaded into TP as well.
-    pub fn init(boot_hart_id: u64, mem_map: &mut HwMemMap) {
+    pub fn init(boot_hart_id: u64, hyp_mem: &mut HypPageAlloc) {
```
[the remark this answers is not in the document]
glg-rv: Literally the title of the commit title is: "with the allocator", rather than the hw map. This has been a slowly moving process (that started on per-cpu page table) to move allocation from map to allocator. I think it's clear.

src/hyp_map.rs
glg-rv: I don't think the move is per se important in rust. It was just a clean up on the way code has evolved since the creation.

src/main.rs
glg-rv: current thinking (offline discussion with @rbradford) is to go towards `_init` and `_main`, almost crt-like. Open to change.

src/hyp_map.rs
```diff
@@ -585,7 +585,11 @@ impl HypMap {
     }
 
     /// Creates a new page table based on this memory map.
-    pub fn new_page_table(&self, hyp_mem: &mut HypPageAlloc) -> HypPageTable {
+    pub fn new_page_table(
```
rbradford, line 588: nit: This commit message needs s/launch/main/ now that that nomenclature is used.

src/smp.rs
```diff
@@ -41,23 +41,14 @@ static PER_CPU_BASE: Once<SupervisorPageAddr> = Once::new();
...
-            .regions()
-            .find(|r| r.region_type() == HwMemRegionType::Available && r.size() >= total_size)
-            .map(|r| r.base())
-            .expect("Not enough free memory for per-CPU area");
-        mem_map
-            .reserve_region(
-                HwReservedMemType::HypervisorPerCpu,
-                RawAddr::from(pcpu_base),
-                total_size,
-            )
```
abrestic-rivos, line 59: i guess we can remove the per-CPU region type from `HwMemMap` now.

src/hyp_layout.rs
```diff
@@ -21,7 +21,18 @@
...
 // +-------------------------+ UMODE_INPUT_START
 // | Umode Input Area        |
-// +-------------------------+ +UMODE_INPUT_SIZE
+// +-------------------------+ +UMODE_INPUT_SIZE (UMODE_INPUT_END, UMODE_END)
+// | (unused 4Mb)            |
+// +-------------------------+ HYP_STACKBTM_MIN
+// | (unused)                |
+// +-------------------------+ HYP_STACKTOP - stack size [1]
+// | Hypervisor Stack        |
+// +-------------------------+ HYP_STACK_TOP (0xffff_ffff_ffff_f000)
```
abrestic-rivos, line 30: nit: inconsistency between `STACKTOP` and `STACK_TOP`. i'd prefer the latter.

src/hyp_layout.rs
```diff
@@ -75,6 +86,17 @@ const_assert!(PageSize::Size4k.is_aligned(UMODE_INPUT_START));
 /// Size of the U-mode Input Region.
 pub const UMODE_INPUT_SIZE: u64 = 4 * 1024;
 const_assert!(PageSize::Size4k.is_aligned(UMODE_INPUT_SIZE));
+/// End of the U-mode Input Region.
+const UMODE_INPUT_END: u64 = UMODE_INPUT_START + UMODE_INPUT_SIZE;
+/// Last U-mode address.
+const UMODE_END: u64 = UMODE_INPUT_END;
+
+/// The lowest possible kernel stack address.
+pub const HYP_STACKBTM_MIN: u64 = UMODE_END + 4 * 1024 * 1024;
```
abrestic-rivos, line 95: is the difference between "bottom min" and `top - stack size" because cpu0 has a different stack size?

src/hyp_layout.rs
```diff
@@ -75,6 +86,17 @@ const_assert!(PageSize::Size4k.is_aligned(UMODE_INPUT_START));
...
+/// Address of the hypervisor stack top.
+pub const HYP_STACKTOP: u64 = 0xffff_ffff_ffff_f000;
```
abrestic-rivos, line 97: i'd make the stack alignment something greater than 4kb. say 16kb, or 2MB, in order to be able to take advantage of Svnapot or (unlikely) huge PTEs.

src/hyp_map.rs
```diff
@@ -323,6 +325,63 @@ impl UmodeInputRegion {
...
+        let pte_fields = PteFieldBits::leaf_with_perms(PteLeafPerms::RW);
+        // Unmap stack pages from the 1:1 map.
+        for _ in sv48
+            .unmap_range(
+                self.paddr.as_supervisor_virt(),
+                PageSize::Size4k,
+                self.page_count,
+            )
+            .unwrap()
+        {}
```
abrestic-rivos, line 363: instead of an empty loop, you can use `count()` to consume the iterator.

src/hyp_map.rs
```diff
@@ -323,6 +325,63 @@ impl UmodeInputRegion {
     }
 }
 
+/// Represents a hypervisor stack region.
+pub struct HypStackRegion {
+    vaddr: PageAddr<SupervisorVirt>,
+    paddr: PageAddr<SupervisorPhys>,
+    page_count: u64,
```
abrestic-rivos, line 332: once all stacks are the same size, we should be able to avoid storing vaddr & page_count for each CPU rigth?

src/main.rs
```diff
@@ -559,20 +559,22 @@ extern "C" fn primary_init(hart_id: u64, fdt_addr: u64) -> CpuLaunch {
...
+        sp: this_cpu.stack_top(),
     }
 }
 
 /// Start salus on the bootstrap CPU. This is called once we switch to the hypervisor page-table.
 #[no_mangle]
 extern "C" fn primary_main() {
+    let sp: u64;
+    // Safe since there is no memory access.
+    unsafe { core::arch::asm!("mv {rs}, sp", rs = out(reg) sp) };
```
abrestic-rivos, line 575: the asm shuld have "nomem" to indicate the absence of a memory clobber then. (here and below)

src/smp.rs
```diff
@@ -18,6 +20,8 @@ use crate::vm_id::VmIdTracker;
 
 // The secondary CPU entry point, defined in start.S.
```
abrestic-rivos, line 21: move this comment to be over `_secondary_start` because it doesn't really apply to `_stack_*`

src/smp.rs
```diff
@@ -30,10 +34,11 @@ pub struct PerCpu {
     page_table: HypPageTable,
     umode_task: Once<RefCell<UmodeTask>>,
     online: Once<bool>,
+    stack_top: u64,
 }
 
 /// The number of pages we allocate per CPU: the CPU's stack + it's `PerCpu` structure.
-const PER_CPU_PAGES: u64 = 8;
+const PER_CPU_PAGES: u64 = 0x80;
```
abrestic-rivos, line 41: 512k per CPU?!

src/smp.rs
```diff
@@ -54,15 +62,45 @@ impl PerCpu {
...
+        let mut pcpu_stacks =
+            pcpu_pages.into_chunks_iter(core::num::NonZeroU64::new(PER_CPU_PAGES).unwrap());
         for i in 0..cpu_info.num_cpus() {
             let cpu_id = CpuId::new(i);
+            // Unwrap okay. We allocated the area `with num_cpus() * PER_CPU_PAGES` pages.
+            let pcpu_stack = pcpu_stacks.next().unwrap();
+            // Boot CPU is special. Doesn't use the PCPU_BASE area as stack, only for the PCPU area.
+            // TODO: Do not allocate PCPU area for boot cpu.
+            let stack_pages = if cpu_id == boot_cpu {
+                Self::boot_cpu_stack()
```
abrestic-rivos, line 75: i think we discussed this offline and i'm fuzzy on the details, but why can't we switch tthe boot CPU over to use a stack from `pcpu_stacks`? the whole boot cpu having a different stack size than the secondary cpus thing has hidden a couple of bugs from us so far. (it would also reduce all the special-casing you have to do here.)

src/smp.rs
```diff
@@ -54,15 +62,45 @@ impl PerCpu {
...
+                        pcpu_stack.len(),
+                    )
+                    .unwrap()
+                }
+            };
+            let stack_top = if cpu_id == boot_cpu {
+                crate::hyp_layout::HYP_STACKTOP
+            } else {
+                // Secondary CPUs have PerCpu structure at top of stack.
+                crate::hyp_layout::HYP_STACKTOP - core::mem::size_of::<PerCpu>() as u64
```
abrestic-rivos, line 94: so by keeping `PerCpu` on the stack, we effectively make it impossible for one CPU to access another's `PerCpu`. we don't use it for inter-processor communication currently, but it's not unreasonable that we would want to do so in the future. have you considered putting the `PerCpu`s in a global array instead? i only put them on the stack to begin with because it was convenient to be able to have one pointer for both the initial sp and `PerCpu` when launching a secondary CPU.

src/smp.rs
glg-rv: Agreed.

src/hyp_layout.rs
glg-rv: Whops.

src/hyp_layout.rs
glg-rv: Yes, or rather, was. I want to fix that in the next PR, where I force the stack to be the same on all CPUs (to ease calculation on trap handler for stack bottom). This was just a check on the bootstrap CPU stack (that is controlled at the moment by the linker script)

src/hyp_map.rs
glg-rv: Ack.

src/main.rs
glg-rv: Sure.

src/smp.rs
glg-rv: Right.

src/smp.rs
glg-rv: Yes. Our stack usage isn't great. Mapping takes quite a lot. And when we run in Debug Mode we're essentially all over the place when it come to stack usage.

I am due to restore my previous stack analysis once this is over.

src/smp.rs
glg-rv: Sure, the problem with this is that we would still need a (huge) stack at the beginning (init phase). That stack is allocated by the linker.

I implemented both ways, one where the bootstrap CPU used this PerCPu stack (which simplifies things quite a lot on a architectural point of view) but that wastes the original init-time stack memory.

With this, we can have a following PR that saves stack memory by changing this `smp` code to not allocate a full blown CPU stack just for the PerCpu data stucture. There are various ways to achieve this, non particularly difficult.

src/smp.rs
glg-rv: So the trick here is that only the CPU using the stack doesn't have its stack mapped in the 1:1. Which is why the CPU bootstrap can wait for the CPU to go online in `start_secondary_cpus` -- which effectively requires access of other CPU's PerCpu from the bootstrap CPU.

My comment above (about removing the need for a full stack for the Bootstrap PerCpu Data structure) implies exactly what you suggest. I had implemented it, but I wanted this patch to be as simple as possible and use the current SMP architecture to review the mechanism of stack allocation.

src/smp.rs
glg-rv: Done.

src/hyp_layout.rs
glg-rv: Fixed.

src/hyp_layout.rs
glg-rv: All CPUs have the same size in the new iteration.

src/hyp_layout.rs
glg-rv: Aligned at 2M.

src/smp.rs
glg-rv: Now PerCpu data and stack are separate.

src/smp.rs
glg-rv: PerCpu not on stack anymore.

src/smp.rs
glg-rv: Done.

glg-rv:
Relatively major update. Originally I thought of this patch as something that wouldn't change `smp.rs` fundamentally in order to focus on the stack mapping size. Since the issues that this created came up in the review, decided to make this PR a bit more complex.

This new version of the PR forces all CPUs to have the same stack size. This makes the layout fixed and also simplifies the code.

PerCpu data doesn't live in the stack anymore. This slightly changes the AP bootstrap code because SP and TP are not the same anymore. As an advantage, we do not allocate an extra stack just for the bootstrap CPU PerCpu.

src/hyp_map.rs
```diff
@@ -254,6 +257,72 @@ impl UmodeElfRegion {
...
 }
 
+/// The U-mode Input Region is a private region of the page table used by the hypervisor to pass
+/// data to the current U-mode operation. This avoids mapping hypervisor data directly in
+/// U-mode. This area is written by the hypervisor and mapped read-only in U-mode.
+struct UmodeInputRegion {
+    /// Volatile Slice used to write to this area from the hypervisor.
+    vslice: VolatileSlice<'static>,
+}
+
```
rbradford, line 268: nit: If you don't need any other members then you can use the alternative `struct UmodeInputRegion(VolatileSlice<'static>)` syntax and then refer to it as `.0`

src/hyp_map.rs
```diff
@@ -254,6 +257,72 @@ impl UmodeElfRegion {
...
+        let vslice = unsafe {
+            VolatileSlice::from_raw_parts(start.raw().bits() as *mut u8, UMODE_INPUT_SIZE as usize)
+        };
+        Self { vslice }
+    }
+
+    // Writes `data` in the current CPU's U-mode Input Region.
+    fn store<T: DataInit>(&mut self, data: T) -> Result<(), Error> {
+        let vref = self.vslice.get_ref(0).map_err(Error::UmodeInput)?;
+        vref.store(data);
```
rbradford, line 317: nit: Is it a good idea to zero the area before writing in case the new data is smaller than the old?

src/hyp_map.rs
rbradford: I just realised that this a rearrangment commit so you can disregard these comments as this is not new code (or consider fixing in another commit if you think they're appropriate :-) )

src/smp.rs
```diff
@@ -30,11 +35,9 @@ pub struct PerCpu {
     page_table: HypPageTable,
     umode_task: Once<RefCell<UmodeTask>>,
     online: Once<bool>,
+    stack_top_addr: SupervisorPageAddr,
```
rbradford, line 38: Could you store `HypStackRegion` here instead and add to it accessors for start/end of stack as well as getting it as a slice. @stillson could then use that in the backtrace PR.

src/smp.rs
```diff
@@ -43,54 +46,119 @@ impl PerCpu {
...
+    }
 
-        let me = Self::this_cpu();
-        me.set_online();
+    fn boot_cpu_stack() -> SequentialPages<InternalDirty> {
+        // Get the pages of the current stack as created by the linker.
+        // Safe because these are linker created variables.
+        let stack_start = unsafe { core::ptr::addr_of!(_stack_start) as u64 };
+        let stack_end = unsafe { core::ptr::addr_of!(_stack_end) as u64 };
+        let stack_startaddr = PageAddr::new(RawAddr::supervisor(stack_start))
```
rbradford, line 136: In cases like this I think variable shadowing is fine it would improve the readability.

src/smp.rs
glg-rv: Well, that patch will have to use the virtual mapping not the physical one. Virtual mapping are static, which is one of the feature of the patch. I am not particularly happy about adding this field, but I needed between creation and initialization of CPUs, so that was good enough.

Regarding HypStackRegion, no, if we need the hypstack region we should access it from the HypMap, not here.

src/smp.rs
glg-rv: Readability is subjective. :-)

I find it more readable to step things in small steps. But what you're suggesting is more in line with the rust way so I'll happily change that.

src/hyp_map.rs
glg-rv: The guarantee that this is cleared is due on u-mode execute request. I understand that this might not sound straightforward, but this piece of code (as you correctly noted, this is just a move) was a complex interaction between three systems, so I preferred to have the mechanism here and the checking elsewhere.

In case you are going to go with a type that destroy on drop, that was one of my iterations during review. :-)

src/hyp_map.rs
glg-rv: Heh, yes. In general I don't mind named structs. I find them simpler. Do you have any opinion on why would be better to have in the format you suggest?

rbradford: I think it's just a style/convention thing. If the struct is really a wrapper around another type (so you can implement methods or traits) then I think the tuple format is preferred. If it's truly a new struct that only happens to have one member then named member structs are the norm.

src/smp.rs
abrestic-rivos: understood re: not wanting to waste memory. but should we at least clamp the size of the boot CPU's stack when we set up the VA mapping so that all CPUs have the same stack size? i really want to get rid of this situation where things sometimes work because they're run on the boot CPU vs secondary CPU.

abrestic-rivos (commented):
just a couple minor comments left, otherwise lgtm.

src/smp.rs
```diff
@@ -43,54 +46,119 @@ impl PerCpu {
...
+        // Allocate memory for per-CPU structures.
+        let pcpu_size = core::mem::size_of::<PerCpu>() as u64 * cpu_info.num_cpus() as u64;
         let pcpu_pages =
-            hyp_mem.take_pages_for_hyp_state(PageSize::num_4k_pages(total_size) as usize);
+            hyp_mem.take_pages_for_hyp_state(PageSize::num_4k_pages(pcpu_size) as usize);
         let pcpu_base = pcpu_pages.base();
         PER_CPU_BASE.call_once(|| pcpu_base);
 
+        // Allocate memory for the secondary CPUs stack.
+        let stacks_page_count = HYP_STACK_PAGES * cpu_info.num_cpus() as u64 - 1;
```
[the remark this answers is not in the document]
glg-rv: Whops! This hurts. Thank you, will fix. :-)

src/smp.rs
glg-rv: I do that in HypMap on stack region. I return InvalidStackSize if the linker allocated stack is not _exactly_ `HYP_STACK_PAGES`. But thinking about it, possibly clamping it afterwards could be more generic and allows us a bigger boot stack than runtime requires. Would be wasted memory though.

Although, I kind of like to remove any power to the linker and check at runtime if it was built properly.

src/hyp_map.rs
glg-rv: I think I'll leave it like this for now. In a way there's more to the UmodeInputRegion than a slice, although that's what it is in practice.

src/smp.rs
glg-rv: Changed. Put unsafe across the pageaddr block. Let me know what you think.

src/hyp_layout.rs
```diff
@@ -76,6 +82,25 @@ const_assert!(PageSize::Size4k.is_aligned(UMODE_INPUT_START));
 pub const UMODE_INPUT_SIZE: u64 = 4 * 1024;
 const_assert!(PageSize::Size4k.is_aligned(UMODE_INPUT_SIZE));
 
+/// Address of the hypervisor stack top.
+#[allow(dead_code)]
```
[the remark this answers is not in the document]
glg-rv: I'll leave it as is for now, but will do that in future PRs.

src/smp.rs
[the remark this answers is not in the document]
glg-rv: Fixed.

abrestic-rivos (approved):

Issue #292, opened 2023-03-23
build: Check every commit can be built


sboeuf:
It'd be nice if we could have our CI verifying that each and every commit from a PR can be properly built. The main reason for that is to be able to run a `git bisect` if later in the future we're trying to understand where something might have gone wrong.

glg-rv:
Not sure how I'd feel about that. At this moment the reason at least I use to separate in different commit is to ease the reviewing. I am okay with getting it to a build binary, but I wouldn't like the full blown lint + format.
This is especially true when adding new features, where the introduction of those feature would temporarily have dead code.

sboeuf:
> Not sure how I'd feel about that. At this moment the reason at least I use to separate in different commit is to ease the reviewing. I am okay with getting it to a build binary, but I wouldn't like the full blown lint + format. This is especially true when adding new features, where the introduction of those feature would temporarily have dead code.

Per my experience, being able to bisect is a great benefit. I'm not at all suggesting we should not separate commits, this is quite important to keep patches as small as possible to give the reviewers an easy way to go through code reviews.
Let's say you update a function with a new parameter, you would need to update the caller to make sure it provides the new parameter, that's for sure, but you can defer to another commit the actual change that will leverage this new parameter.
On the dead code side of things, it's usually good to introduce unused code with `#[allow(dead_code)]` in the first commit introducing some new functions, and then remove this label when the following commit starts using these newly defined functions.

glg-rv:
Any bisect script I have used would check if the commit builds, if not go up or down, depending on direction. Especially in a system that's based on CI on PRs you know that there's a commit that works and that is where you should bisect anyway, because it's the PR that possibly introduced the bug as whole, not a single commit.

Forcing every single commit to be both lint+format I feel would be an unnecessary slow down on development efforts, and complicating a developer life when you have to fix up patches before sending, while doing the pre-PR work.

I repeat, I am fine with requiring every single commit to build. But lint+format for each of them is just a big (am paraphrasing) "I hate you" to a programmer.

mnissler-rivos:
+1 to requiring that every commit builds and meets the quality bar we've set on all other dimensions as well (including lint+format, making lint+format changes subsequently ties commits together unnecessarily and effectively means you can only test/revert at pull request level).

That said, if we wanted to have smaller units for review, then squashing before submit could be an option as well (where it makes sense, we certainly don't want to bunch together unrelated/independent changes).

glg-rv:
[...]

PR #293, opened 2023-03-23
.github: Extend workflow to build check all commits

rbradford:
See: #292

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -23,6 +23,7 @@ jobs:
     - uses: actions/checkout@v3
       with:
         submodules: recursive
+        fetch-depth: 0
     - name: Build
       run: bazel build //:salus-all
     - name: Lint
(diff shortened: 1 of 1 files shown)


rbradford:
@sboeuf This is adapted from the version from Cloud Hypervisor but it seems it's not working here :-/

sboeuf:
> @sboeuf This is adapted from the version from Cloud Hypervisor but it seems it's not working here :-/

I think this is because you're missing `fetch-depth: 0` from the way the checkout is performed:
```
- name: Code checkout
        uses: actions/checkout@v3
        with:
          fetch-depth: 0
```
Github documentation says `Set fetch-depth: 0 to fetch all history for all branches and tags.`. That should solve the error `fatal: ambiguous argument 'origin/main..fb189d8df9f22da39da9d597e9864fc24f7f4eff': unknown revision or path not in the working tree.` by making the runner aware of your branch and the associated commits.

sboeuf (approved):

dgreid (approved):

PR #294, opened 2023-03-27
vm: TvmAddMeasuredPages requires non-confidential source memory

rbradford:
Check that the source memory for the TvmAddMeasuredPages call is marked
as non confidential (Shared or Mapped)

### DIFF
--- a/src/vm.rs
+++ b/src/vm.rs
@@ -1522,6 +1522,12 @@ impl<'a, T: GuestStagePagingMode> FinalizedVm<'a, T> {
         // Make sure we can initialize the full set of pages before we start actually inserting
         // them into the destination page table.
         let src_page_addr = self.guest_addr_from_raw(src_addr)?;
+        // Check source pages are not confidential and owned by this VM
+        let _ = self
+            .vm_pages()
+            .get_shareable_pages(src_page_addr, num_pages)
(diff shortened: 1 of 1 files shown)


rbradford:
As far as I could tell this check was not being made. Further not 100% sure this is the best way to check so very much open to alternative suggestions.

src/vm.rs
```diff
@@ -1522,6 +1522,12 @@ impl<'a, T: GuestStagePagingMode> FinalizedVm<'a, T> {
...
         let src_page_addr = self.guest_addr_from_raw(src_addr)?;
+        // Check source pages are not confidential and owned by this VM
+        let _ = self
+            .vm_pages()
+            .get_shareable_pages(src_page_addr, num_pages)
+            .map_err(EcallError::from)?;
+
         let mut initialized_pages = LockedPageList::new(self.page_tracker());
         for (page, addr) in pages.zip(src_page_addr.iter_from()) {
             match page.try_initialize(|bytes| active_pages.copy_from_guest(bytes, addr.into())) {
```
abrestic-rivos, line 1533: `copy_from_guest()` will fail on pages that the host has converted to confidential since they're unmapped in the host. Is there another threat vector this is trying to address?

PR #295, opened 2023-03-28
Vm: Add support to probe Nacl features.

rajnesh-kanwal:
In the recent updates to Nacl extension [0], feature probing was introduced and allows guest hypervisor to probe which features are supported by the host hypervisor.

Also, the SetShmem FID argument has changed from pfn to full address of the shared memory page.

[0] https://lists.riscv.org/g/sig-hypervisors/message/256


rajnesh-kanwal:
> probably need to update `tellus` too?

`tellus` uses the `register_shmem` function in sbi-rs which takes in `shmem_ptr` and converts it to pfn, with `register_shmem` updated `tellus` doesn't need any changes.

I can add a test for `ProbeFeature` if that's what you are suggesting.

abrestic-rivos:
Ah, I was looking at the CI failure (which is because Atish's PR isn't in yet).

atishp04:
I am thinking to add CoVE related features which can be probed as well. 

1. SBI_NACL_FEAT_COVE_GPR
 -- This will describe the the scratch layout for GPRs

2. SBI_NACL_FEAT_COVE_CSR

- Describes the limited CSR shared through NACL compared to all CSR shared through in normal nested acceleration.

This will align both specs more cleanly and clearly describe where CoVE supersedes NACL description.
It will also allow other TSM/host hypervisor implementations to not use the NACL features if they wish to.
It allows extensibility for any possible future use case as well.

src/vm.rs
```diff
@@ -1007,14 +1007,19 @@ impl<'a, T: GuestStagePagingMode> FinalizedVm<'a, T> {
...
-        if shmem_pfn != u64::MAX {
+    fn set_shmem_area(
+        &self,
+        shmem_addr: u64,
+        active_vcpu: &mut ActiveVmCpu<T>,
+    ) -> EcallResult<u64> {
+        if shmem_addr != u64::MAX {
             // Pin the pages that the VM wants to use for the shared state buffer.
-            let shared_page_addr = self.guest_addr_from_raw(shmem_pfn << PFN_SHIFT)?;
+            let shared_page_addr = self.guest_addr_from_raw(shmem_addr)?;
```
atishp04, line 1022: Add a sanity check for the 4K alignment now ?

rajnesh-kanwal: `guest_addr_from_raw` is doing the alignment check.

abrestic-rivos (approved):

abrestic-rivos (approved):

atishp04:
> > I am thinking to add CoVE related features which can be probed as well.
> 
> Doesn't probe just say if those features are present? i.e. if we want to know whether or not the shmem scratch area has the COVE layout, don't we need just one "feature" that indicates we have COVE?

One will work too. I was suggesting two that aligns better in terms of current spec alignment (scratch area/ CSR area).
We can describe the scratch & CSR access exception under the same feature.

PR #296, opened 2023-03-28
Check for stack overflow in trap handler

glg-rv:
This change adds support for having a trap handler (that panics) in case the kernel stack is exhausted.

RISC-V lacking double fault, we check the stack pointer at the entry of the trap handler. When this happens, we switch to an emergency frame. Since this is a panic frame, we have only one for all CPUs, and we use a simple spinlock to allocate it in case two CPUs run out of stack at the same time.

I couldn't find a way to not clobber one register of the frame, and having to choose I went for T1. I don't think this is important, the most important data we get from such a failure is: IP and stack pointer.

### DIFF
--- a/src/trap.S
+++ b/src/trap.S
@@ -2,17 +2,15 @@
 // Licensed under the Apache License, Version 2.0, see LICENSE for details.
 // SPDX-License-Identifier: Apache-2.0
 
-/// Trap handler entry point.
+.attribute arch, "rv64gcv"
 .section .text.init
-.global _trap_entry
-.balign 4
-_trap_entry:
-    /*
-     * Store the trap frame directly on the current stack. We don't have HU support yet so we
-     * should already be in HS. Traps taken from VS/VU are handled separately by _guest_exit.
-     * Trashes sscratch.
-     */
-    csrw sscratch, sp
+
+/*
+ * Store the trap frame directly on the current stack.
+ * `sscratch` contains the SP to be saved in the frame.
+ * `sscratch` is not preserved.
+ */
+.macro save_frame
     addi sp, sp, -({tf_size})
     sd   ra, ({tf_ra})(sp)
     sd   gp, ({tf_gp})(sp)
@@ -51,6 +49,31 @@ _trap_entry:
     sd   t1, ({tf_sstatus})(sp)
     csrr t2, sepc
     sd   t2, ({tf_sepc})(sp)
+.endm
+
+// Early trap handler. Doesn't check for stack overflow as hypervisor mappings are not enabled yet.
+.global _trap_init_entry
+.balign 4
+_trap_init_entry:
+    /* Skip the buffer overflow check at init time. */
+    j .L_same_stack_trap
+
+/// Trap handler entry point.
+.global _trap_entry
+.balign 4
+_trap_entry:
+    /* Save T0 in sscratch */
+    csrw sscratch, t0
+    /* Check for stack overflow. If it is, switch to overflow stack and panic _quickly_. */
+    li t0, {hyp_stack_bottom}
+    blt sp, t0, _stack_overflow
+    /* Restores T0 from sscratch */
+    csrr t0, sscratch
+
+.L_same_stack_trap:
+    /* Save frame on current stack. */
+    csrw sscratch, sp
+    save_frame
 
     /* Now enter the rust trap handler. */
     move a0, sp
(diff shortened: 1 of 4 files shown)


src/trap.S
```diff
@@ -94,3 +110,68 @@ _trap_entry:
...
+    csrw sscratch, zero
+    csrr t1, sstatus
+    sd   t1, ({tf_sstatus})(sp)
+    csrr t2, sepc
+    sd   t2, ({tf_sepc})(sp)
+
+    /* Now enter the rust trap handler. */
+    move a0, sp
+    call handle_stack_overflow
+
```
rbradford, line 171: You could do an XOR swap to preserve the value in t1. :-)

x = a xor c
y = b xor c 

You can then calculate:

c = x xor y
a = x xor c
b = y xor c

src/trap.S
```diff
@@ -94,3 +110,68 @@ _trap_entry:
...
+.data
+_emergency_stack_lock:
+.long 0
+_emergency_stack_bottom:
+ .space 64*1024
+_emergency_stack_top:
```
abrestic-rivos, line 177: can we declare these in rust? and do we actually need 16k here?

src/trap.S
```diff
@@ -94,3 +110,68 @@ _trap_entry:
...
+    sd   s9, ({tf_s9})(sp)
+    sd   s10, ({tf_s10})(sp)
+    sd   s11, ({tf_s11})(sp)
+    sd   t0, ({tf_t0})(sp)
+    sd   t1, ({tf_t1})(sp)
+    sd   t2, ({tf_t2})(sp)
+    sd   t3, ({tf_t3})(sp)
+    sd   t4, ({tf_t4})(sp)
+    sd   t5, ({tf_t5})(sp)
+    sd   t6, ({tf_t6})(sp)
```
abrestic-rivos, line 159: can we avoid duplicating stack-frame-saving between this and the normal trap flow?

src/trap.S
```diff
@@ -2,11 +2,27 @@
...
-/// Trap handler entry point.
+.attribute arch, "rv64gcv"
 .section .text.init
+
+// Early trap handler. Doesn't check for stack overflow as hypervisor mappings are not enabled yet.
+.global _trap_init_entry
+.balign 4
+_trap_init_entry:
+        /* Skip the buffer overflow check at init time. */
+        j .L_trap_handler
```
abrestic-rivos, line 13: nit: inconsistent indentation between this block and everything below.

abrestic-rivos (commented):
thanks!

src/trap.S
rbradford: I was mistaken - I don't think you can do it like this. 😳

src/trap.S
glg-rv: Ah yes, sorry. Somehow the indentation in my emacs configuration is really incompatible with the way asm is formatted here, so it's all manual. Will fix.

src/trap.S
glg-rv: Any suggest on what would be the rust way? I don't think I can use macros here.

src/trap.S
glg-rv: Yes, will do. No, I am just being paranoid in stack usage, will reduce.

src/trap.S
glg-rv: Yeah I need a place to maintain data anyway, and I guessed that loosing a register is good enough for a end-of-world scenario.

src/trap.S
abrestic-rivos: oh, i'd just suggest jumping to a common bit of asm that sets up the trap frame & calls the trap handler

src/trap.S
```diff
@@ -2,11 +2,27 @@
...
+
+/// Trap handler entry point.
 .global _trap_entry
 .balign 4
 _trap_entry:
+    /* Check for stack overflow. If it is, switch to emergency stack and panic _quickly_. */
+     csrw sscratch, t0
+     li t0, {hyp_stack_bottom}
+     blt sp, t0, _stack_overflow
```
atulkharerivos, line 22: Is `{hyp_stack_bottom}`  the actual bottom of stack + one trap-frame worth of space? Otherwise, think we need to ensure that the blt on 22 accounts for it, since can potentially encounter a fault in the stores below if we just check for bottom of stack.

src/trap.S
glguida: Hm, I thought I couldn't because I had no registers to save whether this was a stack overflow (and hence panic straight away) or a simple exception. But thinking about it I could add a field in the stack frame that tells me what it is.
Although it's going to be a pagefault in the kernel, so it would fault anyway even if we enter the normal trap handler (but I'd still like a way to differentiate it).

src/trap.S
glguida: If it does, it will be a stack overflow, so will fault and work as it should. :-)

atulkharerivos: Got it...makes sense, but it'll just be a double-fault, and since we'll fail the check the second time around, we are OK, since we'll switch to the emergency stack.

src/trap.S
glg-rv: (sorry, comment above was from personal account).

Turns out that it gets really complicated having a single entry in rust, and saving a single 1/0 in the frame to check whether it's overflow or not sould still require an extra register.

Also turns out that I *can* use macros. So I created a parameter-less macro to save the frame in the current stack. Avoiding duplication of stack saving.

src/trap.S
glg-rv: I could declare stack_lock in rust. Regarding the overflow stack itself, it was a bit hairy to get the pointer to the top of the stack. I went for definition of the overflow stack in linker script.

abrestic-rivos (approved):

PR #301, opened 2023-03-29
ci: Run clippy with warnings treated as errors

sboeuf:
We don't want clippy warnings to be merged into the main branch, which is why we use -Dwarnings clippy flag to indicate that all warnings should be treated as errors. This will indicate the bazel build failed and prevent any pull request with such warnings from being merged.


abrestic-rivos (approved):

glg-rv:
LGTM.

stillson (approved):
Ship it

PR #303, opened 2023-03-30
Improve boot-time

glg-rv:
We are currently walking TWO times the page tables for each PTE we map.

One when we lock, the other when we map. This is the main reason why our paging code is so slow.

This patch  addresses both issues for the hypervisor mapping case.

This is just the first step to make our paging code faster. On my machine this patch reduces the boot of salus (real time) from 24 seconds to 7 seconds.

### DIFF
--- a/riscv-page-tables/src/page_table.rs
+++ b/riscv-page-tables/src/page_table.rs
@@ -1043,7 +1043,6 @@ impl<'a, T: FirstStagePagingMode> FirstStageMapper<'a, T> {
     /// # Safety
     ///
     /// Don't create aliases.
-    #[allow(dead_code)]
     pub unsafe fn map_contiguous(
         &self,
         start_vaddr: PageAddr<T::MappedAddressSpace>,
--- a/riscv-page-tables/src/page_table.rs
+++ b/riscv-page-tables/src/page_table.rs
@@ -693,34 +693,44 @@ impl<T: PagingMode> PageTableInner<T> {
         }
     }
 
-    /// Locks the invalid leaf PTE mapping `vaddr`, filling in any missing intermediate page tables
-    /// using `get_pte_page`.
-    fn lock_leaf_for_mapping(
+    /// Locks the invalid leaf PTEs mappings starting from `vaddr` for `num_pages` pages of size
+    /// `page_size`, filling in any missing intermediate page tables using `get_pte_page`.
+    fn lock_leaves_for_mapping(
         &mut self,
-        vaddr: PageAddr<T::MappedAddressSpace>,
+        start_addr: PageAddr<T::MappedAddressSpace>,
+        num_pages: u64,
         page_size: PageSize,
         get_pte_page: &mut dyn FnMut() -> Option<Page<InternalClean>>,
     ) -> Result<()> {
-        let mut table = PageTable::from_root(self);
-        // match the level to the page size
-        while table.level.leaf_page_size() != page_size {
-            table = table.next_level_or_fill_fn(RawAddr::from(vaddr), get_pte_page)?;
-        }
-        let entry = table.entry_for_addr_mut(RawAddr::from(vaddr));
-        use TableEntryType::*;
-        match entry {
-            Invalidated(i) => {
-                i.lock();
-                Ok(())
+        start_addr
+            .checked_add_pages_with_size(num_pages, page_size)
+            .ok_or(Error::AddressOverflow)?;
+
+        let iterator = start_addr
+            .iter_from_with_size(page_size)
+            .ok_or_else(|| Error::AddressMisaligned(start_addr.bits()))?;
+
+        for vaddr in iterator.take(num_pages as usize) {
+            let mut table = PageTable::from_root(self);
+            // match the level to the page size
+            while table.level.leaf_page_size() != page_size {
+                table = table.next_level_or_fill_fn(vaddr.raw(), get_pte_page)?;
             }
-            Unused(u) => {
-                u.lock();
-                Ok(())
+            let entry = table.entry_for_addr_mut(vaddr.raw());
+            use TableEntryType::*;
+            match entry {
+                Invalidated(i) => {
+                    i.lock();
+                }
+                Unused(u) => {
+                    u.lock();
+                }
+                LockedMapped(_) | LockedUnmapped(_) => return Err(Error::PteLocked),
+                Leaf(_) => return Err(Error::MappingExists),
+                Table(_) => return Err(Error::TableEntryNotLeaf),
             }
-            LockedMapped(_) | LockedUnmapped(_) => Err(Error::PteLocked),
-            Leaf(_) => Err(Error::MappingExists),
-            Table(_) => Err(Error::TableEntryNotLeaf),
         }
+        Ok(())
     }
 
     /// Locks an existing leaf PTE mapping of `vaddr` for remapping.
@@ -859,16 +869,9 @@ impl<T: FirstStagePagingMode> FirstStagePageTable<T> {
         if !addr.is_aligned(page_size) {
             return Err(Error::AddressMisaligned(addr.bits()));
         }
-        let iterator = addr
-            .iter_from_with_size(page_size)
-            .ok_or_else(|| Error::AddressMisaligned(addr.bits()))?;
-        addr.checked_add_pages_with_size(num_pages, page_size)
-            .ok_or(Error::OutOfMapRange)?;
-
-        let mut inner = self.inner.lock();
-        for a in iterator.take(num_pages as usize) {
-            inner.lock_leaf_for_mapping(a, page_size, get_pte_page)?;
-        }
+        self.inner
+            .lock()
+            .lock_leaves_for_mapping(addr, num_pages, page_size, get_pte_page)?;
 
         Ok(FirstStageMapper::new(self, addr, page_size, num_pages))
     }
@@ -1032,18 +1035,10 @@ impl<T: PagingMode> GuestStagePageTable<T> {
         num_pages: u64,
         get_pte_page: &mut dyn FnMut() -> Option<Page<InternalClean>>,
     ) -> Result<GuestStageMapper<T>> {
-        let mut mapper = GuestStageMapper::new(self, addr, 0);
-        let mut inner = self.inner.lock();
-        for a in addr
-            .iter_from_with_size(page_size)
-            .ok_or(Error::AddressMisaligned(addr.bits()))?
-            .take(num_pages as usize)
-        {
-            inner.lock_leaf_for_mapping(a, page_size, get_pte_page)?;
-            mapper.num_pages += 1;
-        }
-
-        Ok(mapper)
+        self.inner
+            .lock()
+            .lock_leaves_for_mapping(addr, num_pages, page_size, get_pte_page)?;
+        Ok(GuestStageMapper::new(self, addr, num_pages))
     }
 
     /// Prepares for remapping `num_pages` pages of size `page_size` starting at `addr` in the mapped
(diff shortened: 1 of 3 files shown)


rbradford (commented):
That is a great result. How did you measure it?

glg-rv:
> That is a great result. How did you measure it?

`time NCPU=4 QEMU=../qemu make run_tellus` at third run, to avoid bazel getting in the way. Not scientific but the difference is so big doesn't require it.

atishp04:
Awesome. Thanks for working on this. We have been carrying an out-of-tree qemu patch to reduce the PCI bar space in virt machine to 2GB to improve the boot time.
 
I see similar improvement on booting Linux while booting 2 cpus.

Without the PCI hack:
$time scripts/run_buildroot.sh
real	1m37.615s
user	2m48.265s
sys	0m0.522s

Without the PCI hack + this series:
real	0m36.997s
user	0m48.935s
sys	0m0.376s

With PCI hack, this series improved the boot time(real time) from ~35sec to ~17sec.

riscv-page-tables/src/page_table.rs
```diff
@@ -585,6 +600,72 @@ impl<T: PagingMode> PageTableInner<T> {
...
+                let entry = table.entry_for_index_mut(i);
+                use TableEntryType::*;
+                match entry {
+                    LockedUnmapped(l) => {
+                        if l.level().leaf_page_size() != page_size {
+                            return Err(Error::PageSizeMismatch(
+                                page_size,
+                                l.level().leaf_page_size(),
+                            ));
+                        }
```
abrestic-rivos, line 648: don't we know that the page size here matches based on walking the page tables in the loop above?

riscv-page-tables/src/page_table.rs
```diff
@@ -878,6 +959,43 @@ impl<'a, T: FirstStagePagingMode> FirstStageMapper<'a, T> {
...
     }
 
+    /// Maps `num_pages` pages contiguously both physically and virtually, starting from
+    /// `start_vaddr`, which maps to `start_paddr`. The caller must guarantee that the `num_pages`
+    /// pages starting from `start_paddr` are safe to map in this page table.
+    ///
+    /// # Safety
+    ///
+    /// Don't create aliases.
+    pub unsafe fn map_contiguous_addr(
```
abrestic-rivos, line 968: personal preference naming nit: i'd cal this one `map_contiguous` and the one below `map_one`

abrestic-rivos (approved):
very nice! i'd like to @sboeuf's huge page stuff in first as he's been waiting to merge that for a while, but after that let's get this in.

riscv-page-tables/src/page_table.rs
glg-rv: Yes, I was just copying what was in map_leaf because this code seems temporary anyway. If we support different page sizes then table should not be unreachable there. So I kind of wanted to understand what the status is, and in general not interfere with future works on huge pages.

riscv-page-tables/src/page_table.rs
glg-rv: Ack.

glg-rv: Done.

riscv-page-tables/src/page_table.rs
```diff
@@ -633,34 +633,44 @@ impl<T: PagingMode> PageTableInner<T> {
...
-                Ok(())
+        start_addr
+            .checked_add_pages_with_size(num_pages, page_size)
+            .ok_or(Error::OutOfMapRange)?;
+
+        let iterator = start_addr
+            .iter_from_with_size(page_size)
+            .ok_or_else(|| Error::AddressMisaligned(start_addr.bits()))?;
+
+        for vaddr in iterator.take(num_pages as usize) {
```
sboeuf: nit: don't need to create an `iterator` variable:
```rust
for vaddr in start_addr.iter_from_with_size(page_size).ok_or_else(|| Error::AddressMisaligned(start_addr.bits()))?.take(num_pages as usize) {
```

riscv-page-tables/src/page_table.rs
```diff
@@ -633,34 +633,44 @@ impl<T: PagingMode> PageTableInner<T> {
...
+            let mut table = PageTable::from_root(self);
+            // match the level to the page size
+            while table.level.leaf_page_size() != page_size {
+                table = table.next_level_or_fill_fn(RawAddr::from(vaddr), get_pte_page)?;
```
sboeuf, line 741: small nit:
```suggestion
                table = table.next_level_or_fill_fn(vaddr.into(), get_pte_page)?;
```

riscv-page-tables/src/page_table.rs
```diff
@@ -633,34 +633,44 @@ impl<T: PagingMode> PageTableInner<T> {
...
             }
-            Unused(u) => {
-                u.lock();
-                Ok(())
+            let entry = table.entry_for_addr_mut(RawAddr::from(vaddr));
```
sboeuf: Same here, I find it less heavy to read to use `vaddr.into()` rather than `RawAddr::from(vaddr)`.

riscv-page-tables/src/page_table.rs
```diff
@@ -660,29 +659,38 @@ impl<T: PagingMode> PageTableInner<T> {
...
             .ok_or(Error::OutOfMapRange)?;
 
-        let iterator = start_addr
-            .iter_from_with_size(page_size)
-            .ok_or_else(|| Error::AddressMisaligned(start_addr.bits()))?;
+        let mut remaining = num_pages;
+        let mut vaddr = start_addr;
 
-        for vaddr in iterator.take(num_pages as usize) {
+        while remaining > 0 {
```
sboeuf, line 737: Just for my own comprehension, you need this loop to handle non contiguous pages, is that correct?

riscv-page-tables/src/page_table.rs
```diff
@@ -660,29 +659,38 @@ impl<T: PagingMode> PageTableInner<T> {
...
             let mut table = PageTable::from_root(self);
```
sboeuf, line 738: You could move this before the loop:
```rust
let root_table = PageTable::from_root(self);
while remaining > 0 {
        let mut table = root_table;
```

riscv-page-tables/src/page_table.rs
```diff
@@ -585,6 +600,64 @@ impl<T: PagingMode> PageTableInner<T> {
...
+        &mut self,
+        start_vaddr: PageAddr<T::MappedAddressSpace>,
+        start_paddr: SupervisorPageAddr,
+        num_pages: u64,
+        page_size: PageSize,
+        perms: PteFieldBits,
+    ) -> Result<()> {
+        start_vaddr
+            .checked_add_pages_with_size(num_pages, page_size)
+            .ok_or(Error::OutOfMapRange)?;
```
sboeuf, line 620: nit: we can't really say this is because it's out of map range. It's more checking if the address overflowed over `u64`, or if the address wasn't aligned with the page size.

riscv-page-tables/src/page_table.rs
```diff
@@ -585,6 +600,64 @@ impl<T: PagingMode> PageTableInner<T> {
...
+        start_paddr
+            .checked_add_pages_with_size(num_pages, page_size)
+            .ok_or(Error::OutOfMapRange)?;
```
sboeuf, line 623: Ditto.

riscv-page-tables/src/page_table.rs
```diff
@@ -585,6 +600,64 @@ impl<T: PagingMode> PageTableInner<T> {
...
     }
 
+    /// Creates contiguous mappings, both physically and virtually, of `num_pages` pages, starting
+    /// from `start_vaddr` and `start_paddr`, with the same permissions.
+    ///
+    /// # Safety
+    ///
+    /// The caller must guarantee that app pages reference from `start_paddr` for `num_pages` are
+    /// uniquely owned by the root `GuestStagePageTable`.
+    #[allow(dead_code)]
```
sboeuf: This doesn't need to be dead code since it's called from `map_contiguous`.

riscv-page-tables/src/page_table.rs
```diff
@@ -878,6 +951,43 @@ impl<'a, T: FirstStagePagingMode> FirstStageMapper<'a, T> {
...
     }
 
+    /// Maps `num_pages` pages contiguously both physically and virtually, starting from
+    /// `start_vaddr`, which maps to `start_paddr`. The caller must guarantee that the `num_pages`
+    /// pages starting from `start_paddr` are safe to map in this page table.
+    ///
+    /// # Safety
+    ///
+    /// Don't create aliases.
+    pub unsafe fn map_contiguous(
```
sboeuf, line 960: Missing `#[allow(dead_code)]`.

riscv-page-tables/src/page_table.rs
```diff
@@ -585,6 +600,64 @@ impl<T: PagingMode> PageTableInner<T> {
...
+        start_paddr
+            .checked_add_pages_with_size(num_pages, page_size)
+            .ok_or(Error::OutOfMapRange)?;
+
+        let mut remaining = num_pages;
+        let mut vaddr = start_vaddr;
+        let mut paddr = start_paddr;
+
+        while remaining > 0 {
+            let mut table = PageTable::from_root(self);
```
sboeuf, line 630: You could move it out of the loop.

sboeuf (approved):
Looks good :)

riscv-page-tables/src/page_table.rs
glg-rv: Disagree. The real answer is `raw()`, it allows us to be explicit on what we're passing. I took this snipped from the previous code. But now that we have `raw()` I am going to change into that, which is what I also use elsewhere in this PR.

riscv-page-tables/src/page_table.rs
glg-rv: Same as above. The advantage of `RawAddr::from` over into is that you have a clue on what you're passing when reading the code. I find it much easier to be explicit on types, when it doesn't require much like this case. (again, will use `raw()`.

riscv-page-tables/src/page_table.rs
glg-rv: Heh. Well one could say that if the end of the range overflows the address space we're out of range. :-)

That was essentially the error passed before. Speficifying AddressOverflow over OutOfMap would gather no real advantage.

riscv-page-tables/src/page_table.rs
glg-rv: Same as previously commented.

riscv-page-tables/src/page_table.rs
glg-rv: Very nit and very true. Added that mid-commit to test then more code got added.

riscv-page-tables/src/page_table.rs
glg-rv: Of course. This is temporary and I am just trying to make the iteration explicit.

riscv-page-tables/src/page_table.rs
glg-rv: Ah, now I remember why. :-)

The value is considered moved once assigned to table. So has to be recreated every time anyway.

sboeuf: Ah too bad :)

riscv-page-tables/src/page_table.rs
glg-rv: Changed to raw() :)

riscv-page-tables/src/page_table.rs
glg-rv: No, to handle indexes in the other table.

riscv-page-tables/src/page_table.rs
glg-rv: Changed to Address overflow since I was there.

riscv-page-tables/src/page_table.rs
glg-rv: fixed.

PR #305, opened 2023-03-31
Reduce use of .expect() during boot

rbradford:
- main: Return error from build_memory_map() for hypervisor image check
- main: Don't use .expect() for startup error handling

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -25,6 +25,7 @@
 #![cfg_attr(test, allow(unused))]
 
 use core::alloc::{Allocator, GlobalAlloc, Layout};
+use core::fmt::Display;
 use core::ptr::NonNull;
 use test_system::*;
 extern crate alloc;
@@ -44,7 +45,7 @@ mod vm_interrupts;
 mod vm_pages;
 mod vm_pmu;
 
-use device_tree::{DeviceTree, Fdt};
+use device_tree::{DeviceTree, DeviceTreeError, Fdt};
 use drivers::{
     imsic::Imsic, iommu::Iommu, pci::PcieRoot, pmu::PmuInfo, reset::ResetDriver, uart::UartDriver,
     CpuInfo,
@@ -208,11 +209,11 @@ fn find_available_region(mem_map: &HwMemMap, size: u64) -> Option<SupervisorPage
 }
 
 /// Creates a heap from the given `mem_map`, marking the region occupied by the heap as reserved.
-fn create_heap(mem_map: &mut HwMemMap) {
+fn create_heap(mem_map: &mut HwMemMap) -> Result<(), Error> {
     const HEAP_SIZE: u64 = 16 * 1024 * 1024;
 
-    let heap_base = find_available_region(mem_map, HEAP_SIZE)
-        .expect("Not enough free memory for hypervisor heap");
+    let heap_base = find_available_region(mem_map, HEAP_SIZE).ok_or(Error::HeapOutOfSpace)?;
+
     mem_map
         .reserve_region(
             HwReservedMemType::HypervisorHeap,
@@ -230,6 +231,7 @@ fn create_heap(mem_map: &mut HwMemMap) {
         .unwrap()
     };
     HYPERVISOR_ALLOCATOR.call_once(|| HypAlloc::from_pages(pages.clean()));
+    Ok(())
 }
 
 /// Initialize (H)S-level CSRs to a reasonable state.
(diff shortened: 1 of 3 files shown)


src/main.rs
```diff
@@ -330,13 +332,84 @@ extern "C" fn primary_init(_hart_id: u64, _fdt_addr: u64) -> CpuParams {
...
+    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+        use RequiredDeviceProbe::*;
+        match self {
+            Imsic(e) => write!(f, "IMSIC: {:?}", e),
+            Pci(e) => write!(f, "PCI: {:?}", e),
+            Reset(e) => write!(f, "Reset: {:?}", e),
+            Uart(e) => write!(f, "UART: {:?}", e),
+        }
+    }
+}
```
glg-rv, line 359: I think this is a bit too boiler-plate-y for something that might change in the future. I would rather push something like this in the drivers infrastracture, so that it becomes a standard probe return, rather than forcing it here.

src/main.rs
```diff
@@ -546,27 +623,55 @@ extern "C" fn primary_init(hart_id: u64, fdt_addr: u64) -> CpuParams {
...
 #[cfg(not(test))]
 #[no_mangle]
-extern "C" fn primary_main() {
+extern "C" fn primary_init(hart_id: u64, fdt_addr: u64) -> CpuParams {
+    match _primary_init(hart_id, fdt_addr) {
+        Err(e) => {
+            // Safe as nothing can fail before console is setup
+            println!("Error during initialization on bootstrap CPU: {}", e);
+            // Safe as if shutdown fails it triggers a wfi loop (via abort())
+            poweroff();
```
glg-rv, line 643: I like the idea of controlling what happens on boot error in a better way. I personally think Debug print would just do fine.

src/main.rs
```diff
@@ -600,21 +705,32 @@ extern "C" fn secondary_init(hart_id: u64) -> CpuParams {
...
 
     HOST_VM.wait().run(PerCpu::this_cpu().cpu_id().raw() as u64);
     poweroff();
 }
 
+/// Start salus on secondary CPUs. This is called once we switch to the hypervisor page-table.
+#[cfg(not(test))]
+#[no_mangle]
+extern "C" fn secondary_main() {
+    if let Err(e) = _primary_main() {
```
glg-rv, line 726: This can't be the same function. there are similarities at the moment, and APs wait for host vm, bsp creates it.

I would kind of flip the names, with `_function` the function called from assembler, and `function` those called from rust that return an error.

glg-rv (changes requested):
The init code is due some major changes. the main functions between primary and secondary are simlilar right now simply because of this setup being very new.

In my PR for VA stack that introduced this I wrote that the refactoring of init would require some discussions.

I like the fundamentals of this PR. It's definitely the right direction. But I would keep the difference between primary and secondary explicit, and not try to find common patterns at this moment. Things will change.

I personally don't like the display implementations and would live with Debug, but that's far from being a strong opinion. :-)

src/main.rs
rbradford: Thank you, I think this was a typo after the rebase following the unit test changes. I think this is also why the tests are failing :-)

src/main.rs
rbradford: The other option was to add 4 different errors to the main `Error` enum and I thought this was less intrusive.

src/main.rs
rbradford: Took me a while to understand what you mean here. But I think you mean there shouldn't be a `Display` implementation for the top level Error or any of the other errors. I guess that's a matter of personal preference - I think that the way nested errors are presented with chains of round brackets aren't the nicest and it can be less obvious what the error is than the human written message. But it would reduce the compiled size by not having the strings and also reduce the amount of code.

src/main.rs
rbradford: I'm happy to flip where the `_` prefix is used. I had to make a call on that.

src/main.rs
glg-rv: Yes, we need to check in the future then that other device rpobes use that error code.

src/main.rs
glg-rv: Sure, definitely a matter of personal preference and not important for the sake of this patch. Regarding the message, it's a panic, will require interpretation. I think the amount of code and data required for the string is definitely negligible.

glg-rv (approved):

abrestic-rivos (approved):

Issue #100, opened 2022-09-16
Consider alternatives to using `unwrap()`


abrestic-rivos:
Our usage of `unwrap()` generally falls into one of two categories:
1. During startup when initializing Salus and loading the host VM.
2. At runtime when we've expected we've made the necessary guarantees for an operation to succeed.

For (1), we should probably propagate these errors up to `main()` and let it `expect()` or `panic!()` as appropriate.

For (2), I think we should consider something like a `warn!()` or `bug!()` macro that dumps a message to the console and taints or poisons the TSM as an indication that it may be in an inconsistent state, similar to `WARN_ON` in Linux. TVM's could then opt into (or out of) running on a tainted TSM. While the end result to the user might be the same -- TVMs can't run and the TSM needs to be reset -- it would at least provide a better defined failure state than panicking.

PR #308, opened 2023-04-03
Remove last use of .expect() during Salus boot

rbradford:
This PR removes the last use of .expect() during the Salus boot replacing with error reporting.

- main: Use fmt::Display for errors that have it
- hyp_map, smp: Propagate error from stack memory allocation
- page-tracking: Propagate errors from hypervisor page allocator

See: #100

### DIFF
--- a/src/hyp_map.rs
+++ b/src/hyp_map.rs
@@ -10,7 +10,8 @@ use data_model::{DataInit, VolatileMemory, VolatileMemoryError, VolatileSlice};
 use page_tracking::{HwMemMap, HwMemRegion, HwMemRegionType, HwReservedMemType, HypPageAlloc};
 use riscv_elf::{ElfMap, ElfSegment, ElfSegmentPerms};
 use riscv_page_tables::{
-    FirstStageMapper, FirstStagePageTable, PagingMode, PteFieldBits, PteLeafPerms, Sv48,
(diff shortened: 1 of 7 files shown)


src/hyp_map.rs
```diff
@@ -577,7 +590,7 @@ impl HypMap {
             .next()
             .unwrap();
```
glg-rv, line 591: .expect() was added as an improvement to the unwrap() that was all over the initialization code.

I think in this respect this unwrap, and others if exist, should be equally handled.

In general, I'd like error happening during initialization to be clearly marked as fatal. Either in comments or names.

glg-rv (commented):
Overall lgtm.

src/hyp_map.rs
rbradford: I agree, it would be a lot of work but it would be nice to know that there is no scope for panic.

glg-rv: In this case would be a `ok_or` ? :-)

rbradford: Yes `.ok_or` is a good solution for places like this.

sboeuf (approved):
:+1:

abrestic-rivos (approved):

src/hyp_map.rs
glg-rv: And hasn't been changed why?

PR #310, opened 2023-04-05
page-tracking: Get rid of Unassigning state

### LINKED ISSUE #268: Enhance IMSIC flow
Adapt the IMSIC `unbind`/`rebind` bits to internally use the new `block`/`remove` flow would allow us to completely get rid of the existing __unassignment__ flow.

sboeuf:
Relying on the new block and remove flow, we can get rid of the Unassigning state which was nothing but a duplicate of what is handled through this new design.

The IMSIC file unbind/rebind codepath was the one still using the unassignment method, which is now replaced with the block and remove equivalent.

Fixes #268

### DIFF
--- a/page-tracking/src/page_info.rs
+++ b/page-tracking/src/page_info.rs
@@ -45,11 +45,6 @@ pub enum PageState {
     /// the current owner's address space.
     Converting(TlbVersion),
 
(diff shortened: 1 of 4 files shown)


glg-rv:
While I have not been following very closely the changes in the block/remove, this LGTM.

dgreid (approved):

PR #312, opened 2023-04-05
build: Treat unit tests warnings as errors

### LINKED ISSUE #309: bazel build //:clippy-all doesn't check tests
I think it would be good if this could also run clippy across  test files as well as `#[cfg(test)]` blocks e.g. the equivalent of `cargo clippy --tests`

sboeuf:
Turn on the rustc flag `-Dwarnings` for all bazel rust tests so that all warnings would show up as errors.

Fixes #309

### DIFF
--- a/drivers/src/lib.rs
+++ b/drivers/src/lib.rs
@@ -215,7 +215,7 @@ mod tests {
     #[test]
     fn build_cpu_info() {
         let tree = stub_tree();
-        CpuInfo::parse_from(&tree);
+        CpuInfo::parse_from(&tree).unwrap();
 
         let cpu_info = CpuInfo::get();
         assert!(cpu_info.has_sstc());
@@ -242,7 +242,7 @@ mod tests {
     #[test]
     fn probe_imsic() {
         let tree = stub_tree();
-        CpuInfo::parse_from(&tree);
+        CpuInfo::parse_from(&tree).unwrap();
         let mut mem_map = stub_mem_map();
         Imsic::probe_from(&tree, &mut mem_map).unwrap();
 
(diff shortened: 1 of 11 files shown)


drivers/src/lib.rs
```diff
@@ -215,7 +215,7 @@ mod tests {
     #[test]
     fn build_cpu_info() {
         let tree = stub_tree();
-        CpuInfo::parse_from(&tree);
+        let _ = CpuInfo::parse_from(&tree);
```
rbradford, line 218: Don't we want to panic on error here from the Result? (for the unit test)?

rbradford:
> /cc @rbradford I think this is fixing #309 :)

Although this is definitely an improvement - catching errors in the test build. I don't think it is the equivalent of `cargo clippy --tests` as this make clippy cover the test code?

drivers/src/lib.rs
sboeuf: Yes of course, don't know why I did that... I'll simply unwrap.

sboeuf:
> > /cc @rbradford I think this is fixing #309 :)
> 
> Although this is definitely an improvement - catching errors in the test build. I don't think it is the equivalent of `cargo clippy --tests` as this make clippy cover the test code?

Well I know this is different but this is eventually reaching the same goal of "not allowing warnings to be missed".

rbradford:
> > > /cc @rbradford I think this is fixing #309 :)
> > 
> > 
> > Although this is definitely an improvement - catching errors in the test build. I don't think it is the equivalent of `cargo clippy --tests` as this make clippy cover the test code?
> 
> Well I know this is different but this is eventually reaching the same goal of "not allowing warnings to be missed".

This is definitely an improvement but it doesn't solve #309 which is about clippy not covering the tests as clippy warns about things that rustc doesn't.

rbradford (approved):

dgreid (approved):

PR #315, opened 2023-04-11
Fix U-mode sharing after huge pages support

glg-rv:
Huge Page support broke U-mode guest mapping, and in general did not consider it.

This patch fixes this, but avoids a case that now might happen: that shared pages are mapped in guest as huge pages.

This will require further work.

### DIFF
--- a/src/vm_pages.rs
+++ b/src/vm_pages.rs
@@ -1455,7 +1455,7 @@ impl<'a, T: GuestStagePagingMode> FinalizedVmPages<'a, T> {
         self.do_get_converted_pages::<Page<ConvertedDirty>>(page_addr, page_size, num_pages)
     }
 
-    /// Acquries an exclusive reference to the `num_pages` shared pages starting at `page_addr`.
+    /// Acquries an exclusive reference to the `num_pages` shared host pages starting at `page_addr`.
     pub fn get_shareable_pages(
         &self,
         page_addr: GuestPageAddr,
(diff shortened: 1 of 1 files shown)


src/vm_pages.rs
```diff
@@ -1455,7 +1455,7 @@ impl<'a, T: GuestStagePagingMode> FinalizedVmPages<'a, T> {
         self.do_get_converted_pages::<Page<ConvertedDirty>>(page_addr, page_size, num_pages)
     }
 
-    /// Acquries an exclusive reference to the `num_pages` shared pages starting at `page_addr`.
+    /// Acquries an exclusive reference to the `num_pages` shared host pages starting at `page_addr`.
     pub fn get_shareable_pages(
```
abrestic-rivos, line 1459: Now that I'm looking at this again, the logic in the closure passed to `get_mapped_pages()` doesn't seem right. We're checking that the entire range is physically contiguous, but really we should only care that sub-pages within the huge page are physically contiguous. `do_get_converted_pages()` has a similar problem.

cc @sboeuf

abrestic-rivos (approved):

src/vm_pages.rs
sboeuf: Ah I see, let me have a look!

sboeuf: #316 should cover this.

PR #316, opened 2023-04-11
vm_pages: Only check subpages from a huge page are contiguous

sboeuf:
Both do_get_converted_pages() and get_shareable_pages() were checking the entire range covered by possibly multiple pages was made of contiguous pages. This is not necessary as the ony thing that's needed is to ensure subpages of each huge page are contiguous. This simplifies the requirement.


src/vm_pages.rs
```diff
@@ -1395,18 +1395,23 @@ impl<'a, T: GuestStagePagingMode> FinalizedVmPages<'a, T> {
...
+                // We need to check the first subpage is suitably aligned and that every following
+                // subpage covering the same huge page is contiguous to the previous one. That's
+                // because we have some logic below that allows to create huge pages out of 4k pages,
+                // by skipping the intermediate pages. That's why we must ensure the whole range is
+                // made of physically contiguous subpages for each huge page and they are suitably
+                // aligned.
+                if va.is_aligned(page_size) {
+                    // This is our way of resetting the check for contiguous subpages.
+                    prev_pa = None;
+                }
```
abrestic-rivos, line 1414: I don't think we need to propagate the vaddr to the closure in order to perform this check -- if incrementing `prev_pa` moves us to a `page_size` boundary, just set `prev_pa` to `None`.

abrestic-rivos (approved):

src/vm_pages.rs
sboeuf: Done!

src/vm_pages.rs
```diff
@@ -1402,22 +1402,33 @@ impl<'a, T: GuestStagePagingMode> FinalizedVmPages<'a, T> {
...
+                // subpage covering the same huge page is contiguous to the previous one. That's
+                // because we have some logic below that allows to create huge pages out of 4k pages,
+                // by skipping the intermediate pages. That's why we must ensure the whole range is
+                // made of physically contiguous subpages for each huge page and they are suitably
+                // aligned.
                 if let Some(addr) = prev_pa {
+                    let next_pa = addr.checked_add_pages_with_size(1, ps).unwrap();
                     // Check the current page is contiguous with the previous page
-                    if addr.checked_add_pages_with_size(1, ps) != Some(pa) {
+                    if next_pa.is_aligned(page_size) {
```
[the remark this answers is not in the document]
sboeuf: Done!

PR #329, opened 2025-05-01
Fix 'Property not found' when executing 'qemu-system-riscv64'

iamywang:
When executing `qemu-system-riscv64`, I encounter an error indicating that the properties `smaia` and `ssaia` were not found.

After reviewing QEMU’s source code:
- https://github.com/rivosinc/qemu/blob/salus-integration-10312022/target/riscv/cpu.c#L1067
- https://github.com/rivosinc/qemu/blob/salus-integration-10312022/target/riscv/cpu.c#L1068

I found that the correct property names are `x-smaia` and `x-ssaia` (in `scripts/common_variables`).

### DIFF
--- a/scripts/common_variables
+++ b/scripts/common_variables
@@ -48,7 +48,7 @@ EXTRA_QEMU_ARGS=${EXTRA_QEMU_ARGS:-""}
 EXTRA_CPU=",v=on,vlen=256,elen=64"
 
 CPU_TYPE=rv64${EXTRA_CPU}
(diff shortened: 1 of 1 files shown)


dgreid:
What version of qemu are you using? aia was ratified last year and the x- prefix is no longer needed.

iamywang:
I am using [salus-integration-10312022](https://github.com/rivosinc/qemu/tree/salus-integration-10312022) according to the `Latest known-working branches:` in `README.md`.

PR #334, opened 2025-06-10
Support for multiple PCI roots / IOMMUs

mnissler-rivos:
This PR generalizes salus' handling of PCI segments and IOMMUs:
- Salus can now manage multiple PCI segments and multiple IOMMUs
- Device-to-IOMMU mapping now respects mapping information in the device tree to determine which IOMMU is responsible for a given device.
- The IOMMU device directory code can handle both base and extended format, as well as 1-3 levels of tree depth.
- There's a new build option for A/D PTE bit management: Either rely on A/D updating support (which must be present), or initialize A/D bits to 1.
- Support the Enhanced Allocation PCI capability.

### DIFF
--- a/drivers/src/pci/root.rs
+++ b/drivers/src/pci/root.rs
@@ -3,9 +3,10 @@
 // SPDX-License-Identifier: Apache-2.0
 
 use alloc::alloc::Global;
+use alloc::vec::Vec;
 use arrayvec::{ArrayString, ArrayVec};
 use core::marker::PhantomData;
-use device_tree::{DeviceTree, DeviceTreeResult};
+use device_tree::{DeviceTree, DeviceTreeNode, DeviceTreeResult};
 use hyp_alloc::{Arena, ArenaId};
 use page_tracking::{HwMemMap, PageTracker};
 use riscv_pages::*;
@@ -41,7 +42,7 @@ pub struct PcieRoot {
     msi_parent_phandle: u32,
 }
 
-static PCIE_ROOT: Once<PcieRoot> = Once::new();
+static PCIE_ROOTS: Once<Vec<PcieRoot>> = Once::new();
 
 // A `u64` from two `u32` cells in a device tree.
 struct U64Cell(u32, u32);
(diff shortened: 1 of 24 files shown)


drivers/src/pci/root.rs
```diff
@@ -305,11 +305,18 @@ impl PcieRoot {
...
+            match bar.provenance() {
+                BarProvenance::PciHeader => {
+                    let range = self.alloc_hypervisor_resource(bar.bar_type(), bar.size())?;
+                    // `range.base()` must be within a PCI root window.
+                    let base = self.physical_to_pci_addr(range.base().into()).unwrap();
+                    // BAR index must be valid.
+                    dev.set_bar_addr(bar.index(), base).unwrap();
+                }
+                BarProvenance::EnhancedAllocation => {
+                    // TODO: need to mark the region so it doesn't get exposed to the host!
```
mnissler-rivos, line 349: This TODO is the reason why I've marked the PR as draft, since it potentially breaks isolation.

The problem is that EA regions are part of the MMIO region in the address map, but can't be assigned freely. This creates at least two problems:

1. The current allocation code might create allocations that conflict with EA regions.
2. EA regions for hypervisor devices aren't marked as reserved, so may end up getting mapped into the host VM (what this TODO states).

I think a proper solution would need a more sophisticated resource manager, which can track assigned resources with relevant attributes (EA/fixed vs. allocatable, owner, etc.) rather than the current simplistic split where host resources go before the splitting point, hypervisor resources after.

Generalizing the resource management code is another non-trivial change, so I figured I'd ask for your thoughts before spending time on that, perhaps you have an opinion how this should work or suggestions for clever hacks that allow us to land this PR without revamping resource management.

abrestic-rivos: Yeah, I don't have a simple suggestion for you here. I think we will have to reserve EA resources during enumeration and track ownership of individual ranges within the resource vs the single split we have currently.

mnissler-rivos: OK, I've replaced the TODO with an `unimplemented!()` and added a build flag to skip over it.

PR #335, opened 2025-06-13
fixes to run on spike

dgreid:
some useful changes that allow salus to run on spike in addition to qemu.


atishp04 (approved):
Other than that, LGTM.

